PR #21, opened 2018-06-04
started ui components

JahlomP:
It adds ui components menus


camwiegert:
Thanks for taking the time to do this! We've actually got different plans for the `/components` route. This looks to be what we've got in the [API reference](https://beta.ionicframework.com/docs/api).

PR #30, opened 2018-07-16
feat(components-docs): add link to api docs for each component

jgw96:
This resolves the task we prioritized around adding a link to the api docs for each component in the component docs

- generate new anchor tag that links to api docs
- adds css for this new anchor

### DIFF
--- a/src/components.d.ts
+++ b/src/components.d.ts
@@ -605,6 +605,39 @@ declare global {
 declare global {
 
   namespace StencilComponents {
+    interface DocsLink {
+      'compName': string;
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


camwiegert:
Alright, so upon further investigation. I think it's better to localize this to the `component-preview` component. I think I'd rather not complicate the markdown renderer and instead just dynamically add these links from the page.

If we go down the road of adding special cases to the renderer, I'd worry about it getting out of hand. This way the special case is handled in [the same location](https://github.com/ionic-team/ionic-docs/blob/master/src/components/component-preview/component-preview.tsx#L41-L48) as the rest of the special pieces of the `components` page.

@jgw96 What do you think?

camwiegert:
Going to close this as it was addressed in a different way. Thanks @jgw96!

PR #40, opened 2018-07-26
add app type thumbnails to building/starting

janpio:
makes sense to be able to visualize start app types here as well


camwiegert:
I think the text descriptions perform the same function here.

PR #42, opened 2018-07-26
Link to @ionic-native/ionic-webview in Ionic Native docs instead of npm

janpio:
...


camwiegert:
@janpio Makes sense to me. Thanks!

PR #44, opened 2018-07-26
Storage: Add prev/next links

janpio:
...


src/content/building/storage.md
```diff
@@ -1,3 +1,9 @@
+---
+previousText: 'Webview'
```
camwiegert: @janpio Could we use 'Web View' as `previousText` here to match the page title?

janpio: Sure, changed.

PR #50, opened 2018-07-26
PWA: add missing word

janpio:
...


camwiegert:
@janpio Thanks for opening this, but I believe it is grammatically correct as-is. I can't find a definitive source on this, so please share if you've got one. 😄

janpio:
Hm, the sentence currently is:

> ... they can run just as well as a Progressive Web App as they can a native app.

Reformatted:

> ... they can run just as well 
> as a Progressive Web App 
> as they can 
> a native app.

Rewritten this is:

> It can run just as well
> as a Foo 
> as it can 
> a Bar.

In my "feeling" this should be "as a Bar" as well.

bensperry:
The current sentence is already correct as-is.

PR #129, opened 2018-09-20
docs(installation-ios): warn about Xcode 10

danielsogl:
Currently XCode 10 does not work with Cordova.
See: https://github.com/apache/cordova-ios/issues/407


Athuli7:
This is urgent. XCode errors are too ambigous that people will pull thier hair out before they figure out what's going on.

dwieeb:
Thanks @danielsogl!

PR #136, opened 2018-10-01
Adding electron-builder to docs

Athuli7:
A much simpler approach

The desktop app page just had links to the electron website and it is extremely hard to follow for newcomers. And electron-builder is a much simpler approach, especially if you add electron-builder.

-
-
-

**Fixes**: #


camwiegert:
@Athuli7 Thanks, but we would prefer not to suggest any additional build tools.

PR #145, opened 2018-10-09
[V4] Remove img closing tag

angajime:

The List section is contains an image in the example using the `img` tag which is a self closing tag in HTML5 and it will crash the browser if you try to load it using `ionic serve` as is.

- Removing the `</img>` tag is enough to correct this.


dwieeb:
Great work! Thanks @angajime

Issue #116, opened 2018-08-27
Missing documentation for Platform and Platform.backButton


rveerd:
**I'm submitting a ...**  (check one with "x")
[ ] bug report
[x] feature request

**URL:**
https://beta.ionicframework.com/docs/api

**Current behavior:**
The page mentions `Platform` but this is not a link. Documentation for `Platform` does not exist. Backward incompatible changes esp. hardware back button are not documented.

**Expected behavior:**
A working link to the documentation for `Platform`. Documentation for `Platform`. Documentation of backward incompatible changes.

**Steps to reproduce:**
n/a

**Other information:**
Especially functionality for the hardware back button as this was changed from Ionic 3. Document `backButton` Observable. A minimal example of how to use it to go to the previous page in a typical Ionic app would be great.

Add the change from `registerBackButtonAction` to `backButton` in the migration guide.

bryplano:
There does appear to be an inconsistency there. The "old" documentation does include a page about that: https://ionicframework.com/docs/api/platform/Platform/#platforms

We'll review this with the team.

camwiegert:
The Platform docs are now located [here](https://ionicframework.com/docs/utilities/platform).

Takkuz:
Yes, but it still miss any documentation about `.backButton` properties.

PR #151, opened 2018-10-15
Improved on best practices for platform handling

selipso:
- Added where `Platform` is imported from (Wouldn't have known if not for npm Intellisense)
- Added service-layer wrapper as a recommended method for abstracting platform dependency. This is a best practice in Angular projects and encourages SRP.

**Fixes**: Partial improvement for #116


camwiegert:
@selipso Thanks for this! We'd rather not make that architecture suggestion on the docs, but plan to improve the documentation for `Platform`.

Issue #38, opened 2018-07-26
[beta docs] Broken links


janpio:
A list of broken links in the new docs:

* [x] https://beta.ionicframework.com/docs/faq/tips/ionic/cli linked from https://beta.ionicframework.com/docs/faq/tips/ionic/cli
* [ ] https://beta.ionicframework.com/docs/action-sheet-controller/ActionSheetController linked from https://beta.ionicframework.com/docs/api/action-sheet/
* [ ] https://beta.ionicframework.com/docs/alert-controller/AlertController linked from https://beta.ionicframework.com/docs/api/alert/
* [ ] https://beta.ionicframework.com/docs/alert/AlertController/ and https://beta.ionicframework.com/docs/alert/AlertController linked from https://beta.ionicframework.com/docs/api/select/
* [ ] https://beta.ionicframework.com/docs/api/grid/(../Col linked from https://beta.ionicframework.com/docs/api/grid/
* [ ] https://beta.ionicframework.com/docs/api/item-sliding/(../item-options linked from https://beta.ionicframework.com/docs/api/item-sliding/
* [ ] https://beta.ionicframework.com/docs/intro/support/versioning linked from https://beta.ionicframework.com/docs/intro/support/

Missing images:

* [x] https://beta.ionicframework.com/docs/faq/assets/img/running/ios-xcode-identity-setup.png and https://beta.ionicframework.com/docs/faq/assets/img/running/ios-xcode-signing-setup.png used on https://beta.ionicframework.com/docs/faq/native/
* [x] https://beta.ionicframework.com/docs/components/assets/img/avatar-finn.png linked from https://beta.ionicframework.com/docs/components/

and probably many more, but finding them is a pain because of #37

kael:
One more broken link: https://beta.ionicframework.com/docs/list/List (via https://beta.ionicframework.com/docs/api/item-sliding/)

PR #156, opened 2018-10-24
docs(theming): update broken relative URLs

d2s:

- Update previously broken relative URLs to point to a new location.

- Instead of relative location based on the directory where the specific page is located, use URLs that point to the root of the domain.

_Related to:_ #38


brandyscarney:
Thanks for the PR!

PR #166, opened 2018-11-03
Update apiMap to 4.0.0-beta.15, in accordance with src/versions.ts

kplhub:

In  Docs API Reference 4.0.0-beta.15 the list of components is not displayed, because docs-api-map.ts apiMap still is "4.0.0-beta.0"

Update docs-api-map.ts


dwieeb:
Thanks @kplhub!

PR #218, opened 2018-12-05
deploy.getCurrentVersion() returns a Promise so need an await

ZenSide:
await is missing in sync() example for getCurrentVersion() call. Result is actually a promise and not the ISnapshot result which results to an error 2 lines after (versionId doesn't exists for a Promise)

sync() example is not generating an error

- replace `const currentVersion = Pro.deploy.getCurrentVersion();` by `const currentVersion = await Pro.deploy.getCurrentVersion();`

**Fixes**: #

### DIFF
--- a/src/content/pro/deploy/api.md
+++ b/src/content/pro/deploy/api.md
@@ -94,7 +94,7 @@ ___
 ```js
 async performAutomaticUpdate() {
   try {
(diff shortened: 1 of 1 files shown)


dwieeb:
Thanks @ZenSide!

PR #289, opened 2019-01-16
docs(installation/android.md): strengthen JDK8 requirement wording

rtpHarry:
It's not just recommended, it's required. I tried this today and there is a flat out version check in the source code that locks it to JDK8.

Also, added https to the download link as the site runs on that now.


rtpHarry:
FYI if you don't have much time to research this then here is some help:

I also worked with the cordova-android team to clarify this in their error messages and it's been accepted:

 - https://github.com/apache/cordova-android/pull/619
 - https://github.com/apache/cordova-android/pull/620

If you look in the files changed the version check is just around those changed lines.

PR #324, opened 2019-01-24
Appflow quickstart

nphyatt:
This is a big refactor of the Appflow Docs.
I checked it over pretty well and added a couple redirects but not 💯if I added the redirects to the right place and right syntax but I think I did.


sajjad7110:
Couple of things

PR #329, opened 2019-01-24
docs(building/storage): improve usage snippet

rtpHarry:
Bring the usage snippet over from building/cross-platform.md as it uses the cleaner async / await keywords.


mhartington:
Hmm, I think we'll actually keep this the way it is. Thanks though!

PR #358, opened 2019-01-26
fix: resolve markdown links to handle relative URLs

ltm:


scripts/build-pages/markdown-renderer/link.ts
```diff
@@ -1,9 +1,15 @@
+import { resolve } from 'url';
+
 const isInternal = /^\/docs\/.*/;
 
-export default (href: string, title: string, text: string) => {
+export default function(href: string, title: string, text: string) {
+  if (this.options.baseUrl && href.startsWith('../')) {
```
camwiegert: Should we apply this to all relative links? (`./`, `../`)

ltm: Good point. I've updated the check so it applies to everything except fragment-only URLs. The `url.resolve` method handles other relative and absolute URLs correctly.

camwiegert:
Looking at the way that GitHub resolves links in a README, they seem to resolve all base paths as if they were a directory.

```
githubResolve('/some/path', '../other-path') => '/some/other-path'
githubResolve('/some/path/', '../other-path') => '/some/other-path'
```

Whereas Node's plain url.resolve would handle it like this:

```
url.resolve('/some/path', '../other-path') => '/other-path'
url.resolve('/some/path/', '../other-path') => '/some/other-path'
```

So I think we should handle it like GitHub and append a trailing slash during resolution if there isn't one.

```
resolve(/\/$/.test(basePath) ? basePath : `${basePath}/`, href)
```

Does that make sense?

ltm:
@camwiegert I was achieving that with the trailing slash here:

https://github.com/ionic-team/ionic-docs/blob/dc41c2ed89bf6a6f9cd702355d6dd9c942fdd92c/scripts/build-pages/page-types/api.ts#L20

But it looks like that got lost in the merge.

camwiegert:
@ltm Oh, I see. I think I'd rather move that into the link renderer so that it will apply to every link instead of just the API pages.

ltm:
@camwiegert 👍

PR #381, opened 2019-01-31
docs(): add Utilities and Config sections

liamdebeasi:
This PR adds basic "Config" documentation. It also creates a "Utilities" section for documenting other APIs like this.

### DIFF
--- a/src/pages/utilities/config.md
+++ b/src/pages/utilities/config.md
@@ -105,7 +105,7 @@ Below is a list of config options that Ionic uses.
 | `infiniteLoadingSpinner` | `SpinnerTypes`     | Overrides the default spinner type in all `<ion-infinite-scroll-content>` components.
 | `loadingEnter`           | `AnimationBuilder` | Provides a custom enter animation for all `ion-loading`, overriding the default "animation".
 | `loadingLeave`           | `AnimationBuilder` | Provides a custom leave animation for all `ion-loading`, overriding the default "animation".
-| `loadingSpinner`         | `SpinnerTypes`     | Overrides the default spinner for all `ion-loading` overlays, ie. the ones.
+| `loadingSpinner`         | `SpinnerTypes`     | Overrides the default spinner for all `ion-loading` overlays.
 | `menuIcon`               | `string`           | Overrides the default icon in all `<ion-menu-button>` components.
 | `menuType`               | `string`           | Overrides the default menu type for all `<ion-menu>` components.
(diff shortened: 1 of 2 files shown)


src/pages/utilities/config.md
```diff
@@ -0,0 +1,68 @@
...
+      rippleEffect: false,
+      mode: 'md'
+    }), 
+    AppRoutingModule
+  ],
+  ...
+})
+```
+
+In this example, we are disabling the material design ripple effect across the app as well as forcing the app mode to be material design.
```
brandyscarney: Probably say:

> In the above example, we are disabling the Material Design ripple effect across the app, as well as forcing the mode to be Material Design.

src/pages/utilities/config.md
```diff
@@ -0,0 +1,68 @@
...
+| `actionSheetEnter`       | `AnimationBuilder` | Provides a custom enter animation for all `ion-action-sheet`, overriding the default "animation".
+| `actionSheetLeave`       | `AnimationBuilder` | Provides a custom leave animation for all `ion-action-sheet`, overriding the default "animation".
+| `alertEnter`             | `AnimationBuilder` | Provides a custom enter animation for all `ion-alert`, overriding the default "animation".
+| `alertLeave`             | `AnimationBuilder` | Provides a custom leave animation for all `ion-alert`, overriding the default "animation".
+| `animated`               | `boolean`          | When it's set to `false`, disables all animation and transition across the app.
+| `backButtonIcon`         | `string`           | Overrides the default icon in all `<ion-back-button>` components.
+| `backButtonText`         | `string`           | Overrides the default text in all `<ion-back-button>` components.
+| `hardwareBackButton`     | `boolean`          | Wherever ionic will respond to hardware go back buttons in an Android device.
```
brandyscarney: This description doesn't make sense. I think the booleans should be set up in `if` statements like:

```
If `true`, Ionic will respond to the hardware back button in an Android device
```

src/pages/utilities/config.md
```diff
@@ -0,0 +1,68 @@
...
+nextText: ''
+nextUrl: ''
+---
+
+# Config
+
+The Config lets you configure your entire app. You can set the tab button layout, icon mode, animations, and more here.
+
+## Usage
+
```
brandyscarney: Maybe here we should do something similar to components where there is a Basic Usage like:

```
### Basic example

(add some usage here about changing string or boolean types)

### Customizing Animations

(add some usage here on changing an overlay animation)
```

liamdebeasi:
@brandyscarney made the changes you requested. can you take another look?

Issue #391, opened 2019-01-31
(Documentation) Advanced theming does not explain where to set custom colors


iget-master:
# Support Question

Actually not a support question, but an issue with documentation:

The `Advanced Theming` page (https://ionicframework.com/docs/theming/advanced) teach how to create a custom color, but doesn't tell where to put it.

> To change the default values of a color, all of the listed variations for that color should be set. For example, to change the secondary color to , set the following CSS properties:

**where???**

This kind of documentation issues are present everywhere on new Ionic docs.
Please improve this documentation guys.

brandyscarney:
If you're using the Ionic CLI to start an Angular project, it would be where the colors are defined. Using the conference app as an example, it's here: https://github.com/ionic-team/ionic-conference-app/blob/master/src/theme/variables.scss#L30-L35.

So when you start a new project it would be in your `src/theme/variables.scss` file. 

Because these are CSS variables, and you're setting them in the root, they can technically go in any global CSS file. More info on CSS variables here: https://ionicframework.com/docs/theming/css-variables

Please keep in mind when you make an issue that we spend a lot of time and effort writing documentation. We can't fix something that we don't realize is a problem, so we appreciate issues being filed to bring it to our attention, but please be careful of your tone. PRs are also always welcome for improvements. Thanks!

iget-master:
I've sent a PR with a suggestion about it.

brandyscarney:
Thank you for the PR! Merged: https://github.com/ionic-team/ionic-docs/pull/392

PR #392, opened 2019-02-01
docs(theming) Explain where to place variables

iget-master:
This PR fixes the issue #391.

That's just my proposal of an explanation about where user can override default ionic variables.


brandyscarney:
Thank you 🙂

PR #415, opened 2019-02-05
docs(faq): add CORS errors section

FdezRomero:
This PR adds a new **CORS Errors** section under **FAQ**, with useful information for Ionic developers and solutions to common situations.

- What is CORS?
- How does CORS work
- CORS Headers
- Solutions for CORS Errors

Feedback and suggestions are greatly appreciated 🙏


src/pages/faq/cors.md
```diff
@@ -0,0 +1,250 @@
...
+# CORS Errors
+
+
+## What is CORS?
+
+**Cross-Origin Resource Sharing (CORS)** is a mechanism that browsers and webviews — like the ones powering Capacitor and Cordova — use to restrict HTTP and HTTPS requests made from scripts to resources in a different origin for security reasons, mainly to protect your user's data and prevent attacks that would compromise your app.
+
+In order to know if an external origin supports CORS, the server has to send some <a href="#cors-headers">special headers</a> for the browser to allow the requests.
+
+An **origin** is the combination of the **protocol**, **domain**, and **port** from which your Ionic app or the external resource is served. For example, apps running in Capacitor have `capacitor://localhost` (iOS) or `http://localhost` (Android) as their origin, and Cordova apps may run from `http://localhost:8080` or the `file://` protocol.
```
jcesarmobile: On WebView 3.x it uses `http://localhost` without port for Android, and `ionic://localhost` for iOS

src/pages/faq/cors.md
```diff
@@ -0,0 +1,250 @@
...
+### A. Enabling CORS in a server you control
+
+The correct and easiest solution is to enable CORS by returning the <a href="#server-headers-response-">right response headers</a> from the web server or backend and responding to preflight requests, as it allows to keep using `XMLHttpRequest`, `fetch`, or abstractions like `HttpClient` in Angular.
+
+These are the recommended `Access-Control-Allow-Origin` headers for Ionic apps. Please note that these headers have to be sent from the server, and don't belong in your app code.
+
+**Capacitor**
+
+```http
+Access-Control-Allow-Origin: capacitor://localhost, http://localhost
```
jcesarmobile: I don't think you can have multiple values here

src/pages/faq/cors.md
```diff
@@ -0,0 +1,250 @@
...
+
+*Replace `localhost` with your own hostname if you have changed the default in the Capacitor config.*
+
+**Ionic WebView plugin on Cordova**
+
+```http
+Access-Control-Allow-Origin: http://localhost:8080
+```
+
+*Replace port `8080` with your own if you have changed the default in the plugin config.*
```
jcesarmobile: As commented before, WebView 3.x doesn't have a port anymore. 
As we are still supporting WebView 2.x, it's probably better to have two sections, one for 3.x with the `ionic://localhost`for iOS and `http://localhost` for android and other for 2.x with the port.
Well, in fact, even in 2.x it can use `ionic://localhost` for iOS 11+ if you use `UseScheme` preference.

src/pages/faq/cors.md
```diff
@@ -0,0 +1,250 @@
...
+
+**Ionic Serve for local development**
+
+```http
+Access-Control-Allow-Origin: http://localhost:8100
```
jcesarmobile: This is the ionic serve url? if that so, shouldn't it be the local IP instead of localhost? otherwise I think it won't work on real devices, and probably not even on Android emulator.

src/pages/faq/cors.md
jcesarmobile: ionic is on webview plugin, capacitor on capacitor

src/pages/faq/cors.md
FdezRomero: By "local development" I meant "in a browser in your machine", not in a local external device. Good point though!

src/pages/faq/cors.md
FdezRomero: Right, the server would have to echo back the origin. Maybe it should suggest this instead, what do you think?

src/pages/faq/cors.md
FdezRomero: Moved it to the solutions section, thanks!

src/pages/faq/cors.md
FdezRomero: I fixed all the origins, moved them to different tables depending on the webview and recommend to reflect the origin instead, since only one can be specified in the `Access-Control-Allow-Origin` header. Thanks!

src/pages/faq/cors.md
FdezRomero: Changed the heading to "Local development in the browser", specifying the CLI commands to avoid confusion.

src/pages/faq/cors.md
FdezRomero: Changed to recommend reflecting the origin server-side based on a whitelist, adding a short example for Express apps. Thanks for pointing this out!

camwiegert:
Thanks @FdezRomero!

FdezRomero:
Happy to contribute! 💙

PR #430, opened 2019-02-06
Update outdated ion-col syntax to match current grid documentation

jsonMartin:
While going through the *first-app-v4* Ionic tutorial, I ran into an issue with the grid layout. The tutorial was using the older syntax of `col-6` instead of `size="6"`, which doesn't work properly on V4 of Ionic and caused the photos to all display on one line instead of inside a grid.

By using the updated grid syntax located in the current docs at https://ionicframework.com/docs/api/grid/, the layout performs as intended by adding two photos per row in the grid. This PR updates the documentation to reflect the current grid documentation.


brandyscarney:
Thanks for the PR!

@camwiegert This looks right - but the build is not working? It seems like it never ran, can you take a look?

camwiegert:
@brandyscarney Yeah that's odd. There was a period several days ago where builds were not triggering at all. Since this is just a content change I'm going to go ahead and merge it. Thanks @jsonMartin!

PR #473, opened 2019-02-21
doc(testing): create end-to-end testing documentation

kensodemann:


kensodemann:
I have been getting a fair number of questions pertaining to end-to-end testing while at customer sites, so I took a stab and documenting how it works for an @ionic/angular application. I am not sure who all should review this, so feel free to add someone else if you see fit.

src/pages/building/testing.md
```diff
@@ -7,21 +7,49 @@ nextUrl: '/docs/building/contributing'
...
 
 ## Testing Principles
 
-When testing the application, it is best to keep in mind that the goal of testing is not to verify that the code is correct but is to find problems with the code so bugs can be resolved earlier in the development cycle when fixes are less expensive. This is a subtle but important distinction. 
+When testing the application, it is best to keep in mind that the goal of testing is not to verify that the code is correct but is to find problems with the code so bugs can be resolved earlier in the development cycle when fixes are less expensive. This is a subtle but important distinction.
 
 If we set out to prove that the code is correct, we are more likely to stick to the happy path through the code. If we set out to find problems, we are more likely to more fully exercise the code and find the bugs that are lurking there.
```
bryplano: Would it be prudent to provide an example here to demonstrate the difference?

i.e. here's some basic code for our app - now, here's an example test where we're trying to prove that the code is correct (the "wrong" way); finally, here is a test for the same code that attempts to find problems (the "right" way)

kensodemann: That would be really hard to show, actually. What it would most likely end up looking like is basically just a test with test cases missing, testing only the happy path and not the edge cases, not having a test that passes in a zero for something that eventually gets used as a denominator in a division operation, not having tests that show the code can handle out-of-bounds conditions gracefully, stuff like that.

 It is really more of a holistic mindset to have when approaching how to design your tests. Basically, your mindset when testing should be to try and break your code.

Maybe one or two of the above examples could be worked in... I will have to think about exactly how, though.

kensodemann: Maybe just adding something like "Your mindset when testing shold be to try and break your code." would be a good addition to that... 🤔

bryplano: No worries :)

My testing experience (esp. re: Angular) is limited, so this is something where my reaction was "Okay, that's great in theory, but how can the principle be applied in a practical sense i.e. what type of situation(s) should I watch out for?"

It could always be expanded upon later too!

kensodemann: I will give it some thought on how that can be expanded without writing a whole book. Even if you take something like the first principle listed here: https://www.geeksforgeeks.org/software-engineering-seven-principles-of-software-testing/ and read what they say there, each sentence in their explanation really could be expanded upon.

kensodemann: Pushed out a change that tries to clarify this a bit without getting too deep into the weeds (since you could write a whole book about testing principles, I have read a few, they are boring...)

bryplano: I'm happy with the changes :)

adamdbradley:
Great job, thanks @kensodemann!

Issue #463, opened 2019-02-14
Missing component demos


camwiegert:
There should be a basic demo for each component. Demos can be added creating a folder in `src/demos/api` with the same path as the page. (e.g. `src/demos/api/chip/index.html`)

Feel free to check off components that you've determined don't make sense to demo. (e.g. controllers)

---

- [x] action-sheet-controller
- [x] action-sheet
- [x] alert-controller
- [x] alert
- [x] anchor
- [x] app
- [x] avatar
- [x] back-button
- [x] backdrop
- [x] badge
- [x] button
- [x] buttons
- [x] card-content
- [x] card-header
- [x] card-subtitle
- [x] card-title
- [x] card
- [x] checkbox
- [x] chip
- [x] col
- [x] content
- [x] datetime
- [x] fab-button
- [x] fab-list
- [x] fab
- [x] footer
- [x] grid
- [x] header
- [x] icon
- [x] img
- [x] infinite-scroll-content
- [x] infinite-scroll
- [x] input
- [x] item-divider
- [x] item-group
- [x] item-option
- [x] item-options
- [x] item-sliding
- [x] item
- [x] label
- [x] list-header
- [x] list
- [x] loading-controller
- [x] loading
- [x] menu-button
- [x] menu-controller
- [x] menu-toggle
- [x] menu
- [x] modal-controller
- [x] modal
- [x] nav-pop
- [x] nav-push
- [x] nav-set-root
- [x] nav
- [x] note
- [x] picker-column
- [x] picker-controller
- [x] picker
- [x] platform
- [x] popover-controller
- [x] popover
- [x] progress-bar
- [x] radio-group
- [x] radio
- [x] range
- [x] refresher-content
- [x] refresher
- [x] reorder-group
- [x] reorder
- [x] ripple-effect
- [x] route-redirect
- [x] route
- [x] router-outlet
- [x] router
- [x] row
- [x] searchbar
- [x] segment-button
- [x] segment
- [x] select-option
- [x] select-popover
- [x] select
- [x] skeleton-text
- [x] slide
- [x] slides
- [x] spinner
- [x] split-pane
- [x] tab-bar
- [x] tab-button
- [x] tab
- [x] tabs
- [x] text
- [x] textarea
- [x] thumbnail
- [x] title
- [x] toast-controller
- [x] toast
- [x] toggle
- [x] toolbar
- [x] virtual-scroll

cdarken:
[...]

PR #492, opened 2019-02-25
doc(backdrop): add backdrop example

mhartington:
Ref #463


brandyscarney:
The backdrop demo doesn't show anywhere? Am I missing it?

PR #514, opened 2019-03-05
React vue docs

jthoms1:
Update to Add Vue and React docs.  This includes a change to use Prismjs over Highlight.  Highlight does not support TSX so we are going to use Prismjs in place of it.


jthoms1:
After chatting with Cam I am going to remove the bash language from Prismjs and do it custom so that we can get the highlighting we expect for shell commands.

PR #532, opened 2019-03-07
docs(): add config class example

mhartington:


brandyscarney:
Could we maybe make two headings -

### Globally

#### Basic Example

#### Customizing Animations

### By Component

#### Basic Example

`< your example >`

thoughts?

mhartington:
I like it. Fixed

brandyscarney:
:tada:

PR #549, opened 2019-03-12
added capacitor live-reload via the cli paragraph

pevers:
Paragraph was missing for capacitor.


jcesarmobile:
capacitor command doesn't run the app, just starts the server and opens Xcode, so I think that should be mentioned.

Also the first paragraph says
> The Ionic CLI can build, copy, and deploy Ionic apps to iOS simulators and devices with a single command.
Which is not true for capacitor, so I think the cordova command should go first, and then the capacitor command with some note saying that it doesn't build/run the project, but opens Xcode so the user manually runs it instead.

dotNetkow:
Thanks for the PR - apologies that it sat for so long. We've clarified and [added more details](https://ionicframework.com/docs/building/ios#live-reload-with-capacitor) about live-reload with Capacitor.

PR #557, opened 2019-03-14
Update css-utilities.md

tomavanc:
**Flex Item Properties** 
HTML example code changed to the new CSS utility classes


brandyscarney:
Thanks so much!

PR #601, opened 2019-03-28
Update testing.md

daniellepintz:
I think it is a good idea to explain what Jasmine is when mentioning it. When I first read it I thought Jasmine was a person who had a funny name for mock objects ;)


brandyscarney:
Thank you :)

PR #606, opened 2019-03-30
fix(ionic-native) missing bracket in example for @ionic-native/launch-review

zaarheed:


camwiegert:
@zaarheed Thanks for the PR! We actually pull this data from the [ionic-native repo](https://github.com/ionic-team/ionic-native/blob/master/src/%40ionic-native/plugins/launch-review/index.ts#L21), so you'll need to open a PR there to make this change.

PR #607, opened 2019-04-01
feat(color-generator):  add function of share generated color.

rdlabo:
Add function of share generate color. 
I think this makes the document easy to use for development communication.

1) Color Generator reflect URL Params.
For example: `https://ionicframework.com/docs/theming/color-generator/?primary=#3880ff&secondary=#0cd1e8&tertiary=#7044ff&success=#10dc60&warning=#ffce00&danger=#f04141&dark=#222428&medium=#989aa2&light=#f4f5f8` .

2) Add URL reflect button.
When click this button, add color params to URL by history API.

スクリーンショット 2019-04-01 19 17 08

PR #611, opened 2019-04-02
feat:  Function of share genalate color.

rdlabo:


rdlabo:
sorry. same https://github.com/ionic-team/ionic-docs/pull/607

PR #624, opened 2019-04-10
Wrong method name

mvrkrishnaverma:
Wrong method name Canactive it should be CanActivate


brandyscarney:
Thank you for the PR! Sorry I didn't see this earlier, I have merged it in with: https://github.com/ionic-team/ionic-docs/pull/655

Issue #594, opened 2019-03-25
Ionic v4 layout documentation is not updated


anidotnet:
**URL**
https://ionicframework.com/docs/layout/structure

**What is missing or inaccurate about the content on this page?**

The usage code for layout structures are not updated and it is creating problem while learning the framework. Would highly appreciate it if it is updated appropriately.

brandyscarney:
Thanks for the issue! Which part is not updated?

anidotnet:
I have only seen 2, after that I stopped reading as I suspect the document is not updated. The 2 places I found are as follows:

1. https://ionicframework.com/docs/layout/structure#tabs-layout

```html
<ion-tab tab="home">
  <h1>Home Content</h1>
</ion-tab>
<ion-tab tab="settings">
  <h1>Settings Content</h1>
</ion-tab>
```
`ion-tab` is not supported in v4.

2. https://ionicframework.com/docs/layout/structure#menu-layout

```html
<ion-page class="ion-page" main>
  <ion-header>
    <ion-toolbar>
      <ion-buttons slot="start">
        <ion-menu-toggle>
          <ion-button>
            <ion-icon slot="icon-only" name="menu"></ion-icon>
          </ion-button>
        </ion-menu-toggle>
      </ion-buttons>
      <ion-title>Header</ion-title>
    </ion-toolbar>
  </ion-header>
  <ion-content padding>
    <h1>Main Content</h1>
    <p>Click the icon in the top left to toggle the menu.</p>
  </ion-content>
</ion-page>
```

`ion-page` is not supported in v4 it seems.

brandyscarney:
Thanks for the quick reply! The examples are written in vanilla JavaScript and it looks accurate for the tabs example: https://codepen.io/brandyscarney/pen/aMMrPK

You're right that `ion-page` does not exist anymore, however it should still work with that markup since the `ion-page` class is there: https://codepen.io/brandyscarney/pen/GeeaLR

Are you looking for this documentation for a specific framework? If so, could you please let me know which one? 🙂

anidotnet:
I was looking to create a mobile application using angular 7 and ionic 4. When I used these markup in my ionic + angular project, the page is not getting rendered in the browser.

In developer console I am seeing below error message:

```
ERROR Error: Uncaught (in promise): Error: Template parse errors:
'ion-tab' is not a known element:
1. If 'ion-tab' is an Angular component, then verify that it is part of this module.
[...]

PR #628, opened 2019-04-12
Fix: Tabs Example not Working

xmarkclx:
ion-tab is no longer needed and is deprecated(?).
See https://github.com/ionic-team/ionic/blob/master/CHANGELOG.md#angular-tabs


brandyscarney:
Thank you for the PR! The layout usage examples are written using vanilla HTML / JavaScript and not a specific framework. It is on our list to add more usage examples for this section, please see my comment here for more info: https://github.com/ionic-team/ionic-docs/issues/594#issuecomment-476395835

PR #631, opened 2019-04-16
fix(color-gen): hold generated color on mode changed.

rdlabo:
Now, if change device mode on demo, generated color will be initialized.
This pull request make hold generated color.


rdlabo:
Sorry. I missed.

PR #656, opened 2019-04-25
feat(installation): change install to using `ng add`

rdlabo:
`ng add` is not only add package. It will handle these.

```
CREATE src/theme/variables.css (288 bytes)
UPDATE package.json (1375 bytes)
UPDATE src/app/app.module.ts (466 bytes)
UPDATE angular.json (5889 bytes)
```

I think `ng add` is better.


rdlabo:
replace https://github.com/ionic-team/ionic-docs/issues/657

PR #691, opened 2019-05-11
Update native.json

indatawetrust:
link fixed

### DIFF
--- a/scripts/data/native.json
+++ b/scripts/data/native.json
@@ -74,7 +74,7 @@
   {
     "packageName": "@ionic-native/admob-pro",
     "displayName": "AdMob Pro",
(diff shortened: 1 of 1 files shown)


dotNetkow:
Hi, thanks for the PR. As it turns out, the JSON file you updated is auto-generated and pulled in from the [Ionic Native repo](https://github.com/ionic-team/ionic-native). Updating it here would mean it gets overwritten in the future. To actually update the source file, please [submit a PR for this line](https://github.com/ionic-team/ionic-native/blob/master/src/%40ionic-native/plugins/admob-pro/index.ts#L106).

Apologies that is not more obvious.

brandyscarney:
Closing as the fix needs to be made in the other repo. If you'd like to submit a PR there for us to review we can, thank you!

PR #692, opened 2019-05-13
Update starting.md

raniaSayed:


dwieeb:
@raniaSayed These are Ionic 3 templates and won't work with the latest Ionic CLI. The list of v3 starters in the v3 docs: https://ionicframework.com/docs/v3/cli/starters.html

Issue #734, opened 2019-06-07
404 when clicking "Next" in intro/concepts page


eedrah:
**Describe the bug**
"Next" page on Concepts is Not Found

**To Reproduce**
1. Go to https://ionicframework.com/docs/intro/concepts
2. Scroll down to the bottom and click 'Build your next app'
3. See 404

**Expected behavior**
Page is loaded

eedrah:
PR made

PR #735, opened 2019-06-07
Fix broken "next" link for intro/concepts

eedrah:
Solves #734 

Doesn't solve the issue that on this page that is loaded, the following "next" link doesn't go back to the intro that you left from. But at least the 404 is fixed.


dotNetkow:
thank you!

PR #744, opened 2019-06-12
Clarified Cordova support for Vue and React

dotNetkow:


src/pages/react/overview.md
```diff
@@ -27,4 +27,10 @@ Coming soon.
...
 
-Coming soon.
\ No newline at end of file
+Coming soon.
+
+## Native Tooling
+
+[Capacitor](https://capacitor.ionicframework.com) is the official cross-platform app runtime used to make your `Ionic React` web app run natively on iOS, Android, Electron, and the web. 
+
+While there are no known technical limitations to using `Ionic React` with [Cordova](https://cordova.apache.org/) plugins, Capacitor is officially recommended. **Note:** There are no plans to support `Ionic React` in the [Ionic CLI tooling](/docs/cli) at this time. For more details, please [see here](https://capacitor.ionicframework.com/docs/cordova).
```
jcesarmobile: Should be something like
There are no plans to support Cordova integration for `Ionic React` in the [Ionic CLI tooling](/docs/cli) at this time.
The CLI will support Ionic React, but you can only enable Capacitor integration, not Cordova

brandyscarney:
A couple things on this sentence:

> **Note:** There are no plans to support a Cordova integration for `Ionic React` in the [Ionic CLI tooling](/docs/cli) at this time.

1) I don't think we need to say `**Note:**`. We're already in the paragraph, it's not really a side note. 
2) I think the verbiage could be improved, "There are no plants to support" just sounds a little abrasive. 

I'm not sure of an alternative though on 2, so maybe it's fine.

dotNetkow:
@brandyscarney - removed the Note part, good call. gonna ship the rest as-is.

PR #804, opened 2019-07-03
Remove closing img tag

codeherk:
img tag is a void element. The closing tag causes template parsing errors.


dotNetkow:
nice catch! thanks.

PR #815, opened 2019-07-09
Update theming.md

mergebandit:
Using `Ionic 4.6.0`, the `IonicModule` typedef is the following: 

```
export declare class IonicModule {
    static forRoot(config?: IonicConfig): ModuleWithProviders;
}
```

As a result, the change outlined in the existing docs results in the following error:
```
ERROR in src/app/app.module.ts(19,5): error TS2554: Expected 0-1 arguments, but got 2.
```

The proposed change to the docs only passes the config object.


dotNetkow:
nice catch! thank you.

PR #817, opened 2019-07-10
revert: color generator description change

brandyscarney:
This makes no sense in the context of a non-angular or non-scss app. I don't think it should be here at all. If anything we can have a separate section that goes into more detail of where you can put this code or what `:root` means, but it doesn't *have* to go into the variables file either.


dotNetkow:
Sorry about that, Brandy! that makes sense.

PR #822, opened 2019-07-13
Update angular navigation link

bleuscyther:
the old link is not available


dotNetkow:
good catch, thank you!

PR #829, opened 2019-07-17
docs(theming): split out colors and themes from advanced

brandyscarney:
This separates out the following headings:

https://ionicframework.com/docs/theming/advanced#colors
https://ionicframework.com/docs/theming/advanced#themes

Into their own sections in the theming documentation. 

Screen Shot 2019-07-17 at 11 57 11 AM
Screen Shot 2019-07-17 at 11 57 18 AM

@bensperry and @dotNetkow can you please review the documentation here?


src/components.d.ts
```diff
@@ -151,11 +151,6 @@ export namespace Components {
 
 declare global {
 
-  // Adding a global JSX for backcompatibility with legacy dependencies
```
dotNetkow: this deletion keeps happening when i 'npm run start' locally. I remove it - should you?

brandyscarney: It shouldn't be there in the first place, I'm not sure how it got added back but it was removed by Stencil.

PR #831, opened 2019-07-18
fix(): use modern entry-points for ionic

manucorporat:
# Moves demos to point to jsdeliv CDN
- Version included in URL allows for strong caching
- Stackblitz, codesanbox (for angular, core and react) will use this same CDN. Also production ready sites could use it too, we maximize cache hit by using a CDN in this case

# Fix some warning

@Prop() id conflicts with HTML id, renamed to videoId
@Prop() onSelect coverted into a @Event 

# Update to Stencil 1.2.0-1
- Enable `excludeUnusedDependencies` helps to reduce from 25KB to 4KB the size of the site's JS entry-point
- Makes bundling in dev and production mode much much faster 
 (incremental build around 300ms)


perrygovier:
Review App in Heroku
https://ionic-docs-pr-831.herokuapp.com/docs/

PR #858, opened 2019-07-30
docs(devapp): add ionic cordova prepare command

Dellos7:
Added a note when explaining the use of the `ionic serve --devapp` command because when using the default project the following error is triggered: Could not find cordova integration in the default project. and we need to run `ionic cordova prepare` in order to be able to use the devapp


src/pages/appflow/devapp.md
```diff
@@ -36,6 +36,8 @@ First, ensure you are running the latest release of the Ionic CLI. Run `npm inst
 
 Once the latest CLI is installed, run `ionic serve --devapp` in the app of your choice on your computer and let it finish building. Next, open your iOS or Android device and connect to _the same network_ as your computer (through Wi-Fi). Open the DevApp, and you should see your local app show up in the list.
 
+> Note that currently you might need to run `ionic cordova prepare` with th default project before running `ionic serve --devapp` in order to copy the cordova assets to prepare the native build for the mobile platform.
```
dwieeb: typo: th -> the

dotNetkow:
thanks for the PR!

PR #875, opened 2019-08-09
chore(build): update native data

stephanejais:


stephanejais:
@mlynch Hi, being new here and having looked around I can't figure out how to trigger the CI checks. 

The goal of this PR is to update the ionic native docs, and I've seen more PR's with the exact same goal. Are there plans to update the ionic native docs and its website with the latest community plugins?

Kind regards.

dotNetkow:
Hi, sorry this sat so long! It's a bit of a manual process to update this list. It appears that [we added it](https://ionicframework.com/docs/native/wonderpush) in a separate update.

PR #888, opened 2019-08-18
Created first app guide landing page

dotNetkow:
Introduce option for user to select their JS framework of choice, then redirect them to the respective guide.


dotNetkow:
image

PR #894, opened 2019-08-20
update ionic-native doc

srinidhirao:
This is basically to get my plugin (unvired-cordova-sdk) published at ionic native site. Please check and merge this pull request so that the developers can have access to documentation straight from the ionic site! 

I executed the script located at ionic-native (docs-json.sh) to generate native.json file. I hope this is the right way.

Thanks.


dotNetkow:
Apologies this sat for so long - looks like your plugin [was added](https://ionicframework.com/docs/native/unvired-cordova-sdk)at a different time. Cheers!

PR #902, opened 2019-08-22
docs(platform): expand variable name

rtpHarry:
I'm not sure if its official but I have inferred that with Ionic4 names are expanded like `ctrl` to `controller`. This shortening of `platform` to `plt` doesn't seem in line with this.


dotNetkow:
I think it's just a preference. That said, looks fine to me. thanks!

PR #903, opened 2019-08-22
docs(css-variables): section name different

rtpHarry:
updated to reflect what the name of the section is on the page - easily inferred but I've opened it because i spotted it and its easier to just offer up the pr to see if you are bothered about this level of specificity.


dotNetkow:
good call! thank you

PR #934, opened 2019-09-06
Wrapperd the home component in an <IonPage>.

aviyadeveloper:
Without wrapping the content in <IonPage> the display for the <IonFab> component is broken, as it is pushed by the <IonHeader> component  and is only partly visible and mostly off screen. By adding the <IonPage> wrapper this problem is fixed and so the tutorial should be changed to accommodate it.

I fixed the code sinppets in the relevant parts of the guide.

### DIFF
--- a/src/pages/react/your-first-app.md
+++ b/src/pages/react/your-first-app.md
@@ -128,12 +128,12 @@ Currently, the `Home` component looks like so:
 ![React home component](/docs/assets/img/guides/react/first-app/home-route.png)
 
 ```typescript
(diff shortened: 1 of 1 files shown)


aviyadeveloper:
Was done in response to the following issue:
https://github.com/ionic-team/ionic/issues/19068

dotNetkow:
Thank you for the PR! Really appreciate it.

Looks like we made similar changes as part of our Ionic React launch. Closing as the content basically matches what you proposed here.

PR #542, opened 2019-03-08
docs(): add dark mode theming guide

mhartington:

PR #940, opened 2019-09-06
feat(theming): add guide for dark mode to theming

brandyscarney:
This adds documentation for dark theming, based on Mike's PR here: https://github.com/ionic-team/ionic-docs/pull/542

It has been updated to reflect the latest Ionic Dark Theme Ben and I have put together on Ionic and include how to combine the media query and fallback to write the variables once and use with toggles in an application.

`Co-authored-by: Mike Hartington <mikehartington@gmail.com>`


src/components/menu/templates/main.tsx
```diff
@@ -44,6 +44,7 @@ const items = {
     'menu-theming-css-variables': '/docs/theming/css-variables',
     'Colors': '/docs/theming/colors',
     'Themes': '/docs/theming/themes',
+    'Dark Mode': '/docs/theming/dark-mode',
```
dotNetkow: Perry's translations project is now live, so this token should change to match the others. here, it'd be "menu-theming-dark-mode" instead of "Dark Mode"

src/pages/theming/dark-mode.md
```diff
@@ -0,0 +1,254 @@
...
+
+body.dark {
+  /* dark mode variables go here */
+}
+```
+
+Notice that the variables above are only in the `body.dark` selector now, and the media query has one variable set. Now, in the JavaScript, the class can be added to the body based on the existence of this variable. This function should be called on load of an application, based on the framework being used:
+
+```javascript
+function onLoad() {
```
dotNetkow: this code block and the next are basically the same, minus moving code to the `toggleTheme` function. Normally I'm a fan of teaching/explaining more, but since they are so similar in this case, I suggest removing the "This could be improved upon by adding a toggle" line and just outright using one code block that has the toggleTheme func.

src/pages/theming/dark-mode.md
```diff
@@ -0,0 +1,254 @@
...
+}
+```
+
+This could be improved upon by adding a toggle in the application that calls a `toggleTheme()` method to toggle the class:
+
+```javascript
+function onLoad() {
+  const theme = getComputedStyle(document.body).getPropertyValue('--ion-color-scheme').trim();
+
+  if (theme === 'dark') {
```
dotNetkow: any reason to check for "dark" specifically vs. passing it as a variable? ex:

```
toggleTheme() {
    const theme = getComputedStyle(document.body).getPropertyValue('--ion-color-scheme').trim();
    if (theme) {
      document.body.classList.toggle(theme);
    }
  }
```

src/pages/theming/dark-mode.md
brandyscarney: Done :+1:

src/pages/theming/dark-mode.md
brandyscarney: No reason other than making it more clear what is going on.

src/components/menu/templates/main.tsx
brandyscarney: Just curious why isn't that on the others?

dotNetkow: haha not sure. they must have been missed, i guess

brandyscarney: Spoke with Perry and this has to be added for English only and then the translate service handles the rest, so added it.

src/pages/theming/dark-mode.md
```diff
@@ -0,0 +1,267 @@
...
+
+```css
+@media (prefers-color-scheme: dark) {
+  :root {
+    /* dark mode variables go here */
+  }
+}
+```
+
+Currently, the `prefers-color-scheme` media query has [limited browser support](https://caniuse.com/#feat=prefers-color-scheme), so users will not be able to benefit from having the dark mode applied using this media query in certain browsers. However, the dark mode can still be applied by using a [class fallback](#class-fallback).
```
mhartington: Maybe we should denote here we mean a css class, not a JS/TS class. Just a thought.

PR #984, opened 2019-10-01
Update CDN version

arjunyel:


dotNetkow:
Thanks @arjunyel! If possible, I'd like to not have to update this for each new release.

@brandyscarney, does "@latest" work for these links? For example [this](https://cdn.jsdelivr.net/npm/@ionic/core@latest/dist/ionic/ionic.esm.js). It appears to work fine, loading a minified version too.

brandyscarney:
@dotNetkow I believe there are issues with using latest and it using a cached version, but it is nicer than updating the version each release.

dotNetkow:
thank you @arjunyel!

PR #1045, opened 2019-10-24
fix(menu): angular utility move to angular menu.

rdlabo:
Utilities:
https://ionicframework.com/docs/utilities/config
https://ionicframework.com/docs/utilities/platform

These are angular's guide.


dotNetkow:
Thanks for this PR! Now that Ionic React is released, we do plan to move Angular content around across all of the Framework Guide. We'll get to this PR when we're closer to other changes we're planning.

dotNetkow:
Thank you again! ありがとうございました

The team needed to make more changes, so we ended up merging a different PR.

PR #1049, opened 2019-10-24
Update api.md

barocsi:
Also you should focus on adding that how can an user change the variables later, like update-method  or channel. How can we change it after installing the Deploy plugin?


dotNetkow:
Thanks @barocsi! You are correct, we should document that.  In the meantime, look at your `package.json` - the variables are stored under "cordova-plugin-ionic".

PR #1053, opened 2019-10-25
Add properties section with "resume".

riddletd:
Added a properties section before methods, and added resume as the first property. We needed resume in our app, so we thought others could benefit from this doc update as well.

### DIFF
--- a/src/pages/utilities/platform.md
+++ b/src/pages/utilities/platform.md
@@ -25,20 +25,6 @@ export class MyPage {
 }
 ```
 
-## Properties
-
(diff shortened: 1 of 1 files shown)


dotNetkow:
Trevor, this is great! Thanks so much. While reviewing, I noticed 2 other event methods ([see here](https://github.com/ionic-team/ionic/blob/master/angular/src/providers/platform.ts)), so I added those. Cheers!

riddletd:
Thank you Matt! I like your username. Very clever!

PR #1057, opened 2019-10-29
Refactor Ionic Native component

dotNetkow:


src/pages/enterprise/email-composer.md
```diff
@@ -3,11 +3,30 @@ title: Email Composer
 template: enterprise-plugin
 version: 0.9.2
 minor: 0.9.X
+capacitorConfig:
```
dotNetkow: With this new design, we'd need to add this frontmatter onto each plugin page. this does mean that the <native-ent-install> component is no longer here - it's moved back into a Template.

jcesarmobile: how is this supposed to be generated?

jcesarmobile:
I think we should keep the `<native-ent-install plugin-id="plugin-name" variables=""></native-ent-install>` component, and any additional instructions such as preferences or manual edits to config files should be on the README plugin files.

dotNetkow:
OK - sounds good.

PR #1076, opened 2019-11-08
docs(enterprise): update auth connect file

jcesarmobile:


src/pages/enterprise/auth-connect.md
```diff
@@ -341,12 +360,53 @@ ___
...
+Event handler which can be overridden to handle successful logout events
 
 **Returns:** `void`
 
 ___
 
+___
+<a id="ivconfig"></a>
+
+###  IVConfig
```
dotNetkow: Isn't a bunch of this Identity Vault related, not Auth Connect?

src/pages/enterprise/auth-connect.md
```diff
@@ -589,6 +715,37 @@ ___
...
+
+### [1.3.2] (2019-11-01)
+
+
+
+### [1.3.1] (2019-10-29)
+
+
+
+### [1.3.0] (2019-10-09)
```
dotNetkow: details? Via Luke suggestion, `fix an issue to remove dependency on IdentityVault`?

src/pages/enterprise/auth-connect.md
dotNetkow: update: "they are interfaces that we had to redefine to allow for AuthConnect to be used without IV. Rather than including them we have copies of them."

dotNetkow: @jcesarmobile FYI ^^^^

PR #976, opened 2019-09-27
feat(animation): add animation and gesture docs

liamdebeasi:
closes https://github.com/ionic-team/ionic/issues/19528

TODO:
- [x] Update API methods to reflect latest changes
- [x] Update codepens with Ionic 4.10 (exports `index` with `createAnimations`)
- [x] Add gesture interface and types

PR #1084, opened 2019-11-13
chore(): clarify Angular Config, move Config and Platform to Angular subheader

liamdebeasi:
1. Moved Config and Platform (previously under "Utilities" subheader) to under the "Angular" subheader.
2. Clarified that Ionic Config in Angular is not reactive.
3. Provided an example for configuring apps on a per-component level in a reactive manner.
4. Updated wording to not mention Config service as it is deprecated.
5. Removed the animations example while I was in there. This functionality/example is covered in https://github.com/ionic-team/ionic-docs/pull/976.


dotNetkow:
Before merging, let's set up redirects (done outside of ionic-docs, we'll chat) for these two pages you are moving...

PR #1085, opened 2019-11-13
feat(): pwa sections for react and angular

mhartington:
Initial pass at react/angular dedicated PWA docs.


dotNetkow:
I noticed we've got the current Publishing -> [PWA page](https://ionicframework.com/docs/publishing/progressive-web-app) that is Angular focused. Should definitely keep the page there, but thoughts on making it more generic? Maybe just... after the intro paragraph, remove all content (since you moved it under the dedicated Angular page) then make two headers, one for Angular and React with links to those sub pages.

mhartington:
Updated the duplicated content. Replaced with cards for the dedicated sections.

PR #1096, opened 2019-11-15
docs(bug_report): consistent formatting and casing

rtpHarry:
I obviously have too much time on my hands but there were a few formatting discrepancies I noticed while submitting the last bug.


perrygovier:
Thanks!

PR #1106, opened 2019-11-24
Update your-first-app.md

TheOtherRealm:
Added the proper way to deal with icons - it took me a while to figure out that the problem was simply a lack of import - '`IonIcon`' is not anywhere else in the documentation!  It would be good if someone who actually knows whats going on could create some documentation for icons in react.  Till then, this should make it so people don't run into errors like I did.


dotNetkow:
Good catch and thank you for the PR! We are actively updating the React docs now. Appreciate the mention about icons - will do.

PR #1170, opened 2020-01-12
add necessary configs to make tutorial work

nikita-fuchs:
I know, splash screen auto hide is true by default, but with a fresh project it doesn't disappear for me.


dotNetkow:
Thanks for this, but we're intentionally keeping this strictly using Capacitor only. Config.xml is a Cordova concept. Also, for the cleartext portion, that shouldn't be a factor at that stage of the tutorial - page 6 deploying mobile - because no HTTP requests are made. The Live Reload page, maybe, since it boots up a local server. However, I think based on your assistance on that other issue, the --external flag will cover that. All set! Thanks again, really appreciate your attention to detail and logging these issues and PRs.

PR #1171, opened 2020-01-12
add missing "external" flag for android

nikita-fuchs:


dotNetkow:
investigating

dotNetkow:
This was the issue with Live reload - missing the `--external` flag. Thank you!

dwieeb:
In the future we may want to automatically forward the Ionic dev server port from device to computer, in which case `--external` wouldn't be necessary.

Issue #1179, opened 2020-01-15
Saving Photos Tutorial - Compile error due to typo


pluckrose:
**Describe the bug**
On this page https://ionicframework.com/docs/angular/your-first-app/3-saving-photos in the `convertBlobToBase64` function the variable definition is shown like this `blog: Blob` which makes the compiler fail as it should be `blob: Blob`.

It will result in the following error in terminal when compiling:
`[ng]     ERROR in src/app/services/photo.service.ts(87,26): error TS2552: Cannot find name 'blob'. Did you mean 'Blob'?`

**To Reproduce**
1. Go to https://ionicframework.com/docs/angular/your-first-app/3-saving-photos 
2. Scroll down to the second last example which describes `convertBlobToBase64`
3. You'll see the variable definition is wrong (blog instead of blob)

**Expected behavior**
The `blog` variable mentioned above should be `blob` and the compiler will work correctly.

**Screenshots**
<img width="1919" alt="Screenshot 2020-01-15 at 10 51 13" src="https://user-images.githubusercontent.com/9683367/72428258-30f8a100-3785-11ea-93ae-a7c896a3a704.png">

**Browser and OS (please complete the following information):**
 - All

**Solution**
- Change `blog` to `blob`
- Created PR https://github.com/ionic-team/ionic-docs/pull/1180

My firs public PR so aplogies if it's missing info.

dotNetkow:
Thank you for the time and attention to detail! Much appreciated. Closing since your PR fixed this.

PR #1180, opened 2020-01-15
Fix typo in variable

pluckrose:
*Fix*
Fixing typo that causes compiliation failure on Saving Photo Tutorial.

Changing blog to blob fixes this.

*Issue*
On this page https://ionicframework.com/docs/angular/your-first-app/3-saving-photos in the convertBlobToBase64 function the variable definition is shown like this `blog: Blob` which makes the compiler fail as it should be `blob: Blob.`

It will result in the following error in terminal when compiling:
`[ng] ERROR in src/app/services/photo.service.ts(87,26): error TS2552: Cannot find name 'blob'. Did you mean 'Blob'?
`

Created this issue for visibility https://github.com/ionic-team/ionic-docs/issues/1179


dotNetkow:
Great catch! thank you.

PR #1182, opened 2020-01-15
Update platform.md

bitlather:
isPlatform is available from @ionic/core, not @ionic/react.

Perhaps this is a version issue?

See https://github.com/ionic-team/ionic/issues/19122

### DIFF
--- a/src/pages/react/platform.md
+++ b/src/pages/react/platform.md
@@ -10,7 +10,7 @@ previousUrl: '/docs/react/config'
 The `isPlatform` method can be used to test if your app is running on a certain platform:
 
 ```typescript
-import { isPlatform } from '@ionic/react';
+import { isPlatform } from '@ionic/core';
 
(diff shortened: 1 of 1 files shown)


bitlather:
note: I did not test `import { getPlatforms } from '@ionic/react';` from that same documentation because I don't need it. Not sure if that is also in @ionic/core.

elylucas:
`IsPlatform` is actually exported from @ionic/react, and it is a wrapper around the isPlatform exported from core and automatically passes in the window reference.

Let me know if you can't obtain it from ionic/react for some reason.

bitlather:
> `IsPlatform` is actually exported from @ionic/react, and it is a wrapper around the isPlatform exported from core and automatically passes in the window reference.
> 
> Let me know if you can't obtain it from ionic/react for some reason.

My build failed when I tried to import from @ionic/react, but it passed when I imported from @ionic/core

PR #1187, opened 2020-01-16
Adding tab2.page.html step to display saved images

pluxain:


pluxain:
I see this step is part of the next section

PR #1199, opened 2020-01-24
Update 2-taking-photos.md

kiev78:
moved the install of PhotoService to prev. step since it was required there


dotNetkow:
thank you!

PR #1217, opened 2020-02-11
Update navigation.md

lukevp:
Add information on how to install the correct packages, this was not trivial information to find and should be in the component docs.


perrygovier:
This file is automatically generated. We'll need to update it in the https://github.com/ionic-team/ionic-native repo

PR #1218, opened 2020-02-11
fix(native): Fix Document Scanner plugin usage example

amoebiusss:
Fix function name in Document Scanner usage example, according to plugin docs:
https://github.com/NeutrinosPlatform/cordova-plugin-document-scanner#usage


perrygovier:
Hey @amoebiusss, Thanks and good catch. Unfortunately this file is autogenerated. Sorry, we should document this better.  You'll want to make this change to the ionic-native source here: https://github.com/ionic-team/ionic-native/blob/master/src/%40ionic-native/plugins/document-scanner/index.ts#L62

amoebiusss:
Hi @perrygovier. Thanks, I'll check it

PR #1255, opened 2020-02-23
docs(ion-segment): set unique values for segments

eikooc:
Sets the call segment's value to call to avoid overlapping values.
Overlapping values highlight multiple segments.


brandyscarney:
Thank you!

PR #1261, opened 2020-02-26
Add a notice about invalid code signature errors

jstjnsn:
Multiple people have reported issues on the capacitor github about invalid code signatures when running their app on an external device. As explained here: https://github.com/ionic-team/capacitor/issues/2458 this has to do with using a free Apple Developer account. I thought it would be a good idea to add a notice about this to the tutorial.


dwieeb:
I'm told this is supposedly fixed in later versions of Xcode?

dotNetkow:
Looks like no longer an issue with Xcode. If comes up again, let us know here. thanks.

PR #1286, opened 2020-03-11
fixed anonymous function syntax

0xflotus:
Could it be, this was a typo?


brandyscarney:
Thank you for the PR! Requesting a review from @liamdebeasi, looks like this is in a few places in this doc.

liamdebeasi:
Thanks for the PR! It looks like this was already resolved via https://github.com/ionic-team/ionic-docs/pull/1383. Closing.

PR #1322, opened 2020-03-25
Updates a command and a folder in the React PWA document

lscardinali:
In the Line 
"Once this package has been added run `ionic build --prod` and the `www` directory will be ready to deploy as a PWA."
In react, the --prod on the ionic build command doesn't seem needed and the directory that is generated to be deployed is "build" instead of "www"


dotNetkow:
thank you!

PR #1323, opened 2020-03-25
typo

andrewrains:
"since Ionic React managers"
to 
"since Ionic React manages"


dotNetkow:
thank you @andrewrains !

PR #1355, opened 2020-04-09
Make sure that the keytool command shown generates JKS over PKCS12

RoopeHakulinen:
As discussed in Ionic forum (https://forum.ionicframework.com/t/cant-submit-android-keystore/184737/6) the key needs to be in the JKS format which is not the default at least on JDK11 version of the keytool. If the keystore is not in JKS format you'll get an error message whilst uploading it to Ionic Appflow and the error message doesn't do anything to point you to the right direction.


dotNetkow:
Huge thanks @RoopeHakulinen for debugging this and opening a PR!

PR #1356, opened 2020-04-09
docs(hardware-back-button): add hardware back button documentation and examples

liamdebeasi:
blocked https://github.com/ionic-team/ionic/pull/20613


src/pages/developing/hardware-back-button.md
```diff
@@ -0,0 +1,202 @@
...
+  });
+}
+
+```
+</docs-tab>
+</docs-tabs>
+
+This examples shows the application exiting when the user presses the hardware back button and there is nothing left in the navigation stack. It is also possible to display a quit confirmation dialog before quitting the app.
+
+It is recommended to check whether or not the user is on the root page prior to exiting the application. Developers can use the `canGoBack` method on `IonRouterOutlet` in Ionic Angular. Support for this will be coming to Ionic React soon.
```
brandyscarney, line 194: Can we wait until we have this supported in React? We should have a React example

src/pages/developing/hardware-back-button.md
```diff
@@ -5,15 +5,15 @@ previousUrl: '/docs/developing/tips'
 
 # Hardware Back Button
 
-## Overview
-
 By default, when a user presses the Android hardware back button, the current view will be popped off the navigation stack. If there are no views to pop, nothing will happen. This guide will show you how to customize this behavior.
```
brandyscarney, line 8: Can we add a brief summary on what the back button is before jumping into this

liamdebeasi: I have https://github.com/ionic-team/ionic-docs/pull/1356/files#diff-118692f5d92350834ff00b783ab8591aR10 right after, should I move it to the top instead?

brandyscarney: I think something more like what the hardware back button is:

```
The hardware back button is found on most Android devices. In native applications it can be used to close modals, navigate to the previous view, exit an app, and more. By default in Ionic, when the back button is pressed, the current view will be popped off the navigation stack and the previous view will be displayed. If no previous views exist in the navigation stack, nothing will happen.  This guide will show to customize the behavior of the hardware back button.

> The hardware back button refers to the physical back button on an Android device and should not be confused with either the browser back button or `ion-back-button`. The information in this guide only applies to Android devices and is not relevant for iOS devices.
```

Feel free to change up my words, just threw it together.

liamdebeasi:
@brandyscarney Ok I made a few changes -- lmk what you think. I changed `Overview` to `Hardware Back Button in Capacitor and Cordova` -- having no heading resulted in a big block of text that was difficult to parse. Additionally, there's a `Hardware Back Button in a Browser` section, so I figured it made sense to have one for standalone apps.

PR #1358, opened 2020-04-11
Update imports from Angular Animation

### LINKED ISSUE #1277: import { Animation } should be from '@ionic/core' not from '@ionic/angular'
**URL**
The URL at which the content is missing or inaccurate
https://ionicframework.com/docs/utilities/animations#installation

I was trying to add an animation to my ionic 5 angular app following [this guide](https://ionicframework.com/docs/utilities/animations#installation).
It says to` import { Animation, AnimationController } from '@ionic/angular'; `but after building the app I get `error TS2305: Module '"../../../node_modules/@ionic/angular/ionic-angular"' has no exported member 'Animation'.`

It looks like `Animation` has to be imported from `'@ionic/core';` and not from` '@ionic/angular'`

byrw:
fixes #1277

### DIFF
--- a/src/pages/utilities/animations.md
+++ b/src/pages/utilities/animations.md
@@ -57,7 +57,8 @@ Developers using Angular should install the latest version of `@ionic/angular`. 
 
 ```typescript
 
-import { Animation, AnimationController } from '@ionic/angular';
+import { Animation } from '@ionic/core';
+import { AnimationController } from '@ionic/angular';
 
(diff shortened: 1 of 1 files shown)


jcesarmobile:
Thanks for the PR, but it was actually a bug in ionic/angular not re exporting the types, it was fixed long ago.
https://github.com/ionic-team/ionic-framework/pull/20766

PR #1393, opened 2020-04-30
Use PHOTO_STORAGE constant

felipevilabuilt:
In the loadSaved() method we can use the PHOTO_STORAGE constant already declared, instead of the 'photos' string.


dotNetkow:
great catch! thank you.

PR #1405, opened 2020-05-05
feat(parts): add shadow parts to the API docs

brandyscarney:
Blocked by https://github.com/ionic-team/stencil/pull/2412


brandyscarney:
Confirmed this works with the latest ionic dev build

Screen Shot 2020-06-09 at 7 56 22 PM

PR #1413, opened 2020-05-11
Add animationCtrl to the constructor in Angular

timKraeuter:
Add animationCtrl to the constructor of the ModalExample in Angular/Typescript otherwise, the code would not compile since animationCtrl  is used later.


timKraeuter:
Someone interested?

dotNetkow:
awesome! thanks. just added a missing import. cheers!

PR #1462, opened 2020-06-10
chore(build): update stencil to v1.14.0, update lint rules, enable strict mode

brandyscarney:
- Updates Stencil to `1.14.0`
- Updates tsconfig to match `ionic`
- Adds strict mode
- Updates to pass linter and build
- Adds `npm run lint.fix` command to auto fix lint errors where possible

The downgrade on fs-extra was intentional due to a bug with types, can look into upgrading it later, I'll have to find the issue again

Known issues:

- [x] Scrolling down a page then navigating to a new one does not scroll back to the top of the page
- [x] table-of-contents cannot find id of undefined error
- [x] isPremier showing when it shouldn't on native API pages


dotNetkow:
Brandy, can you elaborate on " isPremier showing when it shouldn't on native API pages" ? Looks good to me (perhaps that's why it's checked off - but just checking).

Looking great!

brandyscarney:
> Brandy, can you elaborate on " isPremier showing when it shouldn't on native API pages" ? Looks good to me (perhaps that's why it's checked off - but just checking).
> 
> Looking great!

Yeah I checked off the issues I found after I fixed them. Essentially on pages like this it was saying a premier plugin was available when it wasn't: https://ionic-docs-pr-1462.herokuapp.com/docs/native/secure-storage

I went through comparing many of the plugin pages on this PR to the live version and couldn't find any other differences after I fixed that.

PR #1508, opened 2020-06-25
Capacitor instructions

pedrohills:
Added capacitor instructions to docs.


dwieeb:
@pedrohills I would like to see these docs instead use the [`ionic capacitor build` command](https://ionicframework.com/docs/cli/commands/capacitor-build) (available since CLI 6.6.0).

For a production build, `ionic capacitor build <platform> --prod` will run a prod build of web assets and run `capacitor sync` (which is copy + update). You may also need to mention that after the build finishes, they need to run a native build in Xcode/Android Studio. This may be best documented by linking to instructions for Android/iOS.

jcesarmobile:
closing since the requested changes were not addressed and the pull request presents merge conflicts

PR #1530, opened 2020-07-16
docs(guide): add CSS Shadow Parts guide

### LINKED ISSUE #1343: theming: include documentation on shadow parts
**URL**
https://ionicframework.com/docs/theming/basics

I'm not sure where this should go

Add documentation on how to use shadow parts with this issue: https://github.com/ionic-team/ionic/issues/20916

brandyscarney:
Implements a guide for CSS Shadow Parts. Adapted from https://ionicframework.com/blog/customize-your-ionic-framework-app-with-css-shadow-parts/

Closes #1343

### DIFF
--- a/src/pages/theming/css-shadow-parts.md
+++ b/src/pages/theming/css-shadow-parts.md
@@ -65,7 +65,7 @@ With these parts exposed, the element can now be styled directly using [::part](
 
 The <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/::part" target="_blank" rel="noopener noreferrer">`::part()`</a> pseudo-element allows developers to select elements inside of a shadow tree that have been exposed via a part attribute.
 
-Since we know that the `ion-select` exposes a `placeholder` part for styling the text when there is no value selected, we can customize it in the following way:
+Since we know that `ion-select` exposes a `placeholder` part for styling the text when there is no value selected, we can customize it in the following way:
 
 ```css
 ion-select::part(placeholder) {
--- a/src/pages/theming/css-shadow-parts.md
+++ b/src/pages/theming/css-shadow-parts.md
@@ -98,7 +98,7 @@ ion-item::part(native):hover {
 
 ## Ionic Framework Parts
 
-All exposed parts for an Ionic Framework component can be found under the CSS Shadow Parts heading on its API page. To view all components and their  API pages, see the <a href="/docs/components" target="_blank" rel="noopener noreferrer">Component documentation</a>.
+All exposed parts for an Ionic Framework component can be found under the CSS Shadow Parts heading on its API page. To view all components and their API pages, see the <a href="/docs/components" target="_blank" rel="noopener noreferrer">Component documentation</a>.
 
 In order to have parts a component must meet the following criteria:
 
--- a/src/pages/theming/css-shadow-parts.md
+++ b/src/pages/theming/css-shadow-parts.md
@@ -11,7 +11,7 @@ CSS Shadow Parts allow developers to style CSS properties on an element inside o
 
 ## Why Shadow Parts?
 
-Ionic Framework is a distributed set of <a href="https://www.webcomponents.org/introduction" target="_blank" rel="noopener noreferrer">Web Components</a>. Web Components follow the <a href="https://w3c.github.io/webcomponents/spec/shadow/" target="_blank" rel="noopener noreferrer">Shadow DOM specification</a> in order to encapsulate styles and markup.
+Ionic Framework is a distributed set of <a href="https://developer.mozilla.org/en-US/docs/Web/Web_Components" target="_blank" rel="noopener noreferrer">Web Components</a>. Web Components follow the <a href="https://w3c.github.io/webcomponents/spec/shadow/" target="_blank" rel="noopener noreferrer">Shadow DOM specification</a> in order to encapsulate styles and markup.
 
 > Ionic Framework components are **not all** Shadow DOM components. If the component is a Shadow DOM component, there will be a badge in the top right of its <a href="/docs/components" target="_blank" rel="noopener noreferrer">component documentation</a>. An example of a Shadow DOM component is the <a href="/docs/api/button" target="_blank" rel="noopener noreferrer">button component</a>.
 
--- a/src/pages/theming/css-shadow-parts.md
+++ b/src/pages/theming/css-shadow-parts.md
@@ -1,0 +1,153 @@
+---
+previousText: 'CSS Variables'
+previousUrl: '/docs/theming/css-variables'
+nextText: 'Colors'
+nextUrl: '/docs/theming/colors'
+---
+
+# CSS Shadow Parts
+
+CSS Shadow Parts allow developers to style CSS properties on an element inside of a shadow tree. This is extremely useful in customizing Ionic Framework <a href="https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM" target="_blank" rel="noopener noreferrer">Shadow DOM</a> components.
+
+## Why Shadow Parts?
+
+Ionic Framework is a distributed set of <a href="https://www.webcomponents.org/introduction" target="_blank" rel="noopener noreferrer">Web Components</a>. Web Components follow the <a href="https://w3c.github.io/webcomponents/spec/shadow/" target="_blank" rel="noopener noreferrer">Shadow DOM specification</a> in order to encapsulate styles and markup.
+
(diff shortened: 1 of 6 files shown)


src/pages/theming/css-shadow-parts.md
```diff
@@ -0,0 +1,153 @@
+---
+previousText: 'CSS Variables'
+previousUrl: '/docs/theming/css-variables'
+nextText: 'Colors'
+nextUrl: '/docs/theming/colors'
+---
+
+# CSS Shadow Parts
+
+CSS Shadow Parts allow developers to style CSS properties on an element inside of a shadow tree. This is extremely useful in customizing Ionic Framework <a href="https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM" target="_blank" rel="noopener noreferrer">Shadow DOM</a> components.
```
liamdebeasi, line 10: Should we define what a "shadow tree" or "shadow root" is?

src/pages/theming/css-shadow-parts.md
```diff
@@ -0,0 +1,153 @@
...
+
+The above shows two parts: `placeholder` and `icon`. See the <a href="/docs/api/select#css-shadow-parts" target="_blank" rel="noopener noreferrer">select documentation</a> for all of its parts.
+
+With these parts exposed, the element can now be styled directly using [::part](#how-part-works).
+
+### How ::part works
+
+The <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/::part" target="_blank" rel="noopener noreferrer">`::part()`</a> pseudo-element allows developers to select elements inside of a shadow tree that have been exposed via a part attribute.
+
+Since we know that the `ion-select` exposes a `placeholder` part for styling the text when there is no value selected, we can customize it in the following way:
```
liamdebeasi, line 68:
```suggestion
Since we know that `ion-select` exposes a `placeholder` part for styling the text when there is no value selected, we can customize it in the following way:
```

liamdebeasi: or do "the `ion-select` component"

src/pages/theming/css-shadow-parts.md
```diff
@@ -0,0 +1,153 @@
...
+  color: green;
+}
+```
+
+> There are some known limitations with [vendor prefixed pseudo-elements](#vendor-prefixed-pseudo-elements) and [structural pseudo-classes](#structural-pseudo-classes).
+
+
+## Ionic Framework Parts
+
+All exposed parts for an Ionic Framework component can be found under the CSS Shadow Parts heading on its API page. To view all components and their  API pages, see the <a href="/docs/components" target="_blank" rel="noopener noreferrer">Component documentation</a>.
```
liamdebeasi, line 101:
```suggestion
All exposed parts for an Ionic Framework component can be found under the CSS Shadow Parts heading on its API page. To view all components and their API pages, see the <a href="/docs/components" target="_blank" rel="noopener noreferrer">Component documentation</a>.
```

src/pages/theming/css-shadow-parts.md
```diff
@@ -0,0 +1,153 @@
...
+}
+```
+
+### Chaining Parts
+
+The `::part()` pseudo-element can not match additional `::part()`s.
+
+For example, `my-component::part(button)::part(label)` does not match anything. This is because doing so would expose more structural information than is intended.
+
+If the `<my-component>`’s internal button uses something like `part="label => button-label"` to forward the button’s internal parts up into the panel’s own part element map, then a selector like `my-component::part(button-label)` would select just the one button’s label, ignoring any other labels.
```
liamdebeasi, line 153: I haven't seen this syntax before `label => button-label`. Is there an MDN doc we can link to where people can read more?

src/pages/theming/css-shadow-parts.md
mhartington: I think it might be outside the context of this to go into detail, but maybe links to mdn or something.

src/pages/theming/css-shadow-parts.md
brandyscarney: I got it from Example 4 here: https://www.w3.org/TR/css-shadow-parts-1/#part

src/pages/theming/css-shadow-parts.md
brandyscarney: Yeah I previously had it linking to https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM but that same link is used in the next sentence so it seemed like too much. I can add it back there if we think it should be there.

liamdebeasi: Ah ok if there's already a link then it's probs fine as is.

brandyscarney: I'm not sure honestly. Do people get annoyed by too many links? haha. I wasn't sure if it would annoy someone that they both link to the same place.

src/pages/theming/css-shadow-parts.md
liamdebeasi: Ok that should be fine. I guess it just doesn't make a lot of sense to me - the spec doesn't give much information as to the `=>` syntax.

src/pages/theming/css-shadow-parts.md
liamdebeasi: The more I think about it, I agree with Mike. The term "shadow root" is not Ionic specific, and someone could pretty easily look up the definition (or click the other shadow dom link). Also, I think the expectation is that if you are using shadow parts, you should be familiar with the concept of the shadow dom in general.

src/pages/theming/css-shadow-parts.md
brandyscarney: I agree it's confusing. It's like the following:

```html
<my-component>
  #shadow-root
    <my-button part="button">
      #shadow-root
      <div part="label"></div>
    </my-button>
</my-component>
```

So this wouldn't work:

```css
my-component::part(button)::part(label) {
```

but if the button somehow forwards the part then it would work like:

```css
my-component::part(button-label)
```

I guess it would look like this:

```html
<my-component>
  #shadow-root
    <my-button part="button label => button-label">
      #shadow-root
      <div part="label"></div>
    </my-button>
</my-component>
```

I need to test this out though and then maybe I could add more usage. I don't think it applies to Ionic components right now.

liamdebeasi: Ohh I see ok that makes sense. Yeah I don't think we use this in the framework anywhere?

src/pages/theming/css-shadow-parts.md
brandyscarney: So no link?

src/pages/theming/css-shadow-parts.md
brandyscarney: Not right now. This could maybe end up being a problem for things like `ion-infinite-scroll-content`, for example, if we allowed them to target parts on the spinner.

src/pages/theming/css-shadow-parts.md
liamdebeasi: No link 👍

src/pages/theming/css-shadow-parts.md
```diff
@@ -0,0 +1,153 @@
...
+/* Does NOT work */
+my-component::part(scroll)::-webkit-scrollbar {
+  background: green;
+}
+```
+
+See <a href="https://github.com/w3c/csswg-drafts/issues/4530" target="_blank" rel="noopener noreferrer">this issue on GitHub</a> for more information.
+
+
+### Structural Pseudo-Classes
```
mhartington, line 131: Could we have an example just to show what _would_ work?

brandyscarney: For structural pseudo classes?

src/pages/theming/css-shadow-parts.md
```diff
@@ -0,0 +1,153 @@
...
+nextUrl: '/docs/theming/colors'
+---
+
+# CSS Shadow Parts
+
+CSS Shadow Parts allow developers to style CSS properties on an element inside of a shadow tree. This is extremely useful in customizing Ionic Framework <a href="https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM" target="_blank" rel="noopener noreferrer">Shadow DOM</a> components.
+
+## Why Shadow Parts?
+
+Ionic Framework is a distributed set of <a href="https://www.webcomponents.org/introduction" target="_blank" rel="noopener noreferrer">Web Components</a>. Web Components follow the <a href="https://w3c.github.io/webcomponents/spec/shadow/" target="_blank" rel="noopener noreferrer">Shadow DOM specification</a> in order to encapsulate styles and markup.
```
adamdbradley, line 14: Probably link to mdn web component docs: https://developer.mozilla.org/en-US/docs/Web/Web_Components

PR #1539, opened 2020-07-17
Add iOS details for adding a profile

pettys:
First time appflow user, and I got stuck here for a while. This addition
would have got me through this quickly, I think.


dotNetkow:
thanks Jason!

PR #1551, opened 2020-07-21
fix broken docs link

markgoho:


dotNetkow:
Thanks Mark, for fixing these links!

PR #1584, opened 2020-08-07
Fixed a broken link

shide1989:
MIgration guide is available here : 
https://ionicframework.com/docs/reference/migration


dotNetkow:
thank you! I made a slight tweak - we use relative paths, not full ones, so all that was missing was a slash in front of "docs":

"**/**docs/references/migration"

PR #1590, opened 2020-08-14
Update play-store.md

mikelhamer:
Update 404'd link


dotNetkow:
Thanks! It is "deploy" in the url - made that change.  (we now call the feature "Live Update").

mikelhamer:
> 
> 
> Thanks! It is "deploy" in the url - made that change. (we now call the feature "Live Update").

Very cool! Thanks!

PR #1610, opened 2020-08-23
docs: update to property binding syntax

santoshyadavdev:
Property binding syntax is more readable than the interpolation.


santoshyadavdev:
Hi Team, can we get it reviewed, please

dotNetkow:
Thanks. Only thing to note for future reference, is that I try to keep the [reference code repo ](https://github.com/ionic-team/photo-gallery-capacitor-ng)up to date too - I applied your suggestions over there 👍

PR #1616, opened 2020-08-25
Update 4-loading-photos.md

w3p706:
`loadSaved()` is async and sets the photos array to a new reference( `this.photos = JSON.parse(photos.value) || [];`, so the local photos in the Tab2Page remains emtpy.

Or as in the [Repo](https://github.com/ionic-team/photo-gallery-capacitor-ng/blob/master/src/app/tab2/tab2.page.html) reference the variable from the service directly in `tab2.page.html` [Step 2](https://ionicframework.com/docs/angular/your-first-app/2-taking-photos)

```
<ion-col size="6" 
    *ngFor="let photo of photos; index as position">
        <ion-img src="{{ photo.webviewPath }}"></ion-img>
</ion-col>
```

vs [Repo](https://github.com/ionic-team/photo-gallery-capacitor-ng/blob/master/src/app/tab2/tab2.page.html)
```
<ion-col size="6" 
    *ngFor="let photo of photoService.photos; index as position">
      <ion-img src="{{ photo.webviewPath }}" (click)="showActionSheet(photo, position)"></ion-img>
</ion-col>
```

### DIFF
--- a/src/pages/angular/your-first-app/4-loading-photos.md
+++ b/src/pages/angular/your-first-app/4-loading-photos.md
@@ -72,7 +72,9 @@ After, call this new method in `tab2.page.ts` so that when the user first naviga
 
 ```typescript
 ngOnInit() {
-  this.photoService.loadSaved();
+  this.photoService.loadSaved().then(() => {
+    this.photos = this.photoService.photos;
+  });
 }
 ```
 


dotNetkow:
Thanks for calling out the discrepancy. this has been fixed. cheers!

PR #1617, opened 2020-08-26
docs(navigation): add live example of angular routing

Coly010:
## Docs Addition
Add StackBlitz live example to Angular Navigation/Routing page to provide user with a playground to see routing in action.
The StackBlitz has some comments around lifecycle hooks also to help users understand when they are called.

Link to StackBlitz
https://stackblitz.com/edit/ionic-angular-routing

## Docs page screenshot
image

## Notes
The StackBlitz that is linked may need to be moved to an Ionic account on StackBlitz


mhartington:
Thanks!

PR #1618, opened 2020-08-26
docs(navigation): add live example for react routing

Coly010:
## Docs Addition
Add StackBlitz live example to React Navigation/Routing page to provide user with a playground to see routing in action.

Link to StackBlitz
https://stackblitz.com/edit/ionic-react-routing

## Docs page screenshot
image

## Notes
The StackBlitz that is linked may need to be moved to an Ionic account on StackBlitz


mhartington:
Thanks!

PR #1621, opened 2020-08-27
Grammar error 

raevenbauto:
I think there was a typo on the docs. I think it should be "our" instead of "are".


dotNetkow:
thanks!

PR #1636, opened 2020-09-02
Enterprise plugin updates

carlpoole:


carlpoole:
Removed changelog for InAppBrowser plugin prior to 3.0.0 since this is when we forked from Cordova.

PR #1639, opened 2020-09-04
Condense "Build Stacks" version table

ZaLiTHkA:
I found myself constantly scrolling down and up on this page, comparing different versions.. this change would simply make the page *easier* to read.

however, the validity of this change very much depends on the source for this file, is it generated by a script or updated manually each time there's a change?


nphyatt (approved):

dotNetkow:
Thanks for the PR! This does make it easier to read. Currently, this page will be updated manually.

PR #1654, opened 2020-09-13
Add call to photos variable from photoService

tharindusathis:
There is no any variable `photos` in relevant component. So it must be referred from `photoService`.


lokeshrmitra (approved):
I too had to add this myself

dotNetkow:
Yes, good catch. I made some other changes the other day, missed this part. thanks!

PR #1684, opened 2020-09-30
CT-186 update build stack docs with more info

lm3m:
* the UI for eol build stacks is still in progress, an image of what that looks like will be added when that feature is complete

### DIFF
--- a/src/pages/appflow/build-stacks.md
+++ b/src/pages/appflow/build-stacks.md
@@ -7,7 +7,7 @@ nextUrl: "/docs/appflow/web-previews"
 ---
 
 # Build Stacks
-In Appflow build stacks are how we represent the set of software used to build an app on a given platform. This includes the version of npm, Cordova, Ionic, Xcode, and what OS it is all installed on.
+In Appflow, build stacks are how we represent the set of software used to build an app on a given platform. This includes the version of npm, Cordova, Ionic, Xcode, and what OS it is all installed on.
 
 In general the latest build stack is the best choice for most apps. Though, we do have other build stacks available for scenarios where they are needed. A non-default build stack might be needed when a specific version of a build tool is required, for example an older version of Xcode or the Android build tools are needed to support some configuration. 
 
--- a/src/pages/appflow/build-stacks.md
+++ b/src/pages/appflow/build-stacks.md
@@ -7,8 +7,15 @@ nextUrl: "/docs/appflow/web-previews"
 ---
 
 # Build Stacks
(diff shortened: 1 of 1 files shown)


src/pages/appflow/build-stacks.md
```diff
@@ -7,8 +7,15 @@ nextUrl: "/docs/appflow/web-previews"
 ---
 
 # Build Stacks
+In Appflow build stacks are how we represent the set of software used to build an app on a given platform. This includes the version of npm, Cordova, Ionic, Xcode, and what OS it is all installed on.
```
devinshoemaker, line 10:
```suggestion
In Appflow, build stacks are how we represent the set of software used to build an app on a given platform. This includes the version of npm, Cordova, Ionic, Xcode, and what OS it is all installed on.
```

If we want to be truly grammatically correct then a comma should be added here.

src/pages/appflow/build-stacks.md
```diff
@@ -7,8 +7,15 @@ nextUrl: "/docs/appflow/web-previews"
...
 
-Below are the current versions used in Appflow for each build type.
+In general the latest build stack is the best choice for most apps. Though, we do have other build stacks available for scenarios where they are needed. A non-default build stack might be needed when a specific version of a build tool is required, for example an older version of Xcode or the Android build tools are needed to support some configuration. 
+
+All plans have access to the 'latest' Build Stacks, but only scale plans and above have access to the older builds.
+
+After a new build stack becomes available older build stacks for that platform will be labeled as to be retired for a year. After that it will no longer be available.
+
+Below are the versions used in Appflow for each build type..
```
devinshoemaker, line 18:
```suggestion
In general, the latest build stack is the best choice for most apps. Though, we do have other build stacks available for scenarios where they are needed. A non-default build stack might be needed when a specific version of a build tool is required, for example, an older version of Xcode or the Android build tools are needed to support some configuration. 

All plans have access to the 'latest' Build Stacks, but only scale plans and above have access to the older builds.

After a new build stack becomes available older build stacks for that platform will be labeled as to be retired for a year. After that, it will no longer be available.

Below are the versions used in Appflow for each build type...
```

And here are few more minor punctuation fixes.

src/pages/appflow/build-stacks.md
lm3m: yeah, this is why I like copy editors. commas and I don't get along.

devinshoemaker: Grammarly is my copy editor 😅

devinshoemaker (approved):

nphyatt (approved):

PR #1699, opened 2020-10-06
docs(camera): Removed dead Jira links

dominic-ks:
Just removing some old dead links to Apache Jira.


jcesarmobile (changes requested):
Apache jira is not dead, the problem is the links are missing the CB- before the issue number, probably caused by some bug on the tooling used to generate the cordova releasenotes file that we use to generate this file. This should be fixed on Apache file because this is automatically generated and will overwrite your changes on next regeneration

dominic-ks:
Right you are, in which case, that might be beyond me to fix at this point.

PR #1702, opened 2020-10-07
docs(animations): Add React and Angular examples for hooks animations

dillionmegida:
## Description

Add StackBlitz live example for [Before and After Hooks Animations](https://ionicframework.com/docs/utilities/animations#before-and-after-hooks)

Also changed the styles specified in the examples because of A11y reasons. `opacity` might not be easily perceived by some users.

## StackBlitz

Angular - https://stackblitz.com/edit/ionic-angular-before-after-hook-animations
React - https://stackblitz.com/edit/ionic-react-before-after-hook-animations

### DIFF
--- a/src/pages/utilities/animations.md
+++ b/src/pages/utilities/animations.md
@@ -53,7 +53,7 @@ const animation: Animation = createAnimation('')
 </docs-tab>
 <docs-tab tab="angular">
 
-Developers using Angular should install the latest version of `@ionic/angular`. Animations can be created via the `AnimationController` dependency injection. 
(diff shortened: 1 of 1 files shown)


brandyscarney:
Thank you for the PR! I have added these examples as playground demos to our docs via PR https://github.com/ionic-team/ionic-docs/pull/3035. You have been added as a co-author in the [merged commit](https://github.com/ionic-team/ionic-docs/commit/431ceccb8e4a711ea8e17516deaea52132fbbc730). I changed the animation styles to use a `filter` with `invert(75%)` and a `box-shadow` which should be easier than opacity to see. 🙂

PR #1709, opened 2020-10-10
Update lifecycle.md

peterennis:
typo


liamdebeasi (approved):
Good catch, thank you!

PR #1731, opened 2020-10-25
Cancel text is added in operating System label.

Gopal118:
Cancel button text was "Nah" so changed it to"Cancel".


brandyscarney:
Thanks for the PR! I am going to close this as this demo has been removed in favor of the new ones here: https://ionicframework.com/docs/api/select

PR #1755, opened 2020-11-06
Code example should show the new import

ericwhittaker:
In the code example, it needs to show the new import statement "import { RouteComponentProps } from 'react-router';" and remove the import statement that the previous code example had already introduced. This helps the user as they are following along so they can see how the import looks.


ericwhittaker:
No one looked at this and seems to not be worth addressing.

PR #1772, opened 2020-11-29
Update 3-saving-photos.md

aakatheeri:
In savePicture async function, the console gave me an error on "base64Data" string variable:
```console
error    'base64Data' is never reassigned. Use 'const' instead  prefer-const"
```

I fixed by remove the initialization part of base64Data, and make the base64Data variable as a const:
```typescript
// let base64Data: string;
const base64Data = await convertBlobToBase64(blob) as string;
```
The app works fine for me!

**update:** The original code works fine after I finish the whole tutorial.

**Platform Information**
- ionic/vue: 5.4.0
- vue: 3.0.0
- nodeJS: 6.14.8
- npm: 6.14.8
- OS: macOS 10.15.7


jcesarmobile:
closing as duplicate of https://github.com/ionic-team/ionic-docs/pull/1757

PR #1774, opened 2020-12-02
docs(intro): add missing nextText and nextUrl attributes for Development Tip page

boradakash:


brandyscarney:
Thank you for the PR! I am going to close this as we recently migrated the documentation and no longer use these, but we appreciate it!

PR #1788, opened 2020-12-10
Update links for cdn version of ionicons

nagashimam:
The current links for cdn version of ionicons return 404 error so their links should be updated.


brandyscarney:
Thank you for the PR! It looks like this was updated so I am going to close this. Sorry it wasn't updated by merging your PR, but we appreciate it.

PR #1797, opened 2020-12-23
Update app-store.md

zhangshengjie:
Add flag to disable sourceMapsh generation


liamdebeasi:
Thanks for the PR! This guide shows how to manually create a release build inside of Xcode which is why we do not use the `--release` flag in this example. However, I opened a new PR at https://github.com/ionic-team/ionic-docs/pull/2990 that mentions `--release` in case developers prefer to use that. I am going to close this in favor of the new PR, but I will give you co-author credit when the PR merges.

PR #1820, opened 2021-01-27
Update index.html

anmol1995-ionic:


brandyscarney:
Thank you for the PR! I am going to close this as we have moved the documentation files around which is causing conflicts in this PR, but if you think this is something we should add please create an issue for it. 🙂

PR #1836, opened 2021-02-10
Fixed typo

curious-monggo:
uses->use

From 'Below are some tips on uses cases for each of the life cycle events.'
To 'Below are some tips on use cases for each of the life cycle events.'


jcesarmobile:
thanks for the PR

the site structure changed, so I've sent a new PR making the change
https://github.com/ionic-team/ionic-docs/pull/2173

PR #1864, opened 2021-03-05
Fix import useRouter code in quickstart.md (Vue)

felixqueisler:
Fixed the code example for "importing useRouter from the vue-router package.". Before it was in the wrong place (outside <script>) and was the wrong statement.

It was probably a copy and paste error by a previous editor. The code example lacked the aforementioned import for useRouter but hat instead a redundant import for the add-icon in the wrong place.


liamdebeasi:
Thanks for the PR! Looks like this was already fixed via https://github.com/ionic-team/ionic-docs/pull/1881. You can see the changes live on https://ionicframework.com/docs/vue/quickstart.

PR #1887, opened 2021-04-07
Update debugging.md

dmedina2015:
Added footnote for those trying to debug `Ionic Vue` apps. Differently from `Angular` or `React` apps, `Ionic Vue` apps cannot be properly debugged just following the instructions in this page as is. Following directions from official [https://vuejs.org/v2/cookbook/debugging-in-vscode.html](Vue.js Docs), I added the `vue.config.js` with `devtool` property settled to `source-map`. Now, I can debug inside VS Code properly. I'm suggesting to update docs in order to save some time from the Vue community.


thetaPC (changes requested):
I'm still getting source map errors after following the new instructions with Ionic Vue. Please verify that the instructions are still valid. The [Vue docs](https://vuejs.org/v2/cookbook/debugging-in-vscode.html) link that was provided points to Vue 2 and these instructions may not be valid for the latest version.

PR #1917, opened 2021-05-14
docs(react): added fallback route to routing guide

eric-horodyski:


src/pages/react/navigation.md
```diff
@@ -97,6 +97,42 @@ The `DashboardPage` above shows a users list page and a details page. When navig
...
+    <IonRouterOutlet>
+      <Route exact path={match.url} component={UsersListPage} />
+      <Route path={`${match.url}/users/:id`} component={UserDetailPage} />
+      <Route render={() => <Redirect to={match.url} />} />
+    </IonRouterOutlet>
+  );
+};
+```
+
+Here, we see that in the event a location does not match the first two `Route`s it will redirect the Ionic React app to the `match.url` path.
```
liamdebeasi, line 120:
```suggestion
Here, we see that in the event a location does not match the first two `Route`s, `IonRouterOutlet` will redirect the Ionic React app to the `match.url` path. 
```

src/pages/react/navigation.md
```diff
@@ -97,6 +97,42 @@ The `DashboardPage` above shows a users list page and a details page. When navig
 
 An `IonRouterOutlet` should only contain `Route`s or `Redirect`s. Any other component should be rendered either as a result of a `Route` or outside of the `IonRouterOutlet`.
 
+## Fallback Route
+
+A common routing use case is to provide a "fallback" route to be rendered in the event the location navigated to does not match any of the routes defined. 
+
+We can define a route within an `IonRouterOutlet` to act as our "fallback" route -- provided no `path` is provided and it is the last `Route` defined.
```
liamdebeasi, line 104: "provided no path is provided" sounds a bit odd since we use "provided" twice. Can we change to make clearer?

liamdebeasi (changes requested):
Two small comments, otherwise looks great!

src/pages/react/navigation.md
liamdebeasi: I am assuming "it" refers to IonRouterOutlet. Let me know if that is not the case.

elylucas (approved):
agree with liams comments, but lgtm after that

src/pages/react/navigation.md
```diff
@@ -117,12 +116,12 @@ const DashboardPage: React.FC<RouteComponentProps> = ({match}) => {
 };
 ```
 
-Here, we see that in the event a location does not match the first two `Route`s it will redirect the Ionic React app to the `match.url` path. 
+Here, we see that in the event a location does not match the first two `Route`s. The `IonRouterOutlet` will redirect the Ionic React app to the `match.url` path.
```
liamdebeasi, line 119: This first part should be reworded because it is not a complete sentence.

liamdebeasi (changes requested):

src/pages/react/navigation.md
eric-horodyski: D'oh! Let me fix it 😄

liamdebeasi (approved):
Looks great!

PR #1932, opened 2021-05-25
Update 2-taking-photos.md

Applicafroguy:
Added quotation for @capacitor/storage imports


Applicafroguy:
Added quotation for @capacitor/storage imports

PR #1935, opened 2021-05-26
docs(): add performance docs for angular, react, and vue on using loops with ionic components

liamdebeasi:
Inspired by this issue in Framework: https://github.com/ionic-team/ionic-framework/issues/22775#issuecomment-849028954


NiklasMerz (approved):
That's really good advice. Thank you for your help

PR #1942, opened 2021-06-06
Update 2-taking-photos.md

stephanenoutsa:
Removes conflicting Photo declaration which causes app to break


jcesarmobile:
Thanks for the PR

The import is correct, there was an older interface that was renamed to UserPhoto, so there should be no conflict now

PR #2009, opened 2021-07-02
Update virtual-scroll.md

Christo-Deyzel:
I tried to install vue-virtual-scroll@next, but I got not found error. I think it should be "scroller"


jaredcbaum:
Great catch!

PR #2018, opened 2021-07-09
Use different toppings in Toggle doc

ryanwohara:
It appears there was a copy/paste error, or I misunderstand the code.


jaredcbaum:
Good catch!

PR #2066, opened 2021-09-08
Image format is required

I-am-abdulazeez:
from the return of the **savePicture** function, due to the new capacitor API


jcesarmobile:
thanks for the PR

the function returns UserPhoto now instead of Photo, so it should not include the format as UserPhoto doesn't have a format property

PR #2067, opened 2021-09-08
UserPhoto Interface

I-am-abdulazeez:
The Promise from the savePicture should return UserPhoto interface


jcesarmobile:
thanks for the PR

it has been fixed already in another PR

Issue #1384, opened 2020-04-27
docs: improve documentation surrounding using ionicons with React


brandyscarney:
**URL**
Either here: https://ionicons.com/

Or as a menu item in the main documentation 

**What is missing or inaccurate about the content on this page?**
We need to add more documentation surrounding using Ionicons with React to reduce confusion

https://github.com/ionic-team/ionic/issues/20353
https://github.com/ionic-team/ionic/issues/18847
& more

amandaesmith3:
Resolved via https://github.com/ionic-team/ionic-docs/pull/2073 and https://github.com/ionic-team/ionic-docs/pull/2074.

PR #2073, opened 2021-09-21
docs(react): add icons usage to quickstart

amandaesmith3:
Resolves https://github.com/ionic-team/ionic-docs/issues/1384

Loosely adapted from similar content in the Vue quickstart guide: https://ionicframework.com/docs/vue/quickstart#adding-icons

Global imports for icons don't seem to be a thing in React, so I left that piece out.

### DIFF
--- a/docs/react/quickstart.md
+++ b/docs/react/quickstart.md
@@ -354,6 +354,43 @@ return (
 
 Here, when we reload, if there is no app history present, we'll be able to navigate back to our home route.
 
+## Adding Icons
+
+Ionic React comes with (https://ionic.io/ionicons/) pre-installed. All you need to do is import the icon of your choice from the `ionicons` package, and pass it to an `IonIcon` component through the `icon` prop:
+
+```tsx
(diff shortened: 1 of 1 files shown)


docs/react/quickstart.md
```diff
@@ -354,6 +354,43 @@ return (
...
+  <IonContent>
+    <IonButton>
+      <IonIcon icon={camera} />
+      Take Picture
+    </IonButton>
+  </IonContent>
+};
+```
+
+Note that for React, we're passing the imported SVG reference, NOT the icon name as a string.
```
liamdebeasi, line 376:
```suggestion
Note that for React, we are passing the imported SVG reference, NOT the icon name as a string.
```

Small thing, but I usually avoid contractions for folks who do not speak English as their primary language.

liamdebeasi: Also for "NOT" can we make this bold rather than all caps?

liamdebeasi (approved):

PR #2090, opened 2021-11-01
Renaming instances of Photo to UserPhoto and CameraPhoto to Photo

marlon-ionic:
Two issues were addressed (for _angular_, _vue_, and _react_ docs):

1. There were several references to a `Photo` interface that was renamed `UserPhoto` in the current version of the sample project downloads. 
2.  There were also references to the deprecated Capacitor Camera Plugin's `CameraPhoto` Class. Renamed the references to use the new value of `Photo`.


jcesarmobile:
closing as a similar pull request was merged into main branch, which is the one used at the moment, so this is no longer necessary

PR #2113, opened 2021-12-05
docs(): deletes double 'the' typing

djakinho:
Hello, this is my first PR and I saw some 'the' duplicates in the docs that would be fixed. Thanks in advance.


jcesarmobile (approved):
looks good, thanks!

PR #2141, opened 2021-12-17
fixed overlapping of icons

talhafaroow33:
image


brandyscarney:
Thank you for the PR! This no longer appears to be an issue on the site so I am going to close this, but if I am missing something or there is an open issue please let me know.

Screen Shot 2022-09-19 at 5 14 14 PM

PR #2142, opened 2021-12-20
Update 3-saving-photos.md

JonathanAsbury-SPS:
Appears to be a case typo on 23.  Should be Filesystem instead of FileSystem.  Great tutorial by the way.  Covers a lot of ground quickly and in a fun way.


jcesarmobile:
Thanks for the PR

it has been fixed in another PR already

PR #2143, opened 2021-12-20
Update 3-saving-photos.md

JonathanAsbury-SPS:
The return type is not a Promise<Photo> so I've updated lines 31-34, now 31-36.


jcesarmobile:
Thanks for the PR

The function is async, that means it turns whatever you return into a Promise without the need of wrapping it into a new Promise, so the change is not needed.

PR #2151, opened 2022-01-03
Update 7-live-reload.md

papsl:
In step 2 (taking-photos), interface UserPhoto is created and not Photo, so we need to use it when implementing the Delete.


jcesarmobile:
Thanks for the PR

I've fixed this and a few others on https://github.com/ionic-team/ionic-docs/pull/2169

PR #2153, opened 2022-01-05
Return the correct type from `savePicture`

ElteHupkes:
The return statement for `savePicture` doesn't match the interface it states it will return. It returns `UserPhoto` rather than `Photo` (the type is correct on later pages).


jcesarmobile:
Thanks for the PR

I have fixed that and a few others in https://github.com/ionic-team/ionic-docs/pull/2169

PR #2156, opened 2022-01-07
Adds Missing Instruction in Vue Start Step

luiscape:
Step 4 mentions that we have to remove `Array<RouteRecordRaw>` from `router/index.js`. After doing that, we also need to remove the import for `RouteRecordRaw` because that's not used anywhere else. Without that `ionic serve` will throw the following error:

```shell
[vue-cli-service] 
[vue-cli-service] /Users/luis/Programs/mechanical-b/src/router/index.js
[vue-cli-service]   2:10  error  'RouteRecordRaw' is defined but never used  no-unused-vars
[vue-cli-service] 
[vue-cli-service] ✖ 1 problem (1 error, 0 warnings)
```


luiscape:
Is there a process to follow in order to get this PR merged? I'd be happy to close if not useful.

PR #2158, opened 2022-01-11
Update 3-saving-photos.md

jpretak:
It looks like the Promise should be set to expect the UserPhoto type by the time you've made these changes, and not the base Photo type as imported from Capacitor


jcesarmobile:
Thanks for the PR

I have fixed that and a few others in https://github.com/ionic-team/ionic-docs/pull/2169

PR #2168, opened 2022-01-20
docs(angular): Support note for Angular 13 with iOS 13

dtarnawsky:
To support iOS 13, Angular 13 requires targeting es5 rather than es2015.

### DIFF
--- a/docs/reference/browser-support.md
+++ b/docs/reference/browser-support.md
@@ -36,7 +36,7 @@ To figure out what version of the webview a device is running, log `window.navig
 
 ### A Note on Angular 13 Support
 
-The browser support policy for Angular 13 for iOS is the 2 most recent major versions which at time of release is iOS 15 and 14. To support iOS 13 a change is required to `tsconfig.json` in your project to target `es5` rather than the default of `es2015`. Without this change an error of `Unexpected token '.' in promiseReactionJob` will occur on startup of an app in iOS 13.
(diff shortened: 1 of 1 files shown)


docs/reference/browser-support.md
```diff
@@ -34,6 +34,10 @@ As a result, Ionic Framework v6 only supports Android devices and emulators runn
 
 To figure out what version of the webview a device is running, log `window.navigator.userAgent` to the console when inspecting the application using Chrome Dev Tools.
 
+### A Note on Angular 13 Support
+
+The browser support policy for Angular 13 for iOS is the 2 most recent major versions which at time of release is iOS 15 and 14. To support iOS 13 a change is required to `tsconfig.json` in your project to target `es5` rather than the default of `es2015`. Without this change an error of `Unexpected token '.' in promiseReactionJob` will occur on startup of an app in iOS 13.
```
kensodemann, line 39: Basically LGTM. But also I had to read it several times to grok it. I think the flow was not as x->y->z as I needed at this time of day. Maybe something like:

> Angular's support policy for iOS is the two most recent major versions. At the time of release that is iOS 14 and 15. To support iOS 13, change the project target specified in the `tsconfig.json` from the default value of `es2015` to `es5`. Without this change an error of `Unexpected token '.' in promiseReactionJob` will occur on app startup in iOS 13.

kensodemann (approved):

docs/reference/browser-support.md
dtarnawsky: Updated

liamdebeasi (approved):
Looks good! One non-blocking comment: Should this also appear on the v6 migration guide too?

dtarnawsky:
> Looks good! One non-blocking comment: Should this also appear on the v6 migration guide too?

Probably doesn't need it. Not many people need to support iOS 13 as all devices could be upgraded to iOS 14 before they got "stuck".

PR #2179, opened 2022-01-24
Fixing a mistake in the deletePicture definition

Junior-Lukusa:
There's a mistake in the **deletePicture** definition. 

We have to change the type of the **photo** argument from `photo: Photo`  to `photo: UserPhoto`


jcesarmobile (approved):
good catch, thanks!

PR #2185, opened 2022-02-07
feat(docs): playground docs component

sean-perkins:
Introduces the new `<Playground />` component for documenting Ionic Framework components in the docs site.

```jsx
import Playground from '../../src/components/global/Playground';

<Playground 
  src="/patch/to/demo/index.html"
  code={{
    basic: '/docs/path/to/html-code.md'
}} />
```

Supports:
- Source targets for: `basic` (HTML), `angular`, `react` and `vue`. 
- Toggling iOS/MD mode preview for Ionic component
- Generating Stackblitz example for HTML, Angular and React (still outstanding issue with Vue)

Preview path: [`/docs/playground/button`](https://ionic-docs-git-fw-714-ionic1.vercel.app/docs/playground/button)

Code within `/static/code/stackblitz` is required to generate a working Stackblitz app in the selected target. All other directories (i.e. `/static/code/button`) would be the replacement for our existing usage examples. 
## Other information

- Did not commit `package-lock.json` file, need to connect with the team to see what version of npm they are using (appears to be an outdated version)
- Had to downgrade prototype example from Docusaurus v2-beta to v1 to support ionic-docs example (main differences are how we handle code language formatting and static asset paths)
- Network requests aren't batched for usage examples, this would be a worthy optimization in the future.

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -259,11 +259,12 @@ export default function Playground({ children, src, code, title, description }) 
   function openEditor(event) {
     const button = event.target.closest('button');
     // Outer text includes line breaks `\n` to maintain code formatting in editor examples.
-    const codeBlock = button.parentElement.parentElement.querySelector('.code-block pre code')?.outerText;
+    const codeBlock = button.closest('.playground').querySelector('.code-block pre code')?.outerText;
     const editorOptions = {
       title,
       description,
     };
+
     switch (selected) {
       case UsageTarget.Angular:
         openAngularEditor(codeBlock, editorOptions);
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -1,5 +1,6 @@
-import React, { useState, useEffect } from 'react';
-import useScript from 'react-script-hook';
+import React, { useState, useEffect, useRef } from 'react';
+import Tippy from '@tippyjs/react';
+import 'tippy.js/dist/tippy.css';
 
 import sdk from '@stackblitz/sdk';
 
@@ -15,11 +16,6 @@ enum UsageTarget {
 }
 
 const UsageTargetList = Object.keys(UsageTarget);
-
-enum Theme {
-  Light = 'light',
-  Dark = 'dark',
-}
 
 enum Mode {
   iOS = 'ios',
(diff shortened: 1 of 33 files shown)


static/code/button/usage/basic/index.html
```diff
@@ -0,0 +1,27 @@
...
+  <title>Button</title>
+  <script type="module" src="https://cdn.jsdelivr.net/npm/@ionic/core/dist/ionic/ionic.esm.js"></script>
+  <script nomodule src="https://cdn.jsdelivr.net/npm/@ionic/core/dist/ionic/ionic.js"></script>
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@ionic/core/css/ionic.bundle.css" />
+</head>
+
+<body>
+  <ion-button>Default</ion-button>
+
+  <script>
```
sean-perkins, line 16: If there is another pattern that the team is familiar with, I'm open to alternatives. Since the iframe content is of a variable height, we need to resize the preview container to match the height of the frame. This doesn't happen naturally(as far as I know).

This script emits a message that the parent window can listen to, to access the height and respond to the change. If this pattern is something we move forward with, we should likely break this script out into a package/CDN link that all examples reference.

src/components/global/Playground/index.tsx
```diff
@@ -0,0 +1,430 @@
...
+          </div>
+          <div className="playground__control-group playground__control-group--end">
+            {/* Toggle Source Code */}
+            <Tippy content={expanded ? 'Hide source code' : 'Show full source'}>
+              <button
+                className="playground__icon-button"
+                aria-label={expanded ? 'Hide source code' : 'Show full source'}
+                onClick={() => setExpanded(!expanded)}
+              >
+                <svg
```
sean-perkins, line 322: This icon is giving me troubles exporting cleaning from Figma. We will need to eventually swap this out with a revised SVG viewbox & path that matches the design.

amandaesmith3:
Should this be closed now that we've got it split into smaller PRs?

sean-perkins:
@amandaesmith3 I'm going to keep this open, just as a jumping off/reference point for future PRs. Then will close out completely once `component-playground` branch is at feature parity with this 👍

brandyscarney:
Is this good to close now?

PR #2188, opened 2022-02-08
feat(playground): set-up playground component

sean-perkins:
Introduces the new `<Playground />` component for documenting Ionic Framework components in the docs site.

Displays the initial design with partial dark-mode support.

|Light Theme|Dark Theme|
|---|---|
|Screen Shot 2022-02-08 at 5 30 42 PM|Screen Shot 2022-02-08 at 5 31 06 PM|

Split from #2185

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -14,7 +14,7 @@ export default function Playground() {
 
   return (
     <div className="playground">
-      <div className="playground-container">
+      <div className="playground__container">
         <div className="playground__control-toolbar">
           {/* TODO FW:-742: Code language switcher */}
           <div className="playground__control-group">
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -10,6 +10,9 @@ enum Mode {
 export default function Playground() {
   const [mode, setMode] = useState(Mode.iOS);
 
+  const isIOS = mode === Mode.iOS;
+  const isMD = mode === Mode.MD;
+
   // TODO FW-741: Load code snippets remotely
 
   return (
@@ -20,18 +23,14 @@ export default function Playground() {
           <div className="playground__control-group">
             <button
               type="button"
-              className={
-                'playground__control-button ' + (mode === Mode.iOS ? 'playground__control-button--selected' : '')
-              }
+              className={'playground__control-button ' + (isIOS ? 'playground__control-button--selected' : '')}
               onClick={() => setMode(Mode.iOS)}
             >
               iOS
             </button>
             <button
               type="button"
-              className={
-                'playground__control-button ' + (mode === Mode.MD ? 'playground__control-button--selected' : '')
-              }
+              className={'playground__control-button ' + (isMD ? 'playground__control-button--selected' : '')}
               onClick={() => setMode(Mode.MD)}
             >
               MD
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -1,0 +1,52 @@
+import React, { useState } from 'react';
+
+import './playground.css';
+
+enum Mode {
+  iOS = 'ios',
+  MD = 'md',
+}
+
+export default function Playground() {
+  const [mode, setMode] = useState(Mode.iOS);
+
+  // TODO FW-741: Load code snippets remotely
+
+  return (
+    <div className="playground">
+      <div className="playground-container">
+        <div className="playground__control-toolbar">
+          {/* TODO FW:-742: Code language switcher */}
+          <div className="playground__control-group">
+            <button
+              type="button"
+              className={
+                'playground__control-button ' + (mode === Mode.iOS ? 'playground__control-button--selected' : '')
+              }
+              onClick={() => setMode(Mode.iOS)}
+            >
+              iOS
+            </button>
+            <button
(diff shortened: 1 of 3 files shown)


src/components/global/Playground/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <div className="playground">
+      <div className="playground-container">
+        <div className="playground__control-toolbar">
+          {/* TODO FW:-742: Code language switcher */}
+          <div className="playground__control-group">
+            <button
+              type="button"
+              className={
+                'playground__control-button ' + (mode === Mode.iOS ? 'playground__control-button--selected' : '')
```
liamdebeasi, line 24: Can we move this logic to above the return function? Same thing with line 33. Similar to how we do https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/datetime/datetime.tsx#L1590-L1592

src/components/global/Playground/playground.css
```diff
@@ -0,0 +1,106 @@
...
+
+  padding: 8px 12px;
+
+  border-bottom: 1px solid var(--playground-separator-color);
+
+  overflow-y: auto;
+  scrollbar-width: none;
+}
+
+/* Hides the horizontal scrollbar when the toolbar has horizontal overflow */
```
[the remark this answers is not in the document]
sean-perkins: I think we can remove this for now. Originally when prototyping, on smaller viewports the controls would crowd and the height of the scrollbar was visually overflowing the top bar's inner contents. If that problem returns, we can track a bug and work with Ben.

src/components/global/Playground/playground.css
```diff
@@ -0,0 +1,106 @@
...
+
+.playground__control-group .playground__control-button:last-child {
+  margin-right: 12px;
+}
+
+.playground__control-group--end {
+  margin-left: auto;
+  display: grid;
+  gap: 4px;
+  grid-template-columns: repeat(4, auto);
```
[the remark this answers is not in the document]
sean-perkins: Originally, just because I knew that flex box had mixed support for `gap`. But it looks like that isn't the case anymore.

src/components/global/Playground/playground.css
```diff
@@ -0,0 +1,98 @@
...
+  --playground-btn-color: var(--c-carbon-20);
+  --playground-btn-selected-color: #fff;
+  --playground-btn-selected-background: var(--c-carbon-70);
+
+  --playground-separator-color: var(--c-carbon-70);
+}
+
+.playground {
+  /**
+   * @prop --playground-btn-color The text color of the button in the toolbar.
```
liamdebeasi, line 12: Can we document all of the variables?

liamdebeasi (changes requested):

src/components/global/Playground/playground.css
```diff
@@ -0,0 +1,98 @@
...
+  display: flex;
+  min-height: 24px;
+}
+
+.playground__control-group + .playground__control-group {
+  padding-left: 12px;
+}
+
+.playground__control-group + .playground__control-group:not(.playground__control-group--end) {
+  /* Only render the separator between sections when there is an accompanying section */
```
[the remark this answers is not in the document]
sean-perkins: I can rework the comment 👍 

This selector is used to create the vertical divider between groups of controls in the toolbar. You will notice in the Figma design, there is no divider after the "iOS/MD" group.

src/components/global/Playground/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const isIOS = mode === Mode.iOS;
+  const isMD = mode === Mode.MD;
+
+  // TODO FW-741: Load code snippets remotely
+
+  return (
+    <div className="playground">
+      <div className="playground__container">
+        <div className="playground__control-toolbar">
+          {/* TODO FW:-742: Code language switcher */}
```
amandaesmith3, line 22:
```suggestion
          {/* TODO FW-742: Code language switcher */}
```

src/components/global/Playground/playground.css
```diff
@@ -0,0 +1,107 @@
+/* Dark theme overrides */
+[data-theme="dark"] .playground {
+  --playground-btn-color: var(--c-carbon-20);
+  --playground-btn-selected-color: #fff;
+  --playground-btn-selected-background: var(--c-carbon-70);
+
+  --playground-separator-color: var(--c-carbon-70);
+}
+
+.playground {
```
amandaesmith3, line 10: Can we move the variable documentation into one block, similar to how the components do it? Would be a bit easier to read.

src/components/global/Playground/playground.css
```diff
@@ -0,0 +1,107 @@
...
+
+.playground__control-toolbar {
+  display: flex;
+  align-items: center;
+
+  padding: 8px 12px;
+
+  border-bottom: 1px solid var(--playground-separator-color);
+
+  overflow-y: auto;
```
amandaesmith3, line 55: Should this be `overflow-x`? Feels weird to have a toolbar scroll vertically.

amandaesmith3 (changes requested):
Had a few minor nitpicks, nothing too crazy :eyes;

amandaesmith3 (approved):

liamdebeasi (approved):

PR #2189, opened 2022-02-08
feat(playground): report an issue button

sean-perkins:
Introduces the button to open the Ionic docs "new issue" screen in a new window.

Screen Shot 2022-02-22 at 2 27 00 PM

Split from #2185 continuation of #2188


src/components/global/Playground/playground.css
```diff
@@ -104,3 +110,29 @@
   color: var(--playground-btn-selected-color);
   font-weight: 500;
 }
+
+/* Icon buttons at the end of the toolbar */
+
+.playground__icon-button {
+  display: flex;
+  align-items: center;
+  justify-content: center;;
```
liamdebeasi, line 119:
```suggestion
  justify-content: center;
```

src/components/global/Playground/index.tsx
```diff
@@ -40,6 +40,21 @@ export default function Playground() {
           <div className="playground__control-group playground__control-group--end">
             {/* TODO FW-737: Toggle/Collapse Button */}
             {/* TODO FW-738: Report an Issue Button */}
+            <a
+              className="playground__icon-button"
+              href="https://github.com/ionic-team/ionic-docs/issues/new/choose"
+              aria-label="Report an issue"
+              target="_blank"
```
liamdebeasi, line 47: Can we also add `rel="noreferrer"` when linking out to non-ionic websites?

src/components/global/Playground/index.tsx
```diff
@@ -40,6 +40,22 @@ export default function Playground() {
           <div className="playground__control-group playground__control-group--end">
             {/* TODO FW-737: Toggle/Collapse Button */}
             {/* TODO FW-738: Report an Issue Button */}
```
amandaesmith3, line 42: Can this TODO be removed?

src/components/global/Playground/index.tsx
```diff
@@ -40,6 +40,22 @@ export default function Playground() {
...
+              rel="noreferrer"
+            >
+              <svg aria-hidden="true" width="10" height="10" viewBox="0 0 10 10" fill="none">
+                <path
+                  d="M2.23077 1H9M9 1V7.76923M9 1L1 9"
+                  stroke="current"
+                  strokeLinecap="round"
+                  strokeLinejoin="round"
+                />
+              </svg>
```
amandaesmith3, line 57: This icon really doesn't say "report an issue" to me. At first glance, I thought it would open a Stackblitz, since that arrow usually means "open in new page" or something along those lines. Maybe a wrench or something would be better...? At the very least, I think this could use a tooltip.

amandaesmith3 (changes requested):

src/components/global/Playground/index.tsx
sean-perkins: Good catch, Ben updated the icon after I made this PR. Updated the screenshot in the original description, should be a Github icon. 

All these icon buttons will have tooltips as part of a future PR.

amandaesmith3: A Github icon makes me think "open this source code on Github," but I suppose it's better than the arrow 🤔

amandaesmith3 (approved):

PR #2190, opened 2022-02-08
feat(playground): toggle source code visibility

sean-perkins:
Introduces the ability to toggle the visibility of the code block.

<video src="https://user-images.githubusercontent.com/13732623/153088863-fd67519e-9347-4563-ae58-691ad34592c7.mp4"></video>

Split from #2185 continuation of #2189

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -46,15 +46,15 @@ export default function Playground() {
               onClick={() => setCodeExpanded(!codeExpanded)}
             >
               <svg
+                width="16"
+                height="10"
                 aria-hidden="true"
-                xmlns="http://www.w3.org/2000/svg"
-                width="24"
-                height="24"
-                viewBox="0 0 24 24"
+                viewBox="0 0 16 10"
                 fill="none"
+                xmlns="http://www.w3.org/2000/svg"
               >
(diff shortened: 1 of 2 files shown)


src/components/global/Playground/index.tsx
```diff
@@ -9,6 +11,7 @@ enum Mode {
 
 export default function Playground() {
   const [mode, setMode] = useState(Mode.iOS);
+  const [expanded, setExpanded] = useState(false);
```
liamdebeasi, line 14: Can we make this variable more descriptive? `codeExpanded` or something

src/components/global/Playground/index.tsx
```diff
@@ -38,7 +41,23 @@ export default function Playground() {
             </button>
           </div>
           <div className="playground__control-group playground__control-group--end">
-            {/* TODO FW-737: Toggle/Collapse Button */}
+            <button
+              className="playground__icon-button"
+              aria-label={expanded ? 'Hide source code' : 'Show full source'}
+              onClick={() => setExpanded(!expanded)}
+            >
+              <svg
```
liamdebeasi, line 49: The rendered SVG looks very small compared to what's in the Figma design

sean-perkins: Updated 👍  was having an issue with that asset originally, but looks like Ben fixed it.

liamdebeasi (approved):
Good to go other than the comment update

PR #2200, opened 2022-02-22
cordova-android 10+ compatibility with JDK 11

AustineA:
Updating the doc to reflect cordova-android 10+ is now compatible with JDK 11


brandyscarney (approved):

brandyscarney:
Thank you for the PR!

PR #2203, opened 2022-02-23
feat(playground): copy source code button

sean-perkins:
Introduces the button to copy the contents of the code block to the users clipboard.

Screen Shot 2022-02-23 at 4 44 48 PM

You can verify by clicking the button and seeing that "Fake code block" text is copied to the users clipboard. 

Note: Tooltips for all icon buttons will be a later PR.

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useRef, useState } from 'react';
 
 import CodeBlock from '@theme/CodeBlock';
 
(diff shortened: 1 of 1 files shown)


amandaesmith3 (approved):

liamdebeasi (commented):
In Ben's designs, it looks like there are tooltips that tell you what each button does. Should we have it also say "Code copied" or something like that? There's no visual indicator that the code was copied at the moment.

liamdebeasi:
On that note, I noticed none of the buttons provide tooltips/titles. We should probably fix that too.

sean-perkins:
@liamdebeasi all the tooltips will be introduced in a separate PR 👍

liamdebeasi (approved):

PR #2204, opened 2022-02-24
docs(slider): fix param name

djabif:


brandyscarney:
Thank you for the PR! It looks like this was done so I am going to close this, but we appreciate it!

PR #2207, opened 2022-02-25
Update slides.md

ntorrey:
Fix event reference variable in the setSwiperInstance method.


jcesarmobile:
thanks for the PR, but it was fixed in https://github.com/ionic-team/ionic-docs/pull/2246

PR #2215, opened 2022-03-07
feat(playground): ability to open examples in Stackblitz

sean-perkins:
Introduces the ability to open individual examples in Stackblitz for the respective framework target. Does not currently support Vue (FW-715). 

All boilerplate files are located in `/static/code/stackblitz`. These files are required to create a valid Stackblitz example. 

Note: Opening the Stackblitz example will currently not populate a code snippet, sine that feature has not been implemented yet.

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -3,17 +3,26 @@ import React, { useRef, useState } from 'react';
 import CodeBlock from '@theme/CodeBlock';
 
 import './playground.css';
+import { openAngularEditor, openHtmlEditor, openReactEditor, openVueEditor } from './stackblitz.utils';
+import { Mode, UsageTarget, UsageTargetList } from './playground.types';
 
-enum Mode {
-  iOS = 'ios',
-  MD = 'md',
-}
-
-export default function Playground() {
+/**
+ *
+ * @param title Optional title of the generated playground example. Specify to customize the Stackblitz title.
+ * @param description Optional description of the generated playground example. Specify to customize the Stackblitz description.
+ */
+export default function Playground({ title, description }) {
   const codeRef = useRef(null);
 
+  const [usageTarget, setUsageTarget] = useState(UsageTarget.Html);
   const [mode, setMode] = useState(Mode.iOS);
   const [codeExpanded, setCodeExpanded] = useState(false);
+  const [codeSnippet, setCodeSnippet] = useState({
+    html: '',
+    angular: '',
+    react: '',
+    vue: '',
+  });
 
   const isIOS = mode === Mode.iOS;
   const isMD = mode === Mode.MD;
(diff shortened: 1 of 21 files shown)


src/components/global/Playground/index.tsx
```diff
@@ -25,6 +34,34 @@ export default function Playground() {
     copyButton.click();
   }
 
+  function openEditor(event) {
+    const button = event.target.closest('button');
+    // Outer text includes line breaks `\n` to maintain code formatting in editor examples.
+    const codeBlock = button.closest('.playground').querySelector('.code-block pre code')?.outerText ?? '';
```
amandaesmith3, line 40: Should this part use the `codeSnippet` state instead, or were you not able to preserve line breaks when reading the code in from a file or whatever we decided on? (I'm assuming the intention is to populate the code block with `codeSnippet`'s values in a future PR, since otherwise that state isn't really being used.)

src/components/global/Playground/index.tsx
```diff
@@ -25,6 +34,34 @@ export default function Playground() {
...
+        openReactEditor(codeBlock, editorOptions);
+        break;
+      case UsageTarget.Vue:
+        openVueEditor(codeBlock, editorOptions);
+        break;
+    }
+  }
+
+  // TODO: This variable will be used for generating the buttons to change the code language/framework.
+  const languages = UsageTargetList.filter((lang) => Object.keys(codeSnippet).includes(lang.toLowerCase()));
```
amandaesmith3, line 63: Since the default value for `codeSnippet` includes all language keys from the start, the `includes` check here is redundant (will always match all four languages). Should it be checking to make sure the value for the key is truthy instead?

src/components/global/Playground/stackblitz.utils.ts
```diff
@@ -0,0 +1,125 @@
...
+interface EditorOptions {
+  /**
+   * The title of the Stackblitz example.
+   */
+  title?: string;
+  /**
+   * The description of the Stackblitz example.
+   */
+  description?: string;
+}
```
amandaesmith3, line 19: Could we export this and use it as the type for `editorOptions` in `openEditor()` from `Playground.tsx`?

amandaesmith3 (changes requested):

src/components/global/Playground/index.tsx
sean-perkins: Great question. Thoughts on updating this PR so that `codeBlock` is an empty string (for now), then connecting it to the work of two other open tasks:
- Implementing the framework buttons (Angular/React/Vue/etc.)
- Loading code snippets

?

I believe it should be able to use the code snippet source.

src/components/global/Playground/index.tsx
sean-perkins: Good catch. I'm going to remove this for now. I think it should be handled as the scope for adding the buttons.

src/components/global/Playground/index.tsx
amandaesmith3: Makes sense to me 👍

amandaesmith3 (approved):

liamdebeasi (approved):

PR #2216, opened 2022-03-08
Docs: vue quickstart, fix Jest Javascript configuration errors

praxxys:
Currently, the "use Javascript" section is incomplete as Jest is still configured to use Typescript and running `npm run test:unit` will throw errors. The section was revised to to include Jest Javascript configuration changes.

### DIFF
--- a/docs/vue/quickstart.md
+++ b/docs/vue/quickstart.md
@@ -48,15 +48,17 @@ So, if you’d prefer to use JavaScript instead of TypeScript, you can. After ge
 npm uninstall --save typescript @types/jest @typescript-eslint/eslint-plugin @typescript-eslint/parser @vue/cli-plugin-typescript @vue/eslint-config-typescript
 ```
 
-2. Change all `.ts` files to `.js`. In a blank Ionic Vue app, this should only be `router/index.ts` and `main.ts`.
+2. Change all `.ts` files to `.js`. In a blank Ionic Vue app, this should only be `router/index.ts` and `main.ts`. If you're using tests, you also need to rename `tests/unit/example.spec.ts`.
 
 3. Remove `@vue/typescript/recommended` and `@typescript-eslint/no-explicit-any: ‘off’, `from `.eslintrc.js`.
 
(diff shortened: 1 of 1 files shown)


docs/vue/quickstart.md
```diff
@@ -48,13 +48,13 @@ So, if you’d prefer to use JavaScript instead of TypeScript, you can. After ge
...
-2. Change all `.ts` files to `.js`. In a blank Ionic Vue app, this should only be `src/router/index.ts` and `src/main.ts`.
+2. Change all `.ts` files to `.js`. In a blank Ionic Vue app, this should only be `src/router/index.ts` and `src/main.ts`. If you're using tests, also change the extension of files in the `tests` directory.
 
-3. Remove `@vue/typescript/recommended` and `@typescript-eslint/no-explicit-any: ‘off’, `from `.eslintrc.js`.
+3. In `index.html`, change the imported `<script>` file from `/src/main.ts` to `/src/main.js`.
 
-4. Remove `Array<RouteRecordRaw>` from `src/router/index.js`.
+4. Remove `@vue/typescript/recommended` and `@typescript-eslint/no-explicit-any: ‘off’, `from `.eslintrc.js`.
 
-5. Delete the `src/shims-vue.d.ts` file.
```
liamdebeasi, line 57: Should the `shims-vue` line still be present?

liamdebeasi (approved):
One question about the `shims-vue` line, otherwise looks good.

docs/vue/quickstart.md
mapsandapps: That file doesn't exist when I create a blank vue starter.

liamdebeasi: Ah right. That was specific to the Vue CLI applications which you'll get when using Ionic CLI <= 6: https://github.com/ionic-team/starters/blob/main/vue/base/src/shims-vue.d.ts. Developers on Ionic CLI >=7 will get Vite apps which does not have that file. Maybe we can say "Delete the `src/shims-vue.d.ts` file if it exists. This is only needed when using the Vue CLI."

PR #2221, opened 2022-03-10
fix(playground): match code background to designs

amandaesmith3:
This adds a new CSS variable that changes the background color of the expanded code block, and adds values for both light and dark mode in accordance with the designs.

(FW-717 is a ticket to add dark mode styling, but most of it was already present in the `component-playground` branch. As far as I can tell, this was the only piece missing.)


sean-perkins (approved):
Looks good 👍

liamdebeasi (approved):

PR #2222, opened 2022-03-10
feat(playground): load code usage snippets

sean-perkins:
Supports loading the code usage example files (markdown files) for specified languages (framework targets). It will parse the inner contents of the MD file and set the code snippet value to the React component state.

Usage:
```tsx
<Playground code={{
  html: 'button/basic/javascript.md',
  angular: 'button/basic/angular.md',
  react: 'button/basic/react.md',
  vue: 'button/basic/vue.md'
}}></Playground>
```

The object keys aren't important with this PR and will be finally decided with FW-742.

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -1,19 +1,25 @@
-import React, { useRef, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 
 import CodeBlock from '@theme/CodeBlock';
 
 import './playground.css';
+import { loadCodeSnippet } from './code.utils';
 
 enum Mode {
   iOS = 'ios',
   MD = 'md',
 }
 
-export default function Playground() {
+export default function Playground({ code }) {
+  if (!code) {
+    console.warn('No code usage examples provided for this Playground example.');
+    return;
+  }
   const codeRef = useRef(null);
 
   const [mode, setMode] = useState(Mode.iOS);
   const [codeExpanded, setCodeExpanded] = useState(false);
+  const [codeSnippet, setCodeSnippet] = useState({});
 
   const isIOS = mode === Mode.iOS;
   const isMD = mode === Mode.MD;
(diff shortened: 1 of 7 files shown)


src/components/global/Playground/index.tsx
```diff
@@ -1,19 +1,25 @@
...
+export default function Playground({ code }) {
+  if (!code) {
+    console.warn('No code usage examples provided for this Playground example.');
+    return;
+  }
   const codeRef = useRef(null);
 
   const [mode, setMode] = useState(Mode.iOS);
   const [codeExpanded, setCodeExpanded] = useState(false);
+  const [codeSnippet, setCodeSnippet] = useState({});
```
amandaesmith3, line 22: ~ I know I OK'd this in DM, but I wonder if this should be called `codeSnippets`, plural; the singular makes me think it's the active code snippet being shown, rather than an object that maps languages to their respective snippets. 🤔

src/components/global/Playground/index.tsx
```diff
@@ -1,19 +1,25 @@
...
   iOS = 'ios',
   MD = 'md',
 }
 
-export default function Playground() {
+export default function Playground({ code }) {
+  if (!code) {
+    console.warn('No code usage examples provided for this Playground example.');
+    return;
+  }
```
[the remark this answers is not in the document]
sean-perkins: Good catch!

src/components/global/Playground/index.tsx
sean-perkins: I like your reasoning, updated!

amandaesmith3 (approved):

liamdebeasi (approved):
As discussed offline, I proposed https://github.com/ionic-team/ionic-docs/compare/FW-741...FW-741-alt?expand=1 which is an alternative way of doing this feature.

The concern I have with the current proposed approach is:

1. The code is lazy loaded, so there is the chance of a flash of empty content especially if we decide to have the code snippets be visible by default in the future.
2. We have to parse the MDX ourselves when this feature is already built into Docusaurus. 
3. There are 4 additional network requests per playground. So if a page has 10 playgrounds, that means there are 40 additional network requests.


My proposed change would get rid of the lazy loading, manual MDX parsing, and additional network requests by working with Docusaurus to get the usage examples inlined during the build process.

However, I am not blocking my review on this as I think the current proposed approach should work fine. Whichever approach you would like to go with is fine by me 👍 . Great work on this!

PR #2223, opened 2022-03-11
fix(scripts): always load latest core.json

sean-perkins:
Updates the API script to always point to the latest published release (semantic release) for the `core.json` asset. Unpkg is smart enough that when you do not specify a version, it will pull the latest.

### DIFF
--- a/scripts/api.js
+++ b/scripts/api.js
@@ -8,7 +8,7 @@ const DEMOS_PATH = path.resolve('static/demos');
 let COMPONENT_LINK_REGEXP;
 
 (async function () {
-  const response = await fetch('https://unpkg.com/@ionic/docs@6.0.10-dev.11645801048.2804c20/core.json');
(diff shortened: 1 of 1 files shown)


scripts/api.js
```diff
@@ -8,7 +8,7 @@ const DEMOS_PATH = path.resolve('static/demos');
 let COMPONENT_LINK_REGEXP;
 
 (async function () {
-  const response = await fetch('https://unpkg.com/@ionic/docs@6.0.10-dev.11645801048.2804c20/core.json');
+  const response = await fetch('https://unpkg.com/@ionic/doc/core.json');
```
liamdebeasi, line 11:
```suggestion
  const response = await fetch('https://unpkg.com/@ionic/docs/core.json');
```

liamdebeasi (changes requested):

liamdebeasi (approved):

PR #2228, opened 2022-03-15
feat(playground): add language toggle buttons, populate selected code

amandaesmith3:
- Adds the buttons for switching between languages.
- Populates the code block and Stackblitz editor with the code for your selected language.
- Cleans up some redundant code.
- Fixes the layout shifting when control buttons are selected, including a slight refactor of all control buttons to clean things up.

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -4,6 +4,17 @@ import './playground.css';
 import { EditorOptions, openAngularEditor, openHtmlEditor, openReactEditor, openVueEditor } from './stackblitz.utils';
 import { Mode, UsageTarget } from './playground.types';
 
+const CodeBlockButton = ({ language, usageTarget, setUsageTarget }) => {
+  return (
+    <button
+      type="button"
+      title={`Show ${language} code`}
+      className={`playground__control-button ${usageTarget === language ? "playground__control-button--selected" : ""}`}
+      onClick={() => setUsageTarget(language)}
+    >{language}</button>
+  );
+};
+
 /**
  * @param code The code snippets for each supported framework target.
  * @param title Optional title of the generated playground example. Specify to customize the Stackblitz title.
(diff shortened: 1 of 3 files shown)


liamdebeasi (approved):

src/components/global/Playground/playground.types.ts
```diff
@@ -5,11 +5,7 @@ export enum UsageTarget {
   Vue = 'Vue',
 }
 
-export const UsageTargetList = Object.keys(UsageTarget);
```
liamdebeasi, line 8: Minor thing, but the "Basic" key should say "JavaScript".

```html
<Playground code={{ javascript, react, vue, angular }} />
```

src/components/global/Playground/index.tsx
```diff
@@ -76,7 +85,11 @@ export default function Playground({
     <div className="playground">
       <div className="playground__container">
         <div className="playground__control-toolbar">
-          {/* TODO FW-742: Code language switcher */}
+          <div className="playground__control-group">
+            {Object.values(UsageTarget).map(lang => (
+              <CodeBlockButton language={lang} usageTarget={usageTarget} setUsageTarget={setUsageTarget} />
```
liamdebeasi, line 92: If I click "Vue", this sets the active framework to "Vue", but the code snippet key is passed in as "vue" so the code is never actually shown.

Can we change this so that the key is lower case?

liamdebeasi (commented):
Sorry, noticed something after I approved it

src/components/global/Playground/index.tsx
amandaesmith3: The `CodeBlockButton` component uses `UsageTarget[language]` (the lowercase version) for setting `usageTarget`. The key (the uppercase one) is only used for the display. (I pushed a couple commits right as you were reviewing, so that change might've gotten missed)

liamdebeasi (commented):
Is there any way we could avoid the layout shift when a language button is selected?

https://user-images.githubusercontent.com/2721089/158627948-d651544f-2b85-4795-8e23-708d13bc1f11.mov

amandaesmith3:
@liamdebeasi Done 👍 Side note: the `attr()` CSS function is marked as experimental, but its base functionality has full compatibility in all browsers (https://developer.mozilla.org/en-US/docs/Web/CSS/attr()#browser_compatibility).

liamdebeasi (approved):
Great job!

PR #2230, opened 2022-03-15
feat(playground): populate code block and Stackblitz with code for selected framework

amandaesmith3:
The code block will now show the imported example code for each framework, and the Stackblitz examples are populated as well.

This builds off of https://github.com/ionic-team/ionic-docs/pull/2228, which notably removes some redundant type declarations so `UsageTarget` is the one source of truth. This meant the keys in the `code` prop needed to be updated to match.

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -49,8 +49,10 @@ export default function Playground({
   }
 
   function openEditor(event) {
-    // TODO assign code block value based on active framework button and loaded code snippets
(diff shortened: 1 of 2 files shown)


sean-perkins (changes requested):

amandaesmith3:
On second thought, I'm just gonna close this PR and solely use https://github.com/ionic-team/ionic-docs/pull/2228. After implementing Sean's feedback, the change here is pretty much just one line, so it seems a bit silly to juggle both.

PR #2234, opened 2022-03-16
feat(playground): buttons now show tooltips

liamdebeasi:
Hovering over the source code, github, copy code, and stackblitz buttons now show tooltips.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,6 +24,7 @@
         "@ionic-internal/ionic-ds": "^7.0.0",
         "@mdx-js/react": "^1.6.22",
         "@stackblitz/sdk": "^1.6.0",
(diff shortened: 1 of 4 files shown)


sean-perkins (approved):

amandaesmith3:
Ben's design for the tooltips is slightly different from this: https://www.figma.com/file/jykJRam54AGIOVZbQuxxqV/Ionic-Docs?node-id=0%3A1 Do we need to match that?

liamdebeasi:
I can match it

liamdebeasi:
All fixed. The tooltip styles were getting overridden so I made the selector more specific. Also the styles were referencing a CSS variable that was not defined. Tippy is added to the root of the page not as a child of the Playground.

PR #2237, opened 2022-03-20
Update 3-saving-photos.md

Mansym:
Fixed variable names to fix error: 

Argument of type '{ filepath: string; webviewPath: string; }' is not assignable to parameter of type 'UserPhoto'. [ng] Type '{ filepath: string; webviewPath: string; }' is missing the following properties from type 'UserPhoto': filePath, webViewPath

### DIFF
--- a/docs/angular/your-first-app/3-saving-photos.md
+++ b/docs/angular/your-first-app/3-saving-photos.md
@@ -49,8 +49,8 @@ private async savePicture(photo: Photo) {
   // Use webPath to display the new image instead of base64 since it's
   // already loaded into memory
   return {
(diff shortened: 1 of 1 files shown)


jcesarmobile:
Thanks for the PR, but the code is correct

Maybe you didn't copy the code correctly from https://ionicframework.com/docs/angular/your-first-app/taking-photos#displaying-photos
```
export interface UserPhoto {
  filepath: string;
  webviewPath: string;
}
```

Or made a typo if you wrote the code manually, which makes sense as the variables should be really using camel case, but they don't, so let's keep the code in sync.

PR #2242, opened 2022-03-23
doc(ion-checkbox): Add demo of indeterminate state

ycintre-se:
Add indeterminate state in the live demo


brandyscarney:
Thank you for the PR! We have been moving the API documentation and this is now added under this section: https://ionicframework.com/docs/api/checkbox#indeterminate-checkboxes.

I am going to close this since it has been added but we appreciate it!

PR #2256, opened 2022-04-05
Update index.html

arunredhu:
Modal Example fixed to show the `ion-datetime` in `ion-modal`


sean-perkins:
The `index.html` file is for the HTML live example, it does not support or use Angular syntax.

PR #2258, opened 2022-04-05
fix: fix anchor link on Vue quickstart guide

Jamiewarb:
On the vue quickstart guide, there is an anchor link "#optimizing-your-app". It looks like this section title has been changed to "Optimizing your Build", and so the anchor link is no longer correct. This commit fixes the anchor link


sean-perkins (approved):
Thanks for fixing this 👍

PR #2276, opened 2022-04-18
chore(component-playground): remove webpack plugin

sean-perkins:
This was likely added prior to using `iframe` for our demo examples. Removed & confirmed that examples continue to load correctly.


liamdebeasi (approved):
Yep, this was from the pre-iframe experiments. Shouldn't be needed anymore.

PR #2277, opened 2022-04-18
feat(component-playground): use devicePreview to render phone wrappers

sean-perkins:
Introduces new `devicePreview` option that can be enabled per-playground example to render a "device chrome" around the example iframe for iOS vs. MD. Copies the implementation styles close to the existing Ionic Docs, but uses an SVG instead of a PNG asset.

The inner `device-preview` is a web component with shadow-dom, to prevent style conflicts with the device wrapper.

Updated the button example (in our temporary playground doc), to show the device emulation. 

Live preview (Chrome will likely say it is a dangerous site): https://ionic-docs-git-feat-device-preview-ionic1.vercel.app/docs/playground/button

iOS:

MD:

### DIFF
--- a/src/components/global/Playground/device-preview.js
+++ b/src/components/global/Playground/device-preview.js
@@ -151,11 +151,7 @@ class DevicePreview extends HTMLElement {
     this.modeChanged();
   }
 
-  attributeChangedCallback(
-    name: string,
-    _previousValue: string,
-    _nextValue: string
-  ) {
+  attributeChangedCallback(name, _previousValue, _nextValue) {
     if (name === 'mode') {
       this.modeChanged();
     }
--- a/src/components/global/Playground/device-preview.js
+++ b/src/components/global/Playground/device-preview.js
@@ -123,6 +123,26 @@ class DevicePreview extends HTMLElement {
      height: 100%;
    }
 
+   @media only screen and (max-width: 600px) {
+    :host(.ios) .ios-notch,
+    :host(.md) .md-bar,
+    :host(.ios) figure::after {
+      display: none;
+    }
+
+    :host(.ios) figure,
+    :host(.md) figure {
+      border: 0;
+      border-radius: 0;
+      box-shadow: none;
+    }
+
+    :host(.ios) .content {
+      border-radius: 0;
+    }
+
+   }
+
  `;
 
       const _template = document.createElement('template');
--- a/src/components/global/Playground/device-preview.js
+++ b/src/components/global/Playground/device-preview.js
@@ -124,6 +124,9 @@ class DevicePreview extends HTMLElement {
    }
 
    @media only screen and (max-width: 600px) {
+    :host {
+      --device-width: 100%;
+    }
     :host(.ios) .ios-notch,
     :host(.md) .md-bar,
     :host(.ios) figure::after {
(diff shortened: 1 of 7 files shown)


static/usage/common.css
```diff
@@ -237,3 +237,8 @@
 
   height: 100%;
 }
+
+:root {
```
liamdebeasi, line 241: Should this only be needed if there's a device frame?

liamdebeasi (commented):
The layout doesn't look great on mobile: 
image

Is there a better way we can account for smaller viewports?

sean-perkins:
Perhaps on mobile we shouldn't just turn-off the device preview? At that screen size, there's little value of it being displayed.

liamdebeasi:
I think that's a good idea

src/components/global/Playground/device-preview.js
```diff
@@ -0,0 +1,193 @@
+class DevicePreview extends HTMLElement {
```
amandaejohnston, line 1: Below a certain width, the playground borders are getting covered up ([screenshot](https://i.gyazo.com/254a30a765f708b16f5a6fdf375ce6e4.png)). This doesn't happen on the [non-phone wrapper version](https://i.gyazo.com/383c6788d55ad5c7387fbc707ce45b9e.png).

amandaejohnston (changes requested):

src/components/global/Playground/device-preview.js
sean-perkins: Great catch! Pushed up a fix (deploying), on < ~334px width screens the overflow will kick in and you can horizontally scroll within the container.

sean-perkins: Pushed up that change. It is fair to note that it is likely uncommon for a smaller mobile device to be using the doc demos through the playground. We may want to consider actually hiding the playground iframes in the future and have an action to launch the source in a "new tab", which would allow them to consume the example like a PWA.

amandaejohnston (approved):

static/usage/common.css
liamdebeasi: Not sure off the top of my head. Maybe we can leave it until it starts causing issues for us? You're right though -- if you aren't using headers/footers there will be no visible difference.

src/components/global/Playground/index.tsx
```diff
@@ -89,6 +93,10 @@ export default function Playground({
     }
   }, [isDarkTheme]);
 
+  useEffect(() => {
+    import('./device-preview.js').then((comp) => comp.defineCustomElement());
```
liamdebeasi, line 97: Any reason for doing a dynamic import as opposed to the `import { }` syntax?

liamdebeasi (approved):

src/components/global/Playground/index.tsx
sean-perkins: Honestly, I cannot recall. I know I saw this usage pattern _somewhere_ when researching custom elements, but it doesn't appear to have a different to the functionality here. I'll swap that and then merge.

sean-perkins: Now I remember... haha SSR `HTMLElement` doesn't play nicely if not a dynamic import. Moving it back.

liamdebeasi: Are we able to do this instead? https://github.com/ionic-team/ionic-framework/blob/main/packages/react/src/utils/HTMLElementSSR.ts

Ionic react uses this to get the HTMLElement class to work in SSR

PR #2280, opened 2022-04-18
feat(component-playground): multiple files per demo, angular ivy

sean-perkins:
- Adds support for multiple files per demo example 
- Adds support for overriding imports and declarations for Angular App Module
- Fixes Stackblitz Angular examples to open in Angular Ivy enabled mode

Example usage:
```
import angular_app_component_ts from './angular/app.component_ts.md';
import angular_app_component_html from './angular/app.component_html.md';
import angular_modal_example_component_html from './angular/modal-example.component_html.md';
import angular_modal_example_component_ts from './angular/modal-example.component_ts.md';

import vue_example from './vue/example.vue.md';
import vue_modal from './vue/modal.vue.md';

<Playground
  id="modal-controller"
  code={{
    javascript,
    react,
    vue: {
      files: {
        'src/components/Example.vue': vue_example,
        'src/components/Modal.vue': vue_modal,
      },
    },
    angular: {
      files: {
        'src/app/app.component.html': angular_app_component_html,
        'src/app/app.component.ts': angular_app_component_ts,
        'src/app/modal-example.component.html': angular_modal_example_component_html,
        'src/app/modal-example.component.ts': angular_modal_example_component_ts,
      },
      angularModuleOptions: {
        imports: [`import { ModalExampleComponent } from './modal-example.component';`],
        declarations: ['ModalExampleComponent'],
      },
    },
  }}
  src="usage/modal/controller/demo.html"
/>
```


sean-perkins:
Had a mental thought around this, will need to circle back in the morning. Need to disable the "Copy to Clipboard" primary action when there are multiple file examples. Developers will need to use the individual copy to clipboard buttons in each code block sample.

src/components/global/Playground/index.tsx
```diff
@@ -194,8 +258,19 @@ export default function Playground({
                 </svg>
               </a>
             </Tippy>
-            <Tippy theme="playground" arrow={false} placement="bottom" content="Copy source code">
-              <button className="playground__icon-button playground__icon-button--primary" onClick={copySourceCode}>
+            <Tippy
+              theme="playground"
+              arrow={false}
+              placement="bottom"
+              content={hasOutputTargetOptions ? 'Unavailable in multi-file examples' : 'Copy source code'}
```
amandaejohnston, line 265: ~ Could we also mention somehow that you should use the copy buttons on the individual code blocks? Maybe "For multi-file examples, use copy buttons on code blocks" or something.

amandaejohnston (changes requested):

src/components/global/Playground/index.tsx
sean-perkins: Yeah, I like that a lot!

sean-perkins:
Ended up addressing all the comments in the branched PR: https://github.com/ionic-team/ionic-docs/pull/2283.

That PR will get merged back into this one.

src/components/global/Playground/index.tsx
```diff
@@ -96,19 +112,36 @@ export default function Playground({
...
 
   function copySourceCode() {
+    if (hasOutputTargetOptions) {
+      return;
+    }
     const copyButton = codeRef.current.querySelector('button');
     copyButton.click();
   }
 
+  const hasOutputTargetOptions = codeSnippets[usageTarget] && typeof codeSnippets[usageTarget].$$typeof !== 'symbol';
```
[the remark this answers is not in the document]
sean-perkins: Ended up resolving this in this PR: https://github.com/ionic-team/ionic-docs/pull/2283 (will get merged back up into this branch).

src/components/global/Playground/index.tsx
```diff
@@ -264,16 +346,16 @@ const FRAME_SIZES = {
...
     frame.contentWindow.addEventListener('demoReady', () => {
       resolve();
     });
   });
 };
 
-const isFrameReady = (frame: HTMLElement) => {
-  return frame.contentWindow.demoReady === true;
+const isFrameReady = (frame: HTMLIFrameElement) => {
+  return (frame.contentWindow as any).demoReady === true;
```
liamdebeasi, line 360: Why do we need to cast this as `any`?

src/components/global/Playground/index.tsx
```diff
@@ -129,11 +165,42 @@ export default function Playground({
...
+  function getCodeSnippetId(fileName: string) {
+    let fileNameId = fileName;
+    // replace all non-alphanumeric characters with underscores
+    fileNameId = fileNameId.replace(/[^a-zA-Z0-9]/g, '_');
+    return `playground_${id}_${fileNameId}`;
+  }
+
+  function renderCodeSnippets() {
+    if (codeSnippets[usageTarget]) {
+      if (typeof codeSnippets[usageTarget].$$typeof === 'symbol') {
```
liamdebeasi, line 190: What does this line do?

src/components/global/Playground/index.tsx
```diff
@@ -129,11 +165,42 @@ export default function Playground({
...
     Object.keys(code).forEach((key) => {
       // Instantiates the React component from the MDX content.
-      codeSnippets[key] = code[key]({});
+      if (typeof code[key] === 'function') {
+        codeSnippets[key] = code[key]({});
+      } else if (typeof code[key] === 'object') {
+        const fileSnippets = {};
+        for (const fileName of Object.keys(code[key].files)) {
+          fileSnippets[`${fileName}`] = code[key].files[fileName]({});
+        }
```
liamdebeasi, line 174: What does this do? I know it "Instantiates the React component from the MDX content" but not sure why the extra cases are needed.

src/components/global/Playground/index.tsx
```diff
@@ -38,21 +38,40 @@ const CodeBlockButton = ({ language, usageTarget, setUsageTarget, setCodeExpande
   );
 };
 
+type MdxContent = () => {};
+
+interface OutputTargetOptions {
```
liamdebeasi, line 43: What is an "output target option" and why is it needed?

liamdebeasi (commented):
I don't see any red flags in the code, but it would be great to add more comments that explains what the code does and why there needs to be multiple if cases.

src/components/global/Playground/index.tsx
sean-perkins: `demoReady` is a made up prop that we are directly assigning to `window`. The type of `frame.contentWindow` (`Window`) won't include that value and will throw a type error in an IDE that does type checking.

src/components/global/Playground/index.tsx
sean-perkins: I can add a comment block in here explaining this code. 

If the value of `code[key]` is a function, that means you are using a single-file example and are binding the MDX content directly to the prop:

```tsx
import javascript from 'javascript.md';

<Playground code={{ javascript }} />
```

Otherwise, if the type is an object, you are using the advanced configuration object to specify multiple code snippets for a single usage target and have options for customizing the angular `AppModule`. It loops through all your specified files and does the same thing as the single-file example, which constructs the React component from the MDX object. We keep track of the file names, because they coincide with the path of the file when creating the Stackblitz example from the code snippets.

```tsx
import file_one from 'file-one.md';
import file_two from 'file-two.md';

<Playground code={{
  javascript: {
    files: {
      'src/example/file-one.html': file_one,
      'src/example/file-two.html': file_two
    }
  }
}} />
```

src/components/global/Playground/index.tsx
sean-perkins: It is the uglier solution for knowing if the object is MDX content vs. the object syntax, I will flip this to the updated syntax from Amanda's feedback and add a comment explaining it.

src/components/global/Playground/index.tsx
sean-perkins: I could probably call this `UsageTargetOptions` to align with our existing variable naming. We basically have two valid type signatures that can be assigned to the `code` prop. MDX content for single-file examples and an object for multi-file examples + advanced configurations. Having the types separate helps so I can later type-cast and have type checking for the advanced configuration options. This interface could use a comment on its purpose 👍

PR #2283, opened 2022-04-19
feat(playground): playground multi-file tabs

sean-perkins:
Adds tab item support when a playground example has multiple files. Includes left/right navigation and scrolling to access the available files.

The `PlaygroundTabs` component is just the base `Tabs` component from Docusaurus, extended to include custom icon support and left/right navigation.

<details>
 
 
</details>


amandaejohnston:
Is there a way to view an example of this on the preview? https://ionic-docs-git-feat-component-playground-tabs-ionic1.vercel.app/docs/playground/modal gives me a 404, but I was just guessing.

sean-perkins:
@amandaejohnston I split out the changes from my modal PR, so that these could get into `component-playground` first and the team could potentially use it within the other playground sprint tasks. 

If you checkout `FW-1281-presentation`, it combines all the open PRs + writing an example with the modal doc page.

I figured this would be easier than trying to review these changes + the many file addition changes of FW-1281-presentation.

src/components/global/Playground/index.tsx
```diff
@@ -182,19 +185,77 @@ export default function Playground({
...
+        <PlaygroundTabs className="playground__tabs">
+          {Object.keys(codeSnippets[usageTarget]).map((fileName) => (
+            <TabItem
+              className="playground__tab-item"
+              value={fileName}
+              label={fileName}
+              key={fileName}
+              {...{
+                icon: getFileIcon(fileName),
+              }}
```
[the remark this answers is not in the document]
sean-perkins: This is extending the docusaurus `TabItem`, which doesn't have `icon` specified on its available props. To avoid having to have a custom `PlaygroundTabItem` just for the sake of passing props (which it does do), I used this approach to get around the typing issues.

This allows us to only have to maintain the outermost `Tabs` implementation and benefit from any docusaurus improvements to the tab item.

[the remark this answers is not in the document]
amandaejohnston: Aha! Sneaky 😆

src/components/global/PlaygroundTabs/styles.module.css
```diff
@@ -0,0 +1,58 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+ .tabNav {
+  position: relative;
+}
+
+.tabList {
+ margin-bottom: var(--ifm-leading);
```
liamdebeasi, line 13: Can we fix the indentation in this file?

src/components/page/reference/ReleaseNotes/release-notes.json
```diff
@@ -238,15 +248,5 @@
     "tag_name": "v5.8.4",
     "type": "patch",
     "version": "5.8.4"
-  },
```
liamdebeasi, line 241: Can we revert this change?

liamdebeasi (commented):
It's a bit hard to validate that this is working as expected. Do you have a list of what this feature should/should not do?

sean-perkins:
@liamdebeasi since it extends the base `Tabs` component from docusaurus, it should maintain all of that base functionality (moving left/right with arrow keys, one active tab at a time, etc.).

The custom features are as follows:
- Icons will display within the tab item header for supported file types (TS, HTML, CSS/SASS, Vue)
- Left/Right arrow navigation buttons allow users to click to scroll left or right.
- Arrow navigation buttons will only display when the tabs has overflow

`FW-1281-presentation` shows all the open PRs combined with a the modals docs page (controller usage has multiple files for Angular framework).

src/components/global/Playground/index.tsx
```diff
@@ -182,19 +185,77 @@ export default function Playground({
     return `playground_${id}_${fileNameId}`;
   }
 
+  function getFileIcon(fileName: string) {
```
liamdebeasi, line 188: Can we move these SVGs to separate files that we import? The code of this component is getting a bit hairy with all the SVG data inlined.

src/components/global/PlaygroundTabs/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+  const tabRefs: (HTMLLIElement | null)[] = [];
+  const { blockElementScrollPositionUntilNextRender } = useScrollPositionBlocker();
+
+  const handleTabChange = (event: React.FocusEvent<HTMLLIElement> | React.MouseEvent<HTMLLIElement>) => {
+    const newTab = event.currentTarget;
+    const newTabIndex = tabRefs.indexOf(newTab);
+    const newTabValue = values[newTabIndex]!.value;
+
+    if (newTabValue !== selectedValue) {
+      blockElementScrollPositionUntilNextRender(newTab);
```
liamdebeasi, line 82: Why is this needed?

src/components/global/PlaygroundTabs/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+        break;
+      }
+      default:
+        break;
+    }
+
+    focusElement?.focus();
+  };
+
+  useEffect(() => {
```
liamdebeasi, line 108: Why is this needed?

src/components/global/PlaygroundTabs/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+import styles from './styles.module.css';
+
+// A very rough duck type, but good enough to guard against mistakes while
+// allowing customization
+function isTabItem(comp: ReactElement): comp is ReactElement<TabItemProps & { icon?: JSX.Element }> {
+  return typeof comp.props.value !== 'undefined';
+}
+
+function TabsComponent(props: Props): JSX.Element {
```
liamdebeasi, line 24: It seems like we are copy and pasting the Docusaurus Tabs implementation instead of extending it. Is there a way we can extend this without having to re-implement?

sean-perkins: Docusaurus does not expose access to manipulate the render functions that we are extending. We could use reflection, but I would advise against it. We could also roll our own tabs component (which is sorta what we are doing), but then we would need to go through the process of revalidating keyboard interactions and other design considerations that docusaurus has already validated in their codebase.

liamdebeasi: Ok sounds good

sean-perkins: If we start using this pattern in more than just overlays (areas that currently I assume will need multi-file examples for the controller pattern); we could start to add spikes to own the implementation source more as well. AFAIK, most of this implementation could be a web component, since the content is slotted in.

src/components/global/PlaygroundTabs/index.tsx
sean-perkins: Updated the base `button` example, since we are going to delete that before merging `component-playground` into `main`. You will be able to see the arrow navigation behavior in the live demo.

This code is responsible for hiding the arrow navigation initially, until there is horizontal scroll. There is a slight jump for the left nav button when it is initially rendered (once reaching the 40px offset), but discussed with Ben and this was ok for the first iteration. The goal is to not display the arrow navigation until they are needed.

src/components/global/PlaygroundTabs/index.tsx
sean-perkins: This is taken from the docusaurus original source: https://github.com/facebook/docusaurus/blob/4e45e14fdd760a7eff13f842c81654eba5e872a3/packages/docusaurus-theme-common/src/utils/scrollUtils.tsx#L178-L188

In theory, since we aren't using `groupId` in this feature set, it could be removed, but I sided on not removing any unnecessary code from the original implementation in the event we wanted to pull new changes from their code.

liamdebeasi (approved):

PR #2288, opened 2022-04-22
feat(playground): ability to reset demos

sean-perkins:
Introduces a button that allows users to reload the rendered iframe previews back to their original state. Useful for examples where you may present a modal, but not show how to dismiss.

Screen Shot 2022-04-22 at 4 18 50 PM

Builds on the updated stackblitz icon PR, to maintain correct button ordering.

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -228,17 +228,15 @@ export default function Playground({
               <button className="playground__icon-button playground__icon-button--primary" onClick={openEditor}>
                 <svg
                   aria-hidden="true"
-                  width="12"
-                  height="12"
-                  viewBox="0 0 12 12"
+                  width="10"
+                  height="14"
+                  viewBox="0 0 10 14"
                   fill="none"
                   xmlns="http://www.w3.org/2000/svg"
                 >
-                  <path d="M6 11L11 11" stroke="#92A0B3" strokeLinecap="round" strokeLinejoin="round" />
                   <path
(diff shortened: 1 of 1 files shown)


amandaejohnston:
Had we discussed this feature at some point? I don't see it in the design doc, and I'm not I'm following the use case since you could just have a button to dismiss the modal.

I'm open to this, of course, especially since it's a small add. I just want to guard against scope creep/feature bloat.

sean-perkins:
Similar concepts exist in [MUI](https://mui.com/material-ui/react-dialog/). Our playground examples should be minimal around a concept. 

For modals, introducing a dismiss action into each example, would mean that each snippet for each framework would also need to include that example code. A reset action allows users to perform a small action, see the behavior, and reset back to the initial state to repeat the interaction. 

Here is the problem with modals:
<video src="https://user-images.githubusercontent.com/13732623/164793142-40fb3322-922b-47cc-8fbf-7459f63d4846.mp4"></video>

Being able to show how to simply present a modal with the most minimal code is great for new Ionic developers. But once the modal is presented, I cannot repeat that behavior without a full refresh of the page. The same issue with dismissing. I can dismiss the modal, but once I do it once, I cannot repeat the functionality. I don't want to couple multiple concepts in the examples that are showing a specific purpose.

This problem also exists in areas like `ion-select`. Once you make a selection, how does a user return back to the state of no option selected?

amandaejohnston (approved):

PR #2298, opened 2022-04-28
docs(datetime): add localization playgrounds

liamdebeasi:
This PR adds the following playgrounds:

1. How to manually specify a locale.
2. How to manually specify the hour cycle.
3. How to customize what datetime uses as the first day of the week.
4. How to customize the time label.
5. How to use Locale extension tags with datetime.

These fall under the "Localization" category.


docs/api/datetime.md
```diff
@@ -107,9 +114,15 @@ Ionic Framework makes use of the [Intl.DatetimeFormat](https://developer.mozilla
...
 
-For instances where you need a specific locale, you can use the `locale` property to set it.
+For instances where you need a specific locale, you can use the `locale` property to set it. The locale controls both the language and the date and time formats that are displayed.
 
-TODO
+The following example shows how to set the locale to Spanish (Spain).
+
+<CustomLocale />
+
+:::note
```
sean-perkins, line 123: Can we explore if adding another line-break here, would create spacing between the playground example and the note?

Screen Shot 2022-05-02 at 12 47 48 PM

Or, seeing this with text below the playground example as well; maybe we want to add bottom padding to the playground component?

sean-perkins (changes requested):
Content/examples look good! Just one small styling consideration to add space between the playground and the text/alert/etc. that follows.

amandaejohnston (approved):

docs/api/datetime.md
liamdebeasi: Updated the playground styles to add a margin:

image

PR #2302, opened 2022-04-28
docs(datetime): add theming playground

liamdebeasi:
This PR adds the Theming section to the datetime playground.


sean-perkins (approved):
Changes look good. There is a bug with changing from "Angular" to "React" in the playground. It looks like any example that has multiple multi-file examples, the active tab is not being set when it is re-rendered with new contents.

amandaejohnston (commented):
I'm unable to open the React Stackblitz; the playground throws an error in the console. Not sure if it's this code or the playground component itself that's the problem, though 🤔

liamdebeasi:
@amandaejohnston That should be fixed by https://github.com/ionic-team/ionic-docs/pull/2306 once I pull in those changes.

PR #2304, opened 2022-04-28
docs(angular, react, vue): add tabs best practices

liamdebeasi:
This PR does the following things:

**Angular**
- Describes how to define child routes within tabs
- Describes the basic tab functionality in Ionic
- Describes how to switch between tabs/best practices

**React**
- Describes how to define child routes within tabs
- Describes the basic tab functionality in Ionic
- Describes how to switch between tabs/best practices

**Vue**
- Describes how to define child routes within tabs
- Describes the basic tab functionality in Ionic
- Describes how to switch between tabs/best practices

The goal of this PR is to give developers some guidance on a) how to use tabs in Ionic correctly and b) how tabs differs from other tabs implementations.


liamdebeasi:
Added the missing tabs to the angular sample. Re: React/Vue I could go either way. I do like that it is "copy-pastable" but it may not be necessary.

sean-perkins (approved):
I could go either way on the examples too. I think this is good for now and we can gauge what questions we receive (and issues that people report that won't find this doc first) and go from there. 

Would love to have visual diagrams for some of the routing examples in the future.

docs/angular/navigation.md
```diff
@@ -215,3 +217,105 @@ Here we have a "tabs" path that we load. In this example we call the path "tabs"
...
+
+A good example of this in practice is the iOS App Store and Google Play Store mobile applications. These apps both provide tabbed interfaces, but neither one ever routes the user across tabs. For example, the "Games" tab in the iOS App Store app never directs users to the "Search" tab and vice versa.
+
+Let's take a look at a couple common mistakes that are made with tabs.
+
+**A Settings Tab That Multiple Tabs Reference**
+
+A common practice is to create a Settings view as its own tab. This is great if developers need to present several nested settings menus. However, other tabs should never try to route to the Settings tab. As we mentioned above, the only way that the Settings tab should be activated is by a user tapping the appropriate tab button.
+
+If you find that your tabs need to reference the Settings tab, we recommend making the Settings view a modal by using `ion-modal`. This is a practice found in the iOS App Store app, and it works very well. With this approach, any tab can present the modal without breaking the mobile tabs pattern of each tab being its own stack.
```
amandaejohnston, line 297:
```suggestion
If you find that your tabs need to reference the Settings tab, we recommend making the Settings view a modal by using `ion-modal`. This is a practice found in the iOS App Store app. With this approach, any tab can present the modal without breaking the mobile tabs pattern of each tab being its own stack.
```
~ A bit too opinionated for docs, I think 🤔 (Also applies to the other frameworks)

amandaejohnston (approved):

docs/angular/navigation.md
liamdebeasi: Fixed!

PR #2308, opened 2022-05-02
Fix link to alert component

thomasklemm:
Tiny fix for a link that leads to a different component than expected (Modal instead of Alert).


liamdebeasi (approved):

liamdebeasi:
Thank you!

PR #2311, opened 2022-05-03
docs(select): add basic playgrounds

amandaejohnston:
Swaps the Single Selection and Interfaces sections for better document flow, and adds the following playgrounds:
- Single Selection
- Multiple Selection
- Responding to Interaction

### DIFF
--- a/static/usage/select/basic/multiple-selection/angular.md
+++ b/static/usage/select/basic/multiple-selection/angular.md
@@ -1,0 +1,7 @@
+```html
+<ion-select placeholder="Select all fruits that apply" multiple>
+  <ion-select-option>Apples</ion-select-option>
+  <ion-select-option>Oranges</ion-select-option>
+  <ion-select-option>Bananas</ion-select-option>
+</ion-select>
+```
--- a/static/usage/select/basic/multiple-selection/vue.md
+++ b/static/usage/select/basic/multiple-selection/vue.md
@@ -1,0 +1,18 @@
+```html
+<template>
+  <ion-select placeholder="Select all fruits that apply" multiple>
+    <ion-select-option>Apples</ion-select-option>
+    <ion-select-option>Oranges</ion-select-option>
+    <ion-select-option>Bananas</ion-select-option>
+  </ion-select>
+</template>
+
+<script>
+  import { IonSelect, IonSelectOption } from '@ionic/vue';
+  import { defineComponent } from 'vue';
+
+  export default defineComponent({
+    components: { IonSelect, IonSelectOption },
+  });
+</script>
+```
--- a/static/usage/select/basic/responding-to-interaction/angular/app.component_html.md
+++ b/static/usage/select/basic/responding-to-interaction/angular/app.component_html.md
@@ -1,0 +1,16 @@
+```html
+<ion-select
+  placeholder="Select fruit"
+  (ionChange)="handleChange($event)"
+  (ionCancel)="pushLog('ionCancel fired')"
+  (ionDismiss)="pushLog('ionDismiss fired')"
+>
+  <ion-select-option>Apples</ion-select-option>
+  <ion-select-option>Oranges</ion-select-option>
+  <ion-select-option>Bananas</ion-select-option>
+</ion-select>
+<div id="log" class="ion-padding">
+  <p *ngFor="let log of logs">{{ log }}</p>
+  <p>Events will log above</p>
+</div>
+```
--- a/static/usage/select/basic/responding-to-interaction/react.md
+++ b/static/usage/select/basic/responding-to-interaction/react.md
@@ -1,0 +1,33 @@
+```tsx
+import React, { useState } from 'react';
+import { IonSelect, IonSelectOption } from '@ionic/react';
+
+function Example() {
+  const [logs, setLogs] = useState([]);
+
+  const pushLog = (msg) => {
+    setLogs([msg, ...logs]);
+  };
+
+  return (
+    <>
+      <IonSelect
+        placeholder="Select fruit"
+        onIonChange={(e) => pushLog(`ionChange fired with value: ${e.detail.value}`)}
+        onIonCancel={() => pushLog('ionCancel fired')}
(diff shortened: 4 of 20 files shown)


static/usage/select/basic/single-selection/vue.md
```diff
@@ -0,0 +1,18 @@
+```html
+<template>
+  <ion-select placeholder="Select fruit">
+    <ion-select-option>Apples</ion-select-option>
```
liamdebeasi, line 4: Each `ion-select-option` should have a `value` property associated with it.

sean-perkins (changes requested):
Many of the demo examples are not pointing to the correct path for `common.js` and `common.css` (errors in the console). Can we update those relative paths? Would like to see the examples in their desired state before reviewing through.

static/usage/select/basic/multiple-selection/angular.md
```diff
@@ -0,0 +1,7 @@
+```html
+<ion-select placeholder="Select all fruits that apply" multiple>
```
liamdebeasi, line 2:
```suggestion
<ion-select placeholder="Select all fruits that apply" [multiple]="true">
```
Multiple is a property not an attribute so we need to always provided "true" (even though this will technically work. Angular apps with strict mode enabled will complain about this)

static/usage/select/basic/single-selection/vue.md
liamdebeasi: This is true for all of the other examples as well.

static/usage/select/basic/multiple-selection/angular.md
liamdebeasi: This is also true for all of the other examples. The compiler may not complain, but it's still not correct usage.

liamdebeasi (changes requested):
The layout for each demo doesn't look great: 
image

It doesn't look like the `container` styles are being loaded.

static/usage/select/basic/multiple-selection/vue.md
```diff
@@ -0,0 +1,18 @@
+```html
+<template>
+  <ion-select placeholder="Select all fruits that apply" multiple="true">
```
sean-perkins, line 3:
```suggestion
  <ion-select placeholder="Select all fruits that apply" :multiple="true">
```

static/usage/select/basic/responding-to-interaction/angular/app.component_html.md
```diff
@@ -0,0 +1,16 @@
...
+  placeholder="Select fruit"
+  (ionChange)="handleChange($event)"
+  (ionCancel)="pushLog('ionCancel fired')"
+  (ionDismiss)="pushLog('ionDismiss fired')"
+>
+  <ion-select-option value="apples">Apples</ion-select-option>
+  <ion-select-option value="oranges">Oranges</ion-select-option>
+  <ion-select-option value="bananas">Bananas</ion-select-option>
+</ion-select>
+<div id="log" class="ion-padding">
```
sean-perkins, line 12:
```suggestion
<div class="ion-padding">
```

static/usage/select/basic/responding-to-interaction/react.md
```diff
@@ -0,0 +1,33 @@
...
+        placeholder="Select fruit"
+        onIonChange={(e) => pushLog(`ionChange fired with value: ${e.detail.value}`)}
+        onIonCancel={() => pushLog('ionCancel fired')}
+        onIonDismiss={() => pushLog('ionDismiss fired')}
+      >
+        <IonSelectOption value="apples">Apples</IonSelectOption>
+        <IonSelectOption value="oranges">Oranges</IonSelectOption>
+        <IonSelectOption value="bananas">Bananas</IonSelectOption>
+      </IonSelect>
+      <div id="log" className="ion-padding">
```
sean-perkins, line 24:
```suggestion
      <div className="ion-padding">
```

static/usage/select/basic/responding-to-interaction/vue.md
```diff
@@ -0,0 +1,37 @@
...
+    placeholder="Select fruit"
+    @ionChange="pushLog('ionChange fired with value: ' + $event.detail.value)"
+    @ionCancel="pushLog('ionCancel fired')"
+    @ionDismiss="pushLog('ionDismiss fired')"
+  >
+    <ion-select-option value="apples">Apples</ion-select-option>
+    <ion-select-option value="oranges">Oranges</ion-select-option>
+    <ion-select-option value="bananas">Bananas</ion-select-option>
+  </ion-select>
+  <div id="log" class="ion-padding">
```
sean-perkins, line 13:
```suggestion
  <div class="ion-padding">
```

sean-perkins (changes requested):
Small little clean-up items for examples that don't need the document id and one Vue syntax thing. 

Also, it is likely just a preference, but it feels more natural to me that events would log under the header/label, instead of displaying above and pushing the label offscreen.

amandaejohnston:
I added the logs at the top so the latest events would always show, rather than forcing the user to scroll down after the first few (or adding auto-scrolling which would clutter things up). I'm not that married to it though, if you much prefer the other way.

sean-perkins:
Isn't that what the `unshift` accomplishes? Or does having the text "Events show below/above" push the contents off screen, requiring scroll?

amandaejohnston:
Oh, were you thinking have the label separate from the scrolling container so new events get added in-between the label and the last event fired?

amandaejohnston:
I just removed the label completely; I agree that the old version looked odd, but I think it was a little cluttered with the label present.

liamdebeasi (approved):

sean-perkins (approved):
Looks good 👍  just needs a re-base or merge from the base branch.

amandaejohnston:
Re-requesting review after adding the `ion-item`s. I also merged in the base branch to resolve merge conflicts, which pulled in the Interface playgrounds; you can ignore those for the purpose of this PR.

liamdebeasi (approved):

PR #2313, opened 2022-05-03
docs(modal): presenting inline modals

sean-perkins:
This PR adds the component playground examples for presenting inline modals.

### DIFF
--- a/docs/api/modal.md
+++ b/docs/api/modal.md
@@ -42,13 +42,13 @@ A Modal is a dialog that appears on top of the app's content, and must be dismis
 
 When using `ion-modal` with Angular, React, or Vue, the component you pass in will be destroyed when the modal is dismissed. As this functionality is provided by the JavaScript framework, using `ion-modal` without a JavaScript framework will not destroy the component you passed in. If this is a needed functionality, we recommend using the `modalController` instead.
 
-TODO: Playground Example
-
 #### Using triggers
 
 A trigger is an element reference, that when selected will automatically open a modal. Triggers are useful when presenting a modal from a button or similar interaction.
 
-TODO: Playground Example
+import PresentTriggerExample from '@site/static/usage/modal/presenting/inline/trigger/index.md';
+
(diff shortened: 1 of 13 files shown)


docs/api/modal.md
```diff
@@ -42,21 +42,23 @@ A Modal is a dialog that appears on top of the app's content, and must be dismis
...
 #### Using `isOpen`
 
 The `isOpen` property on `ion-modal` allows developers to control the presentation state of the modal from their application state. This means when `isOpen` is set to `true` the modal will be presented and when `isOpen` is set to `false` the modal will be dismissed.
 
 `isOpen` uses one-way data binding, meaning it will not automatically be set to `false` when the modal is dismissed. To keep the state of `isOpen` in sync with your UI state, listen for either `ionModalDidDismiss` or `didDismiss` and set the variable to `false`.
 
-TODO: Playground Example
+import PresentIsOpenExample from '@site/static/usage/modal/presenting/inline/is-open/index.md'; 
+
+<PresentIsOpenExample />
```
amandaejohnston, line 61: Rather than just hardcode `isOpen` to `true`, could we have a button that sets it instead? That would be closer to the code people will actually write. It'd also be more clear what's actually happening in the playground, since right now it's hard to tell that you're looking at a modal unless you reset the playground and see the animation.

amandaejohnston (changes requested):

liamdebeasi (changes requested):
Ditto what Amanda said. It would also be helpful to show how to reset the `isOpen` state by listening to `ionModalDidDismiss`.

amandaejohnston:
Now that I'm seeing some of these live, I wonder if it'd be better to combine the Presenting and Dismissing playgrounds for the most part. The split in info feels a little strained at times -- for example, my initial thought was also to include a `didDismiss` listener, but I realized you were probably going to put that in the Dismissing playground instead. But the paragraph right above the Presenting playground calls that part of the API out, so it feels odd to not include it here.

PR #2316, opened 2022-05-03
docs(select): add objects as values playgrounds

amandaejohnston:
Adds two playgrounds for demonstrating objects as values, with single or multiple selection.

### DIFF
--- a/static/usage/select/objects-as-values/multiple-selection/angular/app.component_ts.md
+++ b/static/usage/select/objects-as-values/multiple-selection/angular/app.component_ts.md
@@ -1,0 +1,46 @@
+```ts
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'app-root',
+  templateUrl: 'app.component.html',
+})
+export class AppComponent {
+  currentFood = "";
+
+  foods = [
+    {
+      id: 1,
+      name: "Apples",
+      type: "fruit"
+    },
+    {
+      id: 2,
+      name: "Carrots",
+      type: "vegetable"
+    },
+    {
+      id: 3,
+      name: "Cupcakes",
+      type: "dessert"
+    }
+  ];
+
+  compareWith(o1, o2) {
+    if(!o1 || !o2) {
+      return o1 === o2;
+    }
+
+    if(Array.isArray(o2)) {
+      return o2.some((o) => o.id === o1.id);
+    }
+
+    return o1.id === o2.id;
+  }
+
+  handleChange(ev) {
+    this.currentFood = JSON.stringify(ev.target.value);
+  }
+}
+
+```
--- a/static/usage/select/objects-as-values/multiple-selection/angular/app.component_ts.md
+++ b/static/usage/select/objects-as-values/multiple-selection/angular/app.component_ts.md
@@ -6,7 +6,7 @@ import { Component } from '@angular/core';
   templateUrl: 'app.component.html',
 })
 export class AppComponent {
-  currentFood = "[]";
+  currentFood = undefined;
 
   foods = [
     {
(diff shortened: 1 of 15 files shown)


static/usage/select/objects-as-values/multiple-selection/demo.html
```diff
@@ -0,0 +1,75 @@
...
+  <script type="module" src="https://cdn.jsdelivr.net/npm/@ionic/core@6/dist/ionic/ionic.esm.js"></script>
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@ionic/core@6/css/ionic.bundle.css" />
+</head>
+
+<body>
+  <ion-app>
+    <ion-content>
+      <div class="container">
+        <ion-select placeholder="Select food" multiple="true"></ion-select>
+        <ion-label class="ion-padding">
```
sean-perkins, line 19: Can we wrap this element in `div` (and apply the same to each usage target). That way the select and the output will stack, vs. on smaller screens the select is truncating.

static/usage/select/objects-as-values/multiple-selection/vue.md
```diff
@@ -0,0 +1,57 @@
+```html
+<template>
+  <ion-select
+    placeholder="Select fruit"
+    :compareWith="compareWith"
+    @ionChange="currentFood = JSON.stringify($event.detail.value)"
+    multiple="true"
```
sean-perkins, line 7:
```suggestion
    :multiple="true"
```

sean-perkins (changes requested):

liamdebeasi (changes requested):
All of the `ion-select` instances here should be placed inside of `ion-item` elements: https://ionicframework.com/docs/api/select#usage

This also helps us avoid issues in the object values demos where the select is expanding to be the width of the logged text: 

image


The logged text should be outside of the `ion-item`.

amandaejohnston:
I wanted to avoid using `ion-item` on select playgrounds since it isn't strictly necessary for the component to work, and I didn't want to draw attention away from the select itself. I can see the case for adding them, though, since it's certainly the most common use case. I added a set width so the select doesn't expand weirdly, but let me know if you feel strongly about adding the items as well.

liamdebeasi:
All of our form components are built with the intent that they'd be used inside of `ion-item` (ion-input, ion-textarea, etc). I think we should show code examples that follow our best practices, not just the minimum code required to get something working.

amandaejohnston:
Sounds good. I'll update this branch first, then go adjust the others.

static/usage/select/objects-as-values/multiple-selection/angular/app.component_ts.md
```diff
@@ -0,0 +1,46 @@
...
+
+    if(Array.isArray(o2)) {
+      return o2.some((o) => o.id === o1.id);
+    }
+
+    return o1.id === o2.id;
+  }
+
+  handleChange(ev) {
+    this.currentFood = JSON.stringify(ev.target.value);
```
sean-perkins, line 42: ~ Angular has a built-in `| json` pipe that can handle this for you. We should consider using that, just to remove any confusion of why the initial value is a string of an array.

Updated Stackblitz example that shows usage: https://stackblitz.com/edit/angular-v3mgxj?file=src%2Fapp%2Fapp.component.html,src%2Fapp%2Fapp.component.ts

This would apply for both Angular examples.

sean-perkins (approved):
Looks good. One suggestion on the Angular code.

liamdebeasi (changes requested):
Can we fix the containers so the item does not span the entire width of the playground? 
image

liamdebeasi (approved):

PR #2317, opened 2022-05-03
docs(modal): dismissing inline modals

sean-perkins:
This PR adds the component playground examples for dismissing inline modals.

PR #2321, opened 2022-05-03
docs(modal): inline playground

sean-perkins:
Re-approach to the inline modal component playground examples. 

Shows two examples:
1. Kitchen sink (presenting, dismissing with data & role)
2. `isOpen` (isolated example of using `isOpen` property)

This will still have the extra padding in the device header, compared to the legacy example. We will likely need to capture that as a separate work item.

This PR is in favor of: #2317 and #2313.

### DIFF
--- a/docs/api/modal.md
+++ b/docs/api/modal.md
@@ -33,7 +33,7 @@ import APITOCInline from '@components/page/api/APITOCInline';
 A Modal is a dialog that appears on top of the app's content, and must be dismissed by the app before interaction can resume. It is useful as a select component when there are a lot of options to choose from, or when filtering items in a list, as well as many other use cases.
 
 
-## Inline Modals
+## Inline Modals (Recommended)
 
 `ion-modal` can be used by writing the component directly in your template. This reduces the number of handlers you need to wire up in order to present the modal.
 
--- a/docs/api/modal.md
+++ b/docs/api/modal.md
@@ -39,9 +39,9 @@ A Modal is a dialog that appears on top of the app's content, and must be dismis
 
 When using `ion-modal` with Angular, React, or Vue, the component you pass in will be destroyed when the modal is dismissed. As this functionality is provided by the JavaScript framework, using `ion-modal` without a JavaScript framework will not destroy the component you passed in. If this is a needed functionality, we recommend using the `modalController` instead.
 
-It is recommended to using triggers to open inline modals. A trigger is an element reference, that when selected will automatically open a modal. Triggers are useful when presenting a modal from a button or similar interaction.
+import InlineModalTriggerExample from '@site/static/usage/modal/inline/basic/index.md';
 
-TODO: Playground Example
+<InlineModalTriggerExample />
 
 ### Using `isOpen`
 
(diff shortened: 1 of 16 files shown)


static/usage/modal/inline/basic/angular/app_component_html.md
```diff
@@ -0,0 +1,30 @@
...
+<ion-app>
+  <ion-header>
+    <ion-toolbar>
+      <ion-title>Inline Modal</ion-title>
+    </ion-toolbar>
+  </ion-header>
+  <ion-content class="ion-padding">
+    <ion-button id="open-modal" expand="block">Open</ion-button>
+    <p>{{ message }}</p>
+    <ion-modal trigger="open-modal" (didDismiss)="onDidDismiss($event)">
```
liamdebeasi, line 11: Optional: I wonder if it would look nicer if we used willDismiss. The benefit is that the name will already have been rendered by the time the modal animation reveals the underlying page. (Unless the data is only passed back on didDismiss?)

static/usage/modal/inline/basic/angular/app_component_ts.md
```diff
@@ -0,0 +1,29 @@
...
+
+  cancel() {
+    this.modal.dismiss(null, 'cancel');
+  }
+
+  confirm() {
+    this.modal.dismiss(this.name, 'confirm');
+  }
+
+  onDidDismiss(ev: any) {
```
liamdebeasi, line 23: Can we use a stronger type? I believe the type here is `OverlayEventDetail<T>`

liamdebeasi (changes requested):

static/usage/modal/inline/basic/angular/app_component_ts.md
sean-perkins: This is due to this bug: https://github.com/ionic-team/ionic-framework/issues/24245. Do we want to show the workaround pattern instead here?

```ts
onDidDismiss(event: Event) {
  const ev = event as CustomEvent<OverlayEventDetail>;
}
```

docs/api/modal.md
```diff
@@ -39,17 +39,19 @@ A Modal is a dialog that appears on top of the app's content, and must be dismis
 
 When using `ion-modal` with Angular, React, or Vue, the component you pass in will be destroyed when the modal is dismissed. As this functionality is provided by the JavaScript framework, using `ion-modal` without a JavaScript framework will not destroy the component you passed in. If this is a needed functionality, we recommend using the `modalController` instead.
 
-It is recommended to using triggers to open inline modals. A trigger is an element reference, that when selected will automatically open a modal. Triggers are useful when presenting a modal from a button or similar interaction.
+import InlineModalTriggerExample from '@site/static/usage/modal/inline/basic/index.md';
 
-TODO: Playground Example
+<InlineModalTriggerExample />
```
amandaejohnston, line 44: In the Stackblitz for React, I get an error after clicking the Cancel button (`modal.dismiss is not a function`).

amandaejohnston (approved):

docs/api/modal.md
liamdebeasi, line 44: Do you think it is worth explicitly calling out that we recommend using `trigger`?

static/usage/modal/inline/basic/angular/app_component_html.md
```diff
@@ -0,0 +1,30 @@
...
+    </ion-toolbar>
+  </ion-header>
+  <ion-content class="ion-padding">
+    <ion-button id="open-modal" expand="block">Open</ion-button>
+    <p>{{ message }}</p>
+    <ion-modal trigger="open-modal" (willDismiss)="onWillDismiss($event)">
+      <ng-template>
+        <ion-header>
+          <ion-toolbar>
+            <ion-button slot="start" fill="clear" (click)="cancel()" color="medium">Cancel</ion-button>
```
liamdebeasi, line 15: Optional: The grayed out Cancel button leads me to believe it is supposed to be disabled, but it is not. Maybe a different color/font weight would be good? On iOS the Confirm button is usually a heavier font weight than the Cancel button.

liamdebeasi (approved):

PR #2323, opened 2022-05-04
docs(modal): can dismiss playground

sean-perkins:
Introduces the can dismiss examples for `boolean` and `function`. 

For the Angular Stackblitz examples, you will need to manually install the dev-build, if you want to confirm the behavior `6.1.4-dev.11651682390.1b48efad`. Likely their UI for adding dependencies will fail, as the dev build isn't included in their cached list. In this case, you will just need to save the Stackblitz example, manually update the values in the `package.json` and save & reload.

### DIFF
--- a/docs/api/modal.md
+++ b/docs/api/modal.md
@@ -74,7 +74,9 @@ Developers can set `canDismiss` to a boolean value. If `canDismiss` is `true`, t
 
 Setting a boolean value should be used when you need to require a particular action to be taken prior to a modal being dismissed. For example, if developers want to require that a "Terms of Use" checkbox is checked prior to closing the modal, they could set `canDismiss` to `false` initially and update it to `true` when the checkbox is checked.
 
-TODO: Playground Example
+import CanDismissBooleanExample from '@site/static/usage/modal/can-dismiss/boolean/index.md';
+
+<CanDismissBooleanExample />
 
 ### Setting a callback function
 
@@ -84,7 +86,9 @@ Setting a callback function should be used when you have complex dismissing crit
 
 Note that setting a callback function will cause the swipe gesture to be interrupted when using a card or sheet modal. This is because Ionic does not know what your callback function will resolve to ahead of time.
 
-TODO: Playground Example
+import CanDismissFunctionExample from '@site/static/usage/modal/can-dismiss/function/index.md';
+
+<CanDismissFunctionExample />
 
 ## Types of modals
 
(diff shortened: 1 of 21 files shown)


docs/api/modal.md
```diff
@@ -74,7 +74,9 @@ Developers can set `canDismiss` to a boolean value. If `canDismiss` is `true`, t
 
 Setting a boolean value should be used when you need to require a particular action to be taken prior to a modal being dismissed. For example, if developers want to require that a "Terms of Use" checkbox is checked prior to closing the modal, they could set `canDismiss` to `false` initially and update it to `true` when the checkbox is checked.
 
-TODO: Playground Example
+import CanDismissBooleanExample from '@site/static/usage/modal/can-dismiss/boolean/index.md';
+
+<CanDismissBooleanExample />
```
amandaejohnston, line 79: For the React and Vue Stackblitzes (and possibly Angular too), after closing the modal and re-opening it, the checkbox starts unchecked, but the modal can be closed: https://user-images.githubusercontent.com/90629384/166950991-18c6ed4f-3eeb-4a0a-84ac-9885b13543cd.mp4

amandaejohnston (changes requested):

docs/api/modal.md
sean-perkins: Good catch 🤔  for simplicity, I'm just going to persist the checked state of the `ion-checkbox`. So they can open the modal, check the checkbox and dismiss the modal. If they re-open the modal, the checkbox will already be pre-selected, since the state is persisted. This will visually make sure that that checked state and the dismiss button are always in sync.

amandaejohnston (approved):

static/usage/modal/can-dismiss/boolean/demo.html
```diff
@@ -0,0 +1,69 @@
...
+  <ion-app>
+    <ion-header>
+      <ion-toolbar>
+        <ion-title>App</ion-title>
+      </ion-toolbar>
+    </ion-header>
+    <ion-content class="ion-padding">
+      <ion-button id="open-modal" expand="block">Open</ion-button>
+
+      <ion-modal trigger="open-modal" swipe-to-close="true">
```
liamdebeasi, line 30: Swipe to close was deprecated, so I don't think we should be showing new devs how to use it.

liamdebeasi: Also the card modal looks weird: 
image

You probably need to wrap the header/content in `ion-app` in `.ion-page`.

liamdebeasi (approved):

PR #2327, opened 2022-05-05
docs(animations): update code snippets for modal animations

sean-perkins:
Fixes a few small issues with the code snippets on this page: https://ionicframework.com/docs/utilities/animations#modals

### DIFF
--- a/docs/utilities/animations.md
+++ b/docs/utilities/animations.md
@@ -1385,7 +1385,7 @@ export class ModalExample {
     }
 
     const leaveAnimation = (baseEl: any) => {
-      return this.enterAnimation(baseEl).direction('reverse');
+      return enterAnimation(baseEl).direction('reverse');
     }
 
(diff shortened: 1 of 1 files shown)


docs/utilities/animations.md
```diff
@@ -1385,7 +1385,7 @@ export class ModalExample {
     }
 
     const leaveAnimation = (baseEl: any) => {
-      return enterAnimation(baseEl).direction('reverse');
+      return this.enterAnimation(baseEl).direction('reverse');
```
amandaejohnston, line 1388: `enterAnimation` is declared in this scope (the `presentModal` method) rather than as a component method, so this change isn't needed (and will cause an error if pasted in).

amandaejohnston (changes requested):

amandaejohnston (approved):

PR #2329, opened 2022-05-09
fix(docs): format line breaks in autogenerated api docs

sean-perkins:
Fixes multi-line formatting of the auto-generated markdown table documentation.

Problem can be seen here:
- https://ionicframework.com/docs/api/refresher#events
- https://ionicframework.com/docs/api/range#events

This PR will format examples with consecutive line breaks, with line breaks in the markdown table. If a line only contains a single line break, it is likely as a result of us formatting line-wrapping and should be joined to the previous line with a space (see `ion-range` events as an example).

Applied this solution to all auto-generate sections that could have multi-line documentation.


liamdebeasi (approved):
nice job!

PR #2333, opened 2022-05-10
docs(angular, react, vue): add docs on linear vs non-linear routing

liamdebeasi:
This PR does a few things:

1. Defines router.go support for Angular, React, and Vue
2. Defines the difference between linear and non-linear routing as well as when to use each.

The goal with this PR is to help highlight some of the notable differences between web and mobile when building with Ionic.


amandaejohnston (changes requested):

docs/angular/navigation.md
```diff
@@ -169,6 +181,70 @@ Here, we have a typical Angular Module setup, along with a RouterModule import,
...
+  width="400"
+  src={useBaseUrl('video/linear-routing-demo.mp4')} 
+  controls
+></video>
+
+The application history in this example has the following path:
+
+`Accessibility` --> `VoiceOver` --> `Speech`
+
+When we press the back button, we follow that same routing path except in reverse. Linear routing is helpful in that it allows for simple and predictable routing behaviors. It also means we can use router Vue Router APIs such as [LocationStrategy.historyGo()](#navigating-using-locationstrategy).
```
amandaejohnston, line 206: Just noticed this; this is the Angular page but it mentions Vue.

amandaejohnston (approved):
Good to go once the below comment is addressed 👍

PR #2339, opened 2022-05-16
fix(angular/your-first-app): update what's next copy

AlexisM19:


sean-perkins (approved):

sean-perkins:
Thanks for fixing this!

PR #2342, opened 2022-05-16
feat(component-playground): typescript support for react examples

sean-perkins:
Updates the React Stackblitz output to use Typescript. 

You can preview the button playground example with this path: `/docs/playground/button`.


amandaejohnston (approved):
Honestly I think it would be better to go ahead and merge this and the one for Vue, so active work has TS available in Stackblitz for ensuring there aren't any errors.

PR #2347, opened 2022-05-18
docs(popover): add nested playground

amandaejohnston:
Adds the Nested Popovers playground. `dismissOnSelect` was folded in here rather than having its own separate playground, since separating the two felt redundant.


liamdebeasi (commented):
Clicking "More options..." does not dismiss the popover on the JavaScript and Angular StackBlitz demos but does dismiss the popover on the Vue and React demos

amandaejohnston:
Looks to be an Ionic bug 🤔 I'll investigate a little and get an issue posted.

amandaejohnston:
Done: https://github.com/ionic-team/ionic-framework/issues/25324

sean-perkins (approved):
Outside of the reported bug, everything else looks good to me 👍

liamdebeasi (approved):

PR #2348, opened 2022-05-19
docs(popover): add customization playgrounds

amandaejohnston:
- Positioning (Size and Alignment Demo)
- Sizing
- Styling

Also removes a dead link from the existing docs.

### DIFF
--- a/static/usage/popover/customization/styling/demo.html
+++ b/static/usage/popover/customization/styling/demo.html
@@ -1,0 +1,43 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <title>Popover</title>
+  <link rel="stylesheet" href="../../../common.css" />
+  <script src="../../../common.js"></script>
+  <script type="module" src="https://cdn.jsdelivr.net/npm/@ionic/core@6/dist/ionic/ionic.esm.js"></script>
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@ionic/core@6/css/ionic.bundle.css" />
+
+  <style>
+    ion-popover {
+      --background: rgba(40, 173, 218, 0.6);
+      --backdrop-opacity: 0.6;
+      --box-shadow: 0 5px 10px 0 rgba(0, 0, 0, 0.6);
+      --color: white;
+      --width: 300px;
+    }
+
+    ion-popover::part(backdrop) {
+      background-color: rgb(6, 14, 106);
+    }
+
+    ion-popover::part(content) {
+      padding: 16px 25px;
+    }
+  </style>
+</head>
+
+<body>
+  <ion-app>
+    <ion-content>
+      <div class="container">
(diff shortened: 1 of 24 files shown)


static/usage/popover/customization/styling/demo.html
```diff
@@ -0,0 +1,43 @@
...
+      --color: white;
+      --width: 300px;
+    }
+
+    ion-popover::part(backdrop) {
+      background-color: rgb(6, 14, 106);
+    }
+
+    ion-popover::part(content) {
+      padding: 16px 25px;
```
liamdebeasi, line 27: This should probably just be the `.ion-padding` class on `ion-content`.

static/usage/popover/customization/styling/demo.html
```diff
@@ -0,0 +1,43 @@
...
+  <title>Popover</title>
+  <link rel="stylesheet" href="../../../common.css" />
+  <script src="../../../common.js"></script>
+  <script type="module" src="https://cdn.jsdelivr.net/npm/@ionic/core@6/dist/ionic/ionic.esm.js"></script>
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@ionic/core@6/css/ionic.bundle.css" />
+
+  <style>
+    ion-popover {
+      --background: rgba(40, 173, 218, 0.6);
+      --backdrop-opacity: 0.6;
```
liamdebeasi, line 16: Any particular reason why you opted to use the backdrop opacity CSS variable instead of setting `opacity` on `::part(backdrop)`?

liamdebeasi (commented):
The Side and Alignment demos do not seem to render correctly in the framework integrations. Maybe a bug in Ionic?

Javascript: 

image


Angular: 
image

static/usage/popover/customization/styling/demo.html
amandaejohnston: Just to show that the variable exists, mostly 😆 I can change it if you like, although that makes me wonder why we have the variable in the first place 😄

liamdebeasi: It's up to you. The variable was added before shadow parts were available in most browsers. Though with the shadow part it's redundant.

amandaejohnston:
That misalignment does look like an Ionic bug, yeah. Interestingly, React and Vue show slightly different behavior, though all three frameworks are misaligned in some capacity. I'll get some repro apps together and post an issue.

static/usage/popover/customization/styling/demo.html
amandaejohnston: So it turns out that `ion-popover` has a default value for `--backdrop-opacity` built-in: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/popover/popover.ios.scss#L11 This overwrites setting `opacity` on the `ion-backdrop` directly, so setting a custom `--backdrop-opacity` on the popover is the cleanest way of doing things.

liamdebeasi (approved):

amandaejohnston:
Issue created: https://github.com/ionic-team/ionic-framework/issues/25337

sean-perkins (approved):

PR #2350, opened 2022-05-24
fix(playground): fall back to first tab if selectedValue isn't in tabs

amandaejohnston:
Currently in the playground, if you switch between two framework tabs (for example, Javascript and React) that both have multiple files in their code snippets, the active tab for the code files is not updated. This means that if the previously selected file name does not appear in the newly selected framework (very common), the new framework will not have any code tabs selected.

no selected tab

This happens because React detects that the only thing changed about the tabs is their children (the `<TabItem>s`), but the `PlaygroundTabs` component isn't designed to account for reactive children, so `selectedValue` doesn't update in response to it.

However, making `selectedValue` respond to dynamic children would require calling `setSelectedValue` during the render cycle. For example, this was my initial strategy:
```tsx
useEffect(() => {
  setSelectedValue(defaultValue);
}, [defaultValue]);
// then set defaultValue to first file name in Playground's renderCodeSnippets function
```
The problem with this is that it triggers a new render, which leads to visual flicker where the new default tab doesn't appear until a frame or two later. The same thing happens with similar strategies such as watching for updates to `values` or `children`.

This PR's current strategy, which falls back to the first tab via a manual check, ensures that the fallback happens in the same render cycle.

### DIFF
--- a/src/components/global/PlaygroundTabs/index.tsx
+++ b/src/components/global/PlaygroundTabs/index.tsx
@@ -110,6 +110,17 @@ function TabsComponent(props: Props): JSX.Element {
     setRightNavVisible(tabsNavEl.current?.scrollWidth > tabsNavEl.current?.offsetWidth);
   }, []);
 
+  /**
+   * If the selected value is not in the available tabs, fall back to the first tab.
+   * This can happen if the tab children are changed after the initial render.
+   * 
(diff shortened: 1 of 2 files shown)


sean-perkins (approved):
Tested the changes out and looks good to me 👍  nice work!

liamdebeasi (approved):

PR #2351, opened 2022-05-25
fix(component-playground): safe area padding

sean-perkins:
Updates the safe area values to match the existing docs examples.

### DIFF
--- a/static/usage/common.css
+++ b/static/usage/common.css
@@ -239,6 +239,6 @@
 }
 
 :root {
-  --ion-safe-area-top: 26px;
-  --ion-safe-area-bottom: 20px;
+  --ion-safe-area-top: 20px;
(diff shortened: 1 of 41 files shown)


sean-perkins:
Since the examples are loaded through an iframe, I don't believe the CSS variables will be inherited from the parent document. Is there another global stylesheet that you are referring to?

liamdebeasi:
Oh! I thought this was limited to the vanilla JS demo but it's not. The issue looks like the `ion-button` is not wrapped in `ion-buttons` in each of the demos.

So it should be 

```html
<ion-buttons slot="end">
  <ion-button>...</ion-button>
</ion-buttons>
```

Doing that will also let you drop the `fill="clear"`. Can you give that a try?

liamdebeasi (approved):

PR #2356, opened 2022-05-30
Clarify target file in step

synthmusic:
context of tutorial changed files, adding explicit clarification


synthmusic:
it's missing the import too, will resubmit with one commit

PR #2361, opened 2022-06-02
docs(range): range labels playground

sean-perkins:
Introduces the component playground examples for `ion-range` and using slots start/end.


docs/api/range.md
```diff
@@ -40,12 +40,11 @@ TODO: Playground Example
 
 ## Range Labels
 
-Labels can be placed on either side of the range by adding the
-`slot="start"` or `slot="end"` to the element. The element doesn't have to
-be an `ion-label`, it can be added to any element to place it to the
-left or right of the range.
+Labels and custom UI elements can be slotted on either side of the range by adding `slot="start"` or `slot="end"` to the element. The element can be any element, such as an `ion-label`, `ion-icon` or a `div`. Contents slotted to the `start` position will display to the left of the range, where as contents slotted to the `end` position will display to the right of the range.
```
liamdebeasi, line 43: "Contents slotted to the `start` position will display to the left of the range, where as contents slotted to the `end` position will display to the right of the range." This only true for LTR mode and will not be true for RTL mode.

liamdebeasi, line 43: "Contents slotted to the `start` position will display to the left of the range, where as contents slotted to the `end` position will display to the right of the range." This only true for LTR mode and will not be true for RTL mode.

liamdebeasi (changes requested):

docs/api/range.md
sean-perkins: Updated the copy to call-out the difference in behavior between LTR/RTL.

amandaejohnston (approved):

PR #2363, opened 2022-06-03
docs(breadcrumbs): add basic and icons playgrounds

amandaejohnston:
- Basic Usage
- Icons on Items
- Custom Separators


sean-perkins:
In strictly the demo examples, can we make it so that clicking "Home" does not shift the page?

amandaejohnston:
Good catch, will adjust the anchor link 👀

sean-perkins (approved):

PR #2369, opened 2022-06-07
docs(accordion): add readonly playgrounds

liamdebeasi:


docs/api/accordion.md
```diff
@@ -49,6 +49,24 @@ TODO Playground
...
 
+## Readonly Accordions
+
+### Individual Accordion
+
+Individual accordions can be disabled with the `readonly` property on `ion-accordion`.
+
+import ReadonlyIndividual from '@site/static/usage/accordion/readonly-individual/index.md';
+
+<ReadonlyIndividual />
```
amandaejohnston, line 60: Could we make this playground a little taller? There's a scroll bar popping up if you switch between the first and third accordions in `md` mode.

amandaejohnston (approved):
Good to go once the below is addressed.

sean-perkins (approved):

PR #2371, opened 2022-06-07
docs(range): pins playground

sean-perkins:
Introduces the component playground for displaying a pin with `ion-range`. 

Quick link: https://ionic-docs-git-range-playground-pins-ionic1.vercel.app/docs/api/range#pins

### DIFF
--- a/docs/api/range.md
+++ b/docs/api/range.md
@@ -58,6 +58,8 @@ TODO: Playground Example
 
 The `pin` attribute will display the value of the Range above the knob when dragged. This allows users to select a specific value within the Range.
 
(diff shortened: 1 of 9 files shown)


sean-perkins:
Need to add an example with `pinFormatter`, adding that soon!

amandaejohnston (approved):

liamdebeasi (changes requested):
The code snippets are missing `<IonContent>` so the range will appear in the middle of the screen instead of at the top.

liamdebeasi:
Good to merge once that is addressed though

sean-perkins:
Good catch, I'll update all open PRs with that change as well 👍

PR #2372, opened 2022-06-07
docs(range): dual knobs

sean-perkins:
Introduces the playground example for using dual knobs with `ion-range`.

### DIFF
--- a/docs/api/range.md
+++ b/docs/api/range.md
@@ -52,7 +52,19 @@ TODO: Playground Example
 
 ## Dual Knobs
 
-TODO: Playground Example
+Dual knobs introduce two knob controls that users can use to select a value at a lower and upper bounds. When selected, the Range will emit an `ionChange` event with a value object, containing the upper and lower values selected.
+
+```js
+// Example value event detail
+{
+  lower: 20,
+  upper: 80
+}
+```
+
+import DualKnobs from '@site/static/usage/range/dual-knobs/index.md';
+
+<DualKnobs />
 
 ## Pins
 
--- a/static/usage/range/dual-knobs/javascript.md
+++ b/static/usage/range/dual-knobs/javascript.md
@@ -1,0 +1,11 @@
+```html
+<ion-range dual-knobs="true"></ion-range>
+
+<script>
+  const range = document.querySelector('ion-range');
+  range.value = {
+    lower: 20,
+    upper: 80,
+  };
+</script>
+```
--- a/static/usage/range/dual-knobs/react.md
+++ b/static/usage/range/dual-knobs/react.md
@@ -1,0 +1,16 @@
+```tsx
+import React from 'react';
+import { IonRange } from '@ionic/react';
+function Example() {
+  return (
+    <IonRange
+      dualKnobs={true}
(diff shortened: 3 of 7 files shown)


static/usage/range/dual-knobs/vue.md
```diff
@@ -0,0 +1,14 @@
+```html
+<template>
+  <ion-range :dualKnobs="true" :value="{ lower: 20, upper: 80 }"></ion-range>
```
amandaejohnston, line 3:
```suggestion
  <ion-range :dual-knobs="true" :value="{ lower: 20, upper: 80 }"></ion-range>
```

amandaejohnston (changes requested):

docs/api/range.md
```diff
@@ -52,7 +52,19 @@ TODO: Playground Example
 
 ## Dual Knobs
 
-TODO: Playground Example
+Dual knobs introduce two knob controls that users can use to select a value at a lower and upper bounds. When selected, the Range will emit an `ionChange` event with a value object, containing the upper and lower values selected.
+
+```js
+// Example value event detail
```
liamdebeasi, line 58: Optional: We could also direct them to the `RangeValue` type: https://ionic-docs-git-range-playground-dual-knobs-ionic1.vercel.app/docs/api/range#rangevalue

liamdebeasi (changes requested):
The way we have our usage setup is a bit confusing. Some examples use `IonApp`, resulting in the component being rendered in the middle of the screen (due to missing `IonContent` usage) and others do not use `IonApp`. Can we make this consistent?

static/usage/range/dual-knobs/react.md
```diff
@@ -0,0 +1,16 @@
...
+import React from 'react';
+import { IonRange } from '@ionic/react';
+function Example() {
+  return (
+    <IonRange
+      dualKnobs={true}
+      value={{
+        lower: 20,
+        upper: 80,
+      }}
```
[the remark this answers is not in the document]
sean-perkins: After verification, this is an Ionic bug: https://github.com/ionic-team/ionic-framework/issues/25444

Unfortunately there isn't much of a workaround, outside of forcing a re-render of the component. For the sake of the example, the code demonstrated is "accurate" to how it should work once the bug is resolved.

amandaejohnston (approved):

static/usage/range/dual-knobs/javascript.md
```diff
@@ -0,0 +1,15 @@
+```html
+<ion-app>
```
liamdebeasi, line 2: Nonblocking: This might be best as a separate ticket, but it would be good if the Javascript stackblitz examples included `ion-app` automatically so we don't need to include them in each code snippet.

liamdebeasi (approved):

static/usage/range/dual-knobs/javascript.md
sean-perkins: I think this likely overlaps with FW-1668, but agree that `ion-app` should automatically be included in the template for all examples.

PR #2373, opened 2022-06-07
docs(breadcrumbs): add collapsing items playgrounds

amandaejohnston:
- Max Items
- Items Before or After Collapse
- Expand on Collapsed Indicator Click
- Popover on Collapsed Indicator Click


static/usage/breadcrumbs/collapsing-items/popover-on-click/demo.html
```diff
@@ -0,0 +1,76 @@
...
+    class PopoverPage extends HTMLElement {
+      constructor() {
+        super();
+      }
+
+      connectedCallback() {
+        let breadcrumbTemplate = ``;
+        this.collapsedBreadcrumbs.forEach(breadcrumb => {
+          breadcrumbTemplate += `
+            <ion-item href="${breadcrumb.href}">
```
liamdebeasi, line 43: Can we remove the line from the last item if it's easy to do? 
image

liamdebeasi: Is it possible to also use an inline popover here instead? Because the demo is lazy loaded, there's a delay between when you click the collapsed indicator and when the popover opens. Using the inline popovers means the popover is already available.

Same goes for the per-framework examples.

static/usage/breadcrumbs/collapsing-items/items-before-after/vue.md
```diff
@@ -0,0 +1,54 @@
+```html
+<template>
+  <ion-content>
+    <ion-label>Before Collapse = 2</ion-label>
```
liamdebeasi, line 4: Minor (and optional), but this is not the correct usage of `ion-label`. These should only be used in combination with components such as `ion-item` and `ion-input`: https://ionicframework.com/docs/api/label

We can probably just use a `div` or a heading element instead. Same goes for the other instances of `ion-label`.

docs/api/breadcrumbs.md
```diff
@@ -51,6 +51,40 @@ import CustomSeparators from '@site/static/usage/breadcrumbs/icons/custom-separa
...
+
+If there are more items than the value of `max-items`, the breadcrumbs will be collapsed. By default, only the first and last items will be shown.
+
+import MaxItems from '@site/static/usage/breadcrumbs/collapsing-items/max-items/index.md';
+
+<MaxItems />
+
+### Items Before or After Collapse
+
+Once the items are collapsed, the number of items to show can be controlled by the `items-before-collapse` and `items-after-collase` properties.
```
liamdebeasi, line 66:
```suggestion
Once the items are collapsed, the number of items to show can be controlled by the `itemsBeforeCollapse` and `itemsAfterCollapse` properties.
```

We typically refer to the property names using camel case as that is how they are documented in the "Properties" section.

docs/api/breadcrumbs.md
```diff
@@ -51,6 +51,40 @@ import CustomSeparators from '@site/static/usage/breadcrumbs/icons/custom-separa
 
 <CustomSeparators />
 
+## Collapsing Items
+
+### Max Items
+
+If there are more items than the value of `max-items`, the breadcrumbs will be collapsed. By default, only the first and last items will be shown.
```
liamdebeasi, line 58:
```suggestion
If there are more items than the value of `maxItems`, the breadcrumbs will be collapsed. By default, only the first and last items will be shown.
```

We typically refer to the property names using camel case as that is how they are documented in the "Properties" section.

liamdebeasi (changes requested):

liamdebeasi (approved):
Great job!

static/usage/breadcrumbs/collapsing-items/popover-on-click/angular/angular_ts.md
```diff
@@ -0,0 +1,21 @@
+```ts
+import { Component, ViewChild } from '@angular/core';
+
+@Component({
+  selector: 'app-root',
+  templateUrl: 'app.component.html',
+})
+export class AppComponent {
+  @ViewChild('popover') popover;
+  constructor() {}
```
sean-perkins, line 10: ~ We should probably remove the constructor when it isn't being used with providers. Angular style conventions recommend having public variables before the constructor.

sean-perkins (changes requested):
Small clean-up things. Overall these examples look great!

static/usage/breadcrumbs/collapsing-items/popover-on-click/vue.md
```diff
@@ -0,0 +1,48 @@
...
+        >
+          <ion-label>{{ breadcrumb.textContent }}</ion-label>
+        </ion-item>
+      </ion-list>
+    </ion-content>
+  </ion-popover>
+</template>
+
+<script lang="ts">
+import { IonBreadcrumb, IonBreadcrumbs, IonContent, IonItem, IonLabel, IonList } from '@ionic/vue';
```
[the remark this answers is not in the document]
amandaejohnston: It's not, good catch :eyes:

sean-perkins (approved):
🚀

PR #2374, opened 2022-06-07
docs(breadcrumbs): add styling playgrounds

amandaejohnston:
Adds the Color Property and CSS Properties playgrounds.

### DIFF
--- a/static/usage/breadcrumbs/styling/color/vue.md
+++ b/static/usage/breadcrumbs/styling/color/vue.md
@@ -1,0 +1,19 @@
+```html
+<template>
+  <ion-breadcrumbs color="primary">
+    <ion-breadcrumb href="#home">Home</ion-breadcrumb>
+    <ion-breadcrumb href="#electronics">Electronics</ion-breadcrumb>
+    <ion-breadcrumb href="#cameras">Cameras</ion-breadcrumb>
+    <ion-breadcrumb href="#film">Film</ion-breadcrumb>
+  </ion-breadcrumbs>
+</template>
+
+<script lang="ts">
+  import { IonBreadcrumb, IonBreadcrumbs } from '@ionic/vue';
+  import { defineComponent } from 'vue';
+
+  export default defineComponent({
+    components: { IonBreadcrumb, IonBreadcrumbs },
+  });
+</script>
+```
--- a/static/usage/breadcrumbs/styling/color/vue.md
+++ b/static/usage/breadcrumbs/styling/color/vue.md
@@ -1,10 +1,10 @@
 ```html
 <template>
-  <ion-breadcrumbs color="primary">
-    <ion-breadcrumb href="#home">Home</ion-breadcrumb>
-    <ion-breadcrumb href="#electronics">Electronics</ion-breadcrumb>
-    <ion-breadcrumb href="#cameras">Cameras</ion-breadcrumb>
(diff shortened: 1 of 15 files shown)


static/usage/breadcrumbs/styling/css-props/demo.html
```diff
@@ -0,0 +1,36 @@
...
+  <title>Breadcrumbs</title>
+  <link rel="stylesheet" href="../../../common.css" />
+  <script src="../../../common.js"></script>
+  <script type="module" src="https://cdn.jsdelivr.net/npm/@ionic/core@6/dist/ionic/ionic.esm.js"></script>
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@ionic/core@6/css/ionic.bundle.css" />
+
+  <style>
+    ion-breadcrumb {
+      --color: rgb(81, 155, 198);
+      --color-active: rgb(150, 112, 220);
```
liamdebeasi, line 16: Can we also add `--color-hover` too? Otherwise the text turns to black when you hover over it

static/usage/breadcrumbs/styling/color/vue.md
```diff
@@ -0,0 +1,19 @@
+```html
+<template>
+  <ion-breadcrumbs color="primary">
```
liamdebeasi, line 3: The Vue example does not work 
image

liamdebeasi (changes requested):

static/usage/breadcrumbs/styling/css-props/index.md
```diff
@@ -0,0 +1,30 @@
...
+import angularHTML from './angular/angular_html.md';
+import angularCSS from './angular/angular_css.md';
+
+<Playground
+  code={{
+    javascript,
+    react: {
+      files: {
+        'src/main.tsx': reactTS,
+        'src/main.css': reactCSS
```
sean-perkins, line 18: ~ Since the primary focus in these examples is the stylesheet (when there is multiple files), would it make sense to show the CSS file example first? The component playground example will render the files in the order they are defined in the object, so moving the CSS file above the tsx/html (for Angular) would render the stylesheet first.

sean-perkins (approved):
Good to go once the `--color-hover` piece is addressed. Agree that a different color would display better over the fallback to black here.

static/usage/breadcrumbs/styling/color/vue.md
sean-perkins: Is this possibly a bug? I'm seeing the color class being applied to `ion-breadcrumbs`, but `ion-breadcrumb` has it's own host level CSS variable values for the colors. 

Looks that `.in-breadcrumbs-color` is not being applied to the individual `ion-breadcrumb` components (only in the Vue example).

static/usage/breadcrumbs/styling/css-props/index.md
amandaejohnston: Good thinking!

static/usage/breadcrumbs/styling/color/vue.md
amandaejohnston: Haven't been able to track down the cause; went ahead and made an issue for it: https://github.com/ionic-team/ionic-framework/issues/25446 I'll move the `color` attribute to the inner `ion-breadcrumb` elements as a workaround for now.

liamdebeasi (approved):

PR #2385, opened 2022-06-13
docs(datetime): add preferWheel playground

liamdebeasi:
Part of https://github.com/ionic-team/ionic-framework/issues/23981

Adds a playground that shows how to use preferWheel

### DIFF
--- a/docs/api/datetime.md
+++ b/docs/api/datetime.md
@@ -231,6 +231,12 @@ Certain `presentation` options have both grid and wheel styles that developers c
 | `time-date`    | Yes             | Yes              |
 | `year`         | No              | Yes              |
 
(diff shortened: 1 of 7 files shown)


liamdebeasi:
Note: The stackblitz examples load the latest stable version of Ionic, so they will not know how to handle `preferWheel`.

sean-perkins (approved):
Playground looks good 👍  seeing a live example like this, does make me question if there should be some sort of loading indication in the future, for when rendering the next/previous months & days from the intersection observer hitting the threshold.

amandaejohnston (approved):

liamdebeasi:
FYI I'm going to change the CDN url back to the latest stable version just so I don't forget to do it when we release 6.2

PR #2414, opened 2022-06-21
fix(playground): tabs now have dark theme

liamdebeasi:
Light Mode (note: there should be no changes):

| `main` | branch |
| ------ | ----- |
| light-before | light-after |

Dark Mode:

| `main` | branch |
| ----- | ----- |
| dark-before | Screen Shot 2022-06-21 at 12 31 07 PM |


amandaejohnston (approved):

sean-perkins (approved):
Looks good to me 👍 logged a dark theme bug with the datetime playground in our backlog. I know I haven't been checking all our examples against dark theme, so this PR was a good reminder.

liamdebeasi:
Yeah I noticed that but didn't have a chance to fix it. It looks like the dark mode we apply in `common.css` overrides the custom theming here. It might make sense to have a way of opt-ing out of the default dark mode theme.

PR #2419, opened 2022-06-24
docs(loading): add base playground setup

liamdebeasi:
- Removes inline TOC, usage examples, and docs demo
- Adds sections with TODOs for playgrounds


amandaejohnston (approved):

liamdebeasi:
I just realized we won't be able to show a demo of the inline loading because it does not exist for vanilla JS. Instead of a playground, I will have an explainer with some code snippets for now until vanilla JS gets inline loading.

PR #2421, opened 2022-06-24
docs(alert): add presenting playground/examples

amandaejohnston:
Adds the controller playground, and usage examples for an inline alert. (Inline alerts aren't available in vanilla JS, so we can't show a demo for it.)

### DIFF
--- a/static/usage/alert/presenting/controller/demo.html
+++ b/static/usage/alert/presenting/controller/demo.html
@@ -15,7 +15,7 @@
   <ion-app>
     <ion-content>
       <div class="container">
-        <ion-button>Click Me</ion-button>
+        <ion-button onclick="presentAlert()">Click Me</ion-button>
       </div>
     </ion-content>
   </ion-app>
@@ -31,9 +31,6 @@
       document.body.appendChild(alert);
       await alert.present();
     }
-
-    const button = document.querySelector('ion-button');
-    button.addEventListener('click', presentAlert);
   </script>
 </body>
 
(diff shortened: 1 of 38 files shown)


static/usage/alert/presenting/controller/demo.html
```diff
@@ -0,0 +1,52 @@
...
+      alert.header = 'Alert';
+      alert.subHeader = 'Important message';
+      alert.message = 'This is an alert!';
+      alert.buttons = ['OK'];
+
+      document.body.appendChild(alert);
+      await alert.present();
+
+      const { role } = await alert.onDidDismiss();
+      output.innerText = `Alert dismissed with role: ${role}`;
```
sean-perkins, line 44: I feel like we should not show the role in these examples, since it will always be `undefined`.

amandaejohnston: You can also get it to show `backdrop` if you click that. I could add a role to the OK button if you'd prefer?

liamdebeasi: It might be worthwhile to make this a destroy/cancel alert where each button has a special role: https://ionicframework.com/docs/api/modal

amandaejohnston: Hmm, I think I'll just remove the role display from this basic example, and add it to the Buttons playground instead. That would make for a better separation of concerns and avoid the `undefined` role showing up.

liamdebeasi (approved):

sean-perkins (approved):
Looks good 👍

PR #2422, opened 2022-06-24
fix(playground): lazy load playgrounds

liamdebeasi:
Got some feedback (https://forum.ionicframework.com/t/problem-loading-example-docs/224503/6) that our soft launch of the playgrounds is causing a bit of slow downs. We have previously discussed lazily loading the playground content, so I made a PR to do that to alleviate some of the performance issues.


sean-perkins (approved):
Verified on a few different playground examples and it works nicely. 

We may want to consider a loading spinner/state for examples not using the device frame in the future. For example with the datetime example, it took several seconds for the playground to render on a fast connection.

liamdebeasi:
Sounds good. We have an existing ticket for that which I bumped to the top of our testing backlog.

amandaejohnston (approved):

PR #2423, opened 2022-06-24
docs(alert): add buttons playground

amandaejohnston:


sean-perkins (changes requested):

static/usage/alert/buttons/javascript.md
```diff
@@ -0,0 +1,30 @@
...
+        handler: () => { output.innerText = 'Alert confirmed'; }
+      }
+    ];
+
+    document.body.appendChild(alert);
+    await alert.present();
+  }
+
+  const button = document.querySelector('ion-button');
+  button.addEventListener('click', presentAlert);
```
liamdebeasi, line 28:
```suggestion
```

Can we use an `onclick` handler instead?

liamdebeasi (changes requested):

liamdebeasi (approved):

static/usage/alert/buttons/angular/angular_ts.md
```diff
@@ -0,0 +1,38 @@
+```ts
+import { Component } from '@angular/core';
+import { AlertController } from '@ionic/angular';
+
+@Component({
+  selector: 'app-root',
+  templateUrl: 'app.component.html',
+})
+export class AppComponent {
+  handlerMessage = "";
```
sean-perkins, line 10: ~ Minor nitpick, this and `roleMessage` could be using single quotes instead of double quotes.

sean-perkins (approved):

PR #2425, opened 2022-06-25
Fix capacitor plugins installation command

yohav:
Need to save dependencies, otherwise ```npx cap sync``` won't sync the plugins


jcesarmobile:
Thanks for the PR, but `--save` Is only needed on npm <= 4, lates npm is 8.
Capacitor 3 requires node 12 or newer, which used npm 6 by default, so nobody should need `--save` anymore.
If your node version is older than 5, please, update it.
If you are using npm 5 or newer check your npm settings are you might have changed them to not save, run `npm config get save` to check current value and `npm config set save=true` to make saving the default behavior again.

PR #2427, opened 2022-06-27
docs(playground): fix datetime and select react urls

liamdebeasi:
Datetime and select still had some pre-TypeScript URLs for React. Currently this causes the apps to fail to load on StackBlitz. I updated the config for these playgrounds with the correct URLs.


sean-perkins (approved):
Both of these playgrounds look good 👍

PR #2428, opened 2022-06-27
docs(alert): add customization playground

amandaejohnston:
Adds the Customization playground and removes a couple of dead anchor links from the docs.

### DIFF
--- a/docs/api/alert.md
+++ b/docs/api/alert.md
@@ -131,7 +131,7 @@ import Radios from '@site/static/usage/alert/inputs/radios/index.md';
 
 Alert uses scoped encapsulation, which means it will automatically scope its CSS by appending each of the styles with an additional class at runtime. Overriding scoped selectors in CSS requires a [higher specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) selector.
 
-We recommend passing a custom class to `cssClass` in the `create` method and using that to add custom styles to the host and inner elements. This property can also accept multiple classes separated by spaces. View the [Usage](#usage) section for an example of how to pass a class using `cssClass`.
+We recommend passing a custom class to `cssClass` in the `create` method and using that to add custom styles to the host and inner elements. This property can also accept multiple classes separated by spaces.
 
 ```css
 /* DOES NOT WORK - not specific enough */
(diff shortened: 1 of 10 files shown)


static/usage/alert/customization/angular/angular_ts.md
```diff
@@ -0,0 +1,28 @@
+```ts
+import { Component } from '@angular/core';
+import { AlertController } from '@ionic/angular';
+
+@Component({
+  selector: 'app-root',
+  templateUrl: 'app.component.html',
+})
+export class AppComponent {
+  constructor(public alertController: AlertController) {}
```
sean-perkins, line 10:
```suggestion
  constructor(private alertController: AlertController) {}
```

docs/api/alert.md
```diff
@@ -154,9 +154,15 @@ Any of the defined [CSS Custom Properties](#css-custom-properties) can be used t
 ```
 
 :::note
- If you are building an Ionic Angular app, the styles need to be added to a global stylesheet file. Read [Style Placement](#style-placement) in the Angular section below for more information.
+ If you are building an Ionic Angular app, the styles need to be added to a global stylesheet file.
 :::
 
+### Example
```
sean-perkins, line 160: If this example is focused around styling, can we update this heading to be "Theming" or "Styling"?

liamdebeasi (changes requested):
Can we make this playground look a bit more "real world"? For example, Chakra UI has a nice sample here: https://chakra-ui.com/docs/components/alert-dialog/usage.

Optional: It might be valuable to keep the theming consistent. For example, Tailwind's samples all have consistent themes with coordinated colors: https://tailwindcss.com/ (in the "Build whatever you want, seriously." section)

amandaejohnston:
Good point! Changed to use the `success` colors instead 👍

sean-perkins:
Screen Shot 2022-06-30 at 12 50 38 PM

The iOS preview could still use some refinement. The cancel button is clipping and the confirm button border radius does not match consistently.

amandaejohnston:
Ah, that'd be from the buttons wrapping; your screen size is likely a bit smaller than mine. I'll just make it a simple yes/no to avoid that.

sean-perkins (approved):
Alert looks great now 👍

PR #2432, opened 2022-06-29
docs(modal, popover): add keepContentsMounted example

liamdebeasi:


docs/api/popover.md
```diff
@@ -233,6 +233,23 @@ type PositionAlign = 'start' | 'center' | 'end';
...
+
+The content of an inline `ion-popover` is unmounted when closed. If this content is expensive to render, developers can use the `keepContentsMounted` property to mount the content as soon as the popover is mounted. This can help optimize the responsiveness of your application as the inner contents will have already been mounted when the popover opens.
+
+import Mount from '@site/static/usage/popover/performance/mount/index.md';
+
+<Mount />
+
+Developers should keep the following in mind when using `keepContentsMounted`:
+
+- This feature should be used as a last resort in order to deal with existing performance problems. Try to identify and resolve performance bottlenecks before using this feature. Additionally, do not use this to anticipate performance problems.
```
sean-perkins, line 248: ~ Would there be any value in mentioning that life cycle events related to components mounting will/can fire before the component is anticipated to become visible, since it would be mounted when the modal is mounted? This would also apply to popover.

Thought process is making it clear that `ngOnInit` in an Angular setting is going to fire when the modal is mounted, not when the modal is presented. So developers should be cautious with what functionality they are using in those life cycle methods.

liamdebeasi: good point. I will add a note

sean-perkins (approved):

PR #2438, opened 2022-07-07
fix(docs): add section headings inline

sean-perkins:
This PR does the following:
- Moves section headings for auto generated sections (Slots, Props, etc.) inline into main docs pages
- Removes auto generating section headings in the auto generated files
- Migrates existing playground pages not already using the sidebar navigation

While migrating these examples it occurred to me why I embedded the section headings originally. When generating the files, we can determine if a section heading should be rendered or not. Since we are moving those headings into the static doc pages, we no longer can easily do this. With the heading always being visible, now the script will generate placeholder sections when the component doesn't use the section, i.e.: "No events available for this component."


liamdebeasi (approved):

scripts/api.js
```diff
@@ -121,12 +115,10 @@ ${methods
 
 function renderParts({ parts }) {
   if (parts.length === 0) {
-    return '';
+    return 'No CSS Shadow Parts available for this component.';
```
amandaejohnston, line 118:
```suggestion
    return 'No CSS shadow parts available for this component.';
```

docs/api/breadcrumbs.md
```diff
@@ -100,9 +100,20 @@ import CSSProps from '@site/static/usage/breadcrumbs/styling/css-props/index.md'
 <CSSProps />
 
 
+## Properties
```
amandaejohnston, line 103: `ion-breadcrumbs` and `ion-picker` are still using the inline TOC (I'm assuming you're only migrating ones that already have playgrounds).

amandaejohnston (approved):
LGTM aside from these comments 👍

PR #2442, opened 2022-07-13
docs(datetime): add multiple selection playground

amandaejohnston:
PR for feature: https://github.com/ionic-team/ionic-framework/pull/25514
Dev build: `6.1.14-dev.11657720710.1df41c89`

The demo example is hard-coded to use the above dev build, but you'll need to install it on the Stackblitz examples manually.

### DIFF
--- a/docs/api/datetime.md
+++ b/docs/api/datetime.md
@@ -277,7 +277,7 @@ Developers can provide their own buttons for advanced custom behavior.
 
 ## Multiple Date Selection
 
-If the `multiple` property is set to `true`, multiple dates can be selected from the calendar picker at once. Clicking a selected date will deselect it.
+If the `multiple` property is set to `true`, multiple dates can be selected from the calendar picker. Clicking a selected date will deselect it.
 
 :::note
 This property is only supported when using `presentation="date"`.
--- a/docs/api/datetime.md
+++ b/docs/api/datetime.md
@@ -279,7 +279,9 @@ Developers can provide their own buttons for advanced custom behavior.
 
 If the `multiple` property is set to `true`, multiple dates can be selected from the calendar picker at once. Clicking a selected date will deselect it.
 
-Note that this property is only supported when `presentation="date"`.
+:::note
+This property is only supported when using `presentation="date"`.
+:::
 
 <MultipleDateSelection />
 
--- a/docs/api/datetime.md
+++ b/docs/api/datetime.md
@@ -34,6 +34,8 @@ import ShowingConfirmationButtons from '@site/static/usage/datetime/buttons/show
 import CustomizingButtons from '@site/static/usage/datetime/buttons/customizing-buttons/index.md';
 import CustomizingButtonTexts from '@site/static/usage/datetime/buttons/customizing-button-texts/index.md';
 
+import MultipleDateSelection from '@site/static/usage/datetime/multiple/index.md';
+
 import Theming from '@site/static/usage/datetime/theming/index.md';
 
 <head>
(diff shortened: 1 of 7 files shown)


static/usage/datetime/multiple/react.md
```diff
@@ -0,0 +1,20 @@
...
+import React, { useRef, useEffect } from 'react';
+import { IonDatetime } from '@ionic/react';
+
+function Example() {
+  const datetime = useRef<null | HTMLIonDatetimeElement>(null);
+
+  useEffect(() => {
+    if(!datetime.current) return;
+    datetime.current.value = ['2022-06-03', '2022-06-13', '2022-06-29'];
+  }, []);
```
amandaejohnston, line 11: This is done to get around this issue, which also applies here: https://github.com/ionic-team/ionic-framework/issues/25041

sean-perkins: Should we update that issue to mention it will affect `ion-datetime` with `multiple`? Also, should we add a comment in the playground so in the future when that issue is resolved, we can update this example and remove this code?

docs/api/datetime.md
```diff
@@ -273,6 +275,14 @@ Developers can provide their own buttons for advanced custom behavior.
 
 <CustomizingButtons />
 
+## Multiple Date Selection
+
+If the `multiple` property is set to `true`, multiple dates can be selected from the calendar picker at once. Clicking a selected date will deselect it.
+
+Note that this property is only supported when `presentation="date"`.
```
sean-perkins, line 282: ~ Docusaurus does have callout components if we wanted to highlight this visually:

```html
:::note
This property is only supported when using `presentation="date"`.
:::
```

static/usage/datetime/multiple/react.md
amandaejohnston: Good call 👀 I updated the issue and will add a tech debt ticket.

amandaejohnston: Ticket made: https://ionic-cloud.atlassian.net/browse/FW-1872 I skipped adding the comment since I didn't want it to show up in the actual docs site (and putting it in `demo.html` felt weird when it's not relevant to that), but I think the Jira ticket should be good enough for tracking this.

docs/api/datetime.md
```diff
@@ -273,6 +275,16 @@ Developers can provide their own buttons for advanced custom behavior.
 
 <CustomizingButtons />
 
+## Multiple Date Selection
+
+If the `multiple` property is set to `true`, multiple dates can be selected from the calendar picker at once. Clicking a selected date will deselect it.
```
sean-perkins, line 280: ~ The "at once" bit stuck out to me as possibly being confusing to the multiple feature. While technically you can select multiple dates at once with manually providing an array of values, the follow-up sentence makes me feel like it implies while clicking/interacting with the calendar picker.

Thoughts around omitting that piece and having it read similar to:
> If the `multiple` property is set to `true`, multiple dates can be selected from the calendar picker. Clicking a selected date will deselect it.

sean-perkins (approved):

docs/api/datetime.md
amandaejohnston: Works for me 👍

liamdebeasi (approved):
Optional: Thoughts on moving the multiple date selection up higher? Given the popularity of the feature ticket on GitHub, it seems like multiple date selection may be fairly common.

Otherwise this change looks great.

amandaejohnston:
Works for me 👀

amandaejohnston:
I moved it right underneath the `presentation` section, since the compatibility note references that.

PR #2446, opened 2022-07-15
chore(title): align component titles

rdlabo:
The component titles were in different formats and have been unified.


rdlabo:
It is closed because there are major conflicts.

PR #2448, opened 2022-07-15
fix(playground): remove code expand/collapse feature

amandaejohnston:
We've received feedback that the collapsing behavior is confusing; namely, all snippets should be expanded by default, and the JS framework tab should not appear selected if its code isn't expanded. To keep things simple, we decided to remove the expand/collapse feature entirely and wait for user feedback before iterating further.

### DIFF
--- a/src/components/global/Playground/playground.css
+++ b/src/components/global/Playground/playground.css
@@ -24,7 +24,7 @@
    * @prop --playground-btn-icon-color The text color of the button icon in the toolbar.
    * @prop --playground-btn-icon-hover-background The background color of the button icon in the toolbar when hovered.
    * @prop --playground-separator-color The color of the separator/border in the toolbar.
-   * @prop --playground-code-background The background color of the code block when expanded.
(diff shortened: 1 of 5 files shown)


src/components/global/Playground/playground.css
```diff
@@ -181,19 +181,8 @@
...
-   * the individual "\n" line breaks and the code snippet will be one line.
-   */
-  height: 0px;
-  overflow: hidden;
-}
-
-.playground__code-block--expanded {
-  height: initial;
   margin-top: 16px;
+  overflow: hidden;
```
sean-perkins, line 185: Is `overflow: hidden` still necessary? Originally it was added to hide the container until it was expanded. Since the container is now expanded by default (with no option to collapse), I don't believe we need to worry about overflow.

amandaejohnston: Good call!

liamdebeasi (approved):

PR #2451, opened 2022-07-19
docs(modal): handleBehavior playground example

sean-perkins:
Introduces a new section for the `handleBehavior` option available with Modal. 

Quick link: https://ionic-docs-git-docs-modal-handle-behavior-ionic1.vercel.app/docs/api/modal#handle-behavior

Misc:
- Fixed missing class for Angular background content example.

Note: The demo and Stackblitz examples are not using the dev-build. I can temporarily update the version they use, if that would be helpful for reviewing. 

Dev-build for testing: `6.1.14-dev.11658252995.1c2781cc`

### DIFF
--- a/docs/api/modal.md
+++ b/docs/api/modal.md
@@ -135,6 +135,14 @@ import SheetBackgroundContentExample from '@site/static/usage/modal/sheet/backgr
 
 <SheetBackgroundContentExample />
 
+#### Handle Behavior
+
+Sheet modals can optionally render a handle indicator used for dragging the sheet between breakpoints. The `handleBehavior` property can be used to configure the behavior of when the handle is tapped or clicked by the user.
+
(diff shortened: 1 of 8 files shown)


docs/api/modal.md
```diff
@@ -135,6 +135,14 @@ import SheetBackgroundContentExample from '@site/static/usage/modal/sheet/backgr
 
 <SheetBackgroundContentExample />
 
+#### Handle Behavior
+
+Sheet modals can optionally render a handle indicator used for dragging the sheet between breakpoints. The `handleBehavior` property can be used to configure the behavior of when the handle is tapped or clicked by the user.
```
liamdebeasi, line 140:
```suggestion
Sheet modals can optionally render a handle indicator used for dragging the sheet between breakpoints. The `handleBehavior` property can be used to configure the behavior of when the handle is activated by the user.
```

Small detail, but you can activate the handle with the keyboard too.

liamdebeasi (approved):

PR #2464, opened 2022-07-27
docs(datetime-button): add base usage example

liamdebeasi:
While I will not be able to finish every playground for `ion-datetime-button` by its release, I figured having a basic usage example was better than nothing.


docs/api/datetime-button.md
```diff
@@ -0,0 +1,81 @@
...
+
+The localized text on `ion-datetime-button` is determined by the `locale` property on the associated `ion-datetime` instance. See [Datetime Localization](./datetime#localization) for more details.
+
+## Usage with Modal
+
+`ion-datetime-button` must be associated with a mounted `ion-datetime` instance. As a result, [Inline Modals](./modal#inline-modals-recommended) with the `keepContentsMounted` property set to `true` must be used.
+
+## Usage with Popover
+
+`ion-datetime-button` must be associated with a mounted `ion-datetime` instance. As a result, [Inline Popovers](./popover#inline-popovers) with the `keepContentsMounted` property set to `true` must be used.
```
amandaejohnston, line 48: Would it be possible to merge these pieces together, since they're almost identical? I'm debating on whether they're even needed at all since this info is also mentioned in Overview. Also worth noting that separate playgrounds for both modal and popover would be partially redundant, since the Basic playground also shows how to use it with a modal. 🤔

scripts/api.js
```diff
@@ -18,7 +18,7 @@ const fs = require('fs');
  */
 
 (async function () {
-  const response = await fetch('https://unpkg.com/@ionic/docs/core.json');
+  const response = await fetch('https://unpkg.com/@ionic/docs@6.1.14-dev.11658247553.1b5403d1/core.json');
```
amandaejohnston, line 21: Don't forget to revert this before you merge 👀

amandaejohnston (approved):

PR #2465, opened 2022-07-27
chore(): add release-notes.json to gitignore

amandaejohnston:
This file is auto-generated and doesn't seem to actually be used when generating the corresponding page (https://ionicframework.com/docs/reference/release-notes) since the JSON is several versions behind.


liamdebeasi (approved):
You may need to also remove the `release-notes.json` file from the repo itself too.

amandaejohnston:
Good catch 👀

PR #2478, opened 2022-08-04
docs(support): update v5 to extended support only

amandaejohnston:
Updates the support table to list Ionic 5 as Extended Support Only, since maintenance ended in June.


liamdebeasi (approved):
Thanks for fixing this!

PR #2481, opened 2022-08-04
fix(playground): ensure dark mode loads on iframes at start

amandaejohnston:
Currently, if the page is set to dark mode on initial load, the playground iframes will still display in light mode. They only update when the mode is toggled on and off after the fact.

While the `useEffect` callback _does_ run on initial page load, the iframe refs haven't yet loaded, so the first run of the callback does nothing. The fix is to make use of a [callback ref](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs) instead, which allows additional side effects to run only after both refs are ready.

This PR also fixes the two async `useEffect` callbacks (both the one that handles the dark mode message, and the one below it that clears the iframe loading dots), as that pattern is not allowed. Further reading: https://devtrium.com/posts/async-functions-useeffect

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -138,8 +138,8 @@ export default function Playground({
     }
   };
 
-  const handleFrameRef = (ref: HTMLIFrameElement, frameMode: 'iOS' | 'md') => {
-    if (frameMode === 'iOS') {
+  const handleFrameRef = (ref: HTMLIFrameElement, frameMode: 'ios' | 'md') => {
+    if (frameMode === 'ios') {
       frameiOS.current = ref;
     } else {
       frameMD.current = ref;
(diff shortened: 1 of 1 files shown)


sean-perkins (approved):
Great work on this!

src/components/global/Playground/index.tsx
```diff
@@ -128,25 +128,58 @@ export default function Playground({
...
     if (frameiOS.current && frameMD.current) {
       await Promise.all([waitForFrame(frameiOS.current), waitForFrame(frameMD.current)]);
 
       const message = { darkMode: isDarkTheme };
       frameiOS.current.contentWindow.postMessage(message);
       frameMD.current.contentWindow.postMessage(message);
     }
+  };
+
+  const handleFrameRef = (ref: HTMLIFrameElement, frameMode: 'iOS' | 'md') => {
```
liamdebeasi, line 141:
```suggestion
  const handleFrameRef = (ref: HTMLIFrameElement, frameMode: 'ios' | 'md') => {
```

I recommend keeping the cases consistent here (similar to what we do in the framework)

liamdebeasi (approved):

PR #2498, opened 2022-08-17
docs(nav): component playground examples

sean-perkins:
This PR is a combination of other previously reviewed Nav playground PRs + fixing the examples to work with the revised "example component" refactor. 

[Preview](https://ionic-docs-git-docs-nav-playground-ionic1.vercel.app/docs/api/nav#navigation-within-a-modal)


amandaejohnston (approved):

docs/api/nav.md
```diff
@@ -24,19 +21,31 @@ import APITOCInline from '@components/page/api/APITOCInline';
...
 
+NavLink is a simplified API when interacting with Nav. Developers can customize the component, pass along component properties, modify the direction of the route animation or define a custom animation when navigating. 
 
-Nav is a standalone component for loading arbitrary components and pushing new components on to the stack.
+import NavLinkExample from '@site/static/usage/nav/nav-link/index.md';
 
-Unlike Router Outlet, Nav is not tied to a particular router. This means that if we load a Nav component, and push other components to the stack, they will not affect the app's overall router. This fits use cases where you could have a modal, which needs its own sub-navigation, without making it tied to the apps URL.
+<NavLinkExample />
+
+## Navigation within a Modal
```
liamdebeasi, line 36: It looks like the button is partially cut off in the Angular example: 
image

docs/api/nav.md
```diff
@@ -24,19 +21,31 @@ import APITOCInline from '@components/page/api/APITOCInline';
...
+
```
liamdebeasi, line 37: In the React example, it looks like the inner contents are not mounted until the modal is fully presented. Is that an Ionic bug?

docs/api/nav.md
```diff
@@ -24,19 +21,31 @@ import APITOCInline from '@components/page/api/APITOCInline';
...
+Modal can use Nav to offer a linear navigation that is independent of the URL.
```
liamdebeasi, line 38: The page transitions are not running in the Vue sample.

liamdebeasi (changes requested):
The NavLink demo looks great but the modal demo still has a few glitches

docs/api/nav.md
sean-perkins: In React and Vue, the usage of `ion-app` creates a container element around the children content, creating a DOM tree like:

```html
<ion-app class="ion-page">
  <div class="ion-page">
    <!-- component here -->
  </div>
</ion-app>
```

Angular does not do the same thing (likely we never noticed/wasn't an issue because `ion-router-outlet` handles this implementation detail. 

```html
<ion-app class="ion-page">
  <!-- component here -->
</ion-app>
```

Should I track this as a bug in Ionic or would we rather update our Playground demo code to workaround this issue (e.g.):
```html
<ion-app>
  <app-example class="ion-page"></app-example>
</ion-app>
```

Developers using `ion-router-outlet` should never run into this issue.

docs/api/nav.md
sean-perkins: Correct the React example uses `didPresent` instead of `willPresent`, because attempting to set the root component before the modal is presented does not work: https://stackblitz.com/edit/angular-vmakvy?file=src%2Fmain.tsx.

I can work to get a bug logged around this behavior.

docs/api/nav.md
sean-perkins: This appears to be a bug in the Vue implementation (or underlying Core bug), but I will dig in further and confirm. Seeing new page elements being constructed for each operation (pop/push/replace), but not seeing the pages unmounted or transitioned correctly.

brandyscarney:
Any update?

liamdebeasi:
ping

brandyscarney (approved):

PR #2502, opened 2022-08-22
docs(support): update JS framework and browser support

liamdebeasi:
Updates the following info for Ionic 7:

- Desktop browser support
- Mobile platform support
- JS Framework support for Angular, React, and Vue
- Updated the Angular 13 note to remove Ionic version-specific language. As a result, I moved it to the Platform support page.

JS Framework Support Policy: https://ionic-docs-git-fw-2005-ionic1.vercel.app/docs/reference/support
Browser Support: https://ionic-docs-git-fw-2005-ionic1.vercel.app/docs/reference/browser-support

### DIFF
--- a/docs/reference/support.md
+++ b/docs/reference/support.md
@@ -22,7 +22,8 @@ The current status of each Ionic Framework version is:
 
 | Version |        Status         |   Released   | Maintenance Ends | Ext. Support Ends |
 | :-----: | :-------------------: | :----------: | :--------------: | :---------------: |
-| V6      | **Active**            | Dec 8, 2021  | TBD              | TBD               |
+| V7      | **Active**            | TBD          | TBD              | TBD               |
+| V6      | Maintenance           | Dec 8, 2021  | TBD              | TBD               |
 | V5      | Extended Support Only | Feb 11, 2020 | June 8, 2022     | Dec 8, 2022       |
 | V4      | Extended Support Only | Jan 23, 2019 | Aug 11, 2020     | Sept 30, 2022     |
 | V3      | End of Support        | Apr 5, 2017  | Oct 30, 2019     | Aug 11, 2020      |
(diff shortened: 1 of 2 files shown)


docs/reference/support.md
```diff
@@ -42,27 +43,34 @@ The Ionic team has compiled a set of recommendations for using the Ionic Framewo
...
+| v6        | v12                     | v14.x^                  | 4.0+       |
+| v5        | v8.2                    | v12.x                   | 3.5+       |
+| v4        | v8.2                    | v11.x                   | 3.5+       |
+| v3        | v5.2.11                 | v5.2.11                 | 2.6.2      |
 
 > ^ Angular 14.x support was added in Ionic v6.1.9.
 
+**Angular 13+ Support On Older Versions of iOS**
+
+Angular's support policy for iOS is the two most recent major versions. This means that changes to your Angular project may be necessary to use Ionic Angular v4-v6 on iOS 13. To support iOS 13, change the project `target` specified in `compilerOptions` in the tsconfig.json to `es5`. Without this change an error of `Unexpected token '.' in promiseReactionJob` will occur on app startup in iOS 13.
```
amandaejohnston, line 56: Should the Ionic versions also include v7?

liamdebeasi: Ionic v7 does not support iOS 13, so this note does not apply. I'll make this clearer in the docs.

liamdebeasi: Oh actually we call out Ionic v4-v6. Can you think of another another way I could make this clearer?

amandaejohnston: Maybe something like this:
> Note that later versions of Ionic do not support iOS 13; see [mobile support table here](/docs/reference/browser-support).

Now that this section is on a different page from said table, the context is a bit lost.

liamdebeasi: Would it be better if I just keep the Angular note where it was before? It just seemed odd to an Angular 13 mention on a page that only talks about browsers/platforms.

amandaejohnston: Nah, I agree that this is the best place for it since it only pertains to Angular devs. I think the only way to get this completely clean would be to merge the two support pages into one, but that seems a little drastic 😆

amandaejohnston (approved):

PR #2506, opened 2022-08-24
docs(hardware-back-button): IonRouterOutlet is an optional provider

red1code:
in `docs/developing/hardware-back-button.md` , I found the following code for Angular, to exit from an Ionic  app:

```
import { IonRouterOutlet, Platform } from '@ionic/angular';
import { App } from '@capacitor/app';
...
constructor(
  private platform: Platform,
  private routerOutlet: IonRouterOutlet
) {
  this.platform.backButton.subscribeWithPriority(-1, () => {
    if (!this.routerOutlet.canGoBack()) {
      App.exitApp();
    }
  });
}
```
The problem is when you run `ionic serve` with that code you will get the following error:

```
ERROR NullInjectorError: R3InjectorError(AppModule)[IonRouterOutlet -> IonRouterOutlet -> IonRouterOutlet]: 
  NullInjectorError: No provider for IonRouterOutlet!
```
so I searched for a solution and I founed this post https://forum.ionicframework.com/t/nullinjectorerror-no-provider-for-ionrouteroutlet/192058/6 , which adds the `Optional()` decorator to the constructor, and that's what I did in my PR.

### DIFF
--- a/docs/developing/hardware-back-button.md
+++ b/docs/developing/hardware-back-button.md
@@ -274,7 +274,7 @@ import { App } from '@capacitor/app';
 
 constructor(
   private platform: Platform,
-  @Optional() private readonly routerOutlet?: IonRouterOutlet
+  @Optional() private routerOutlet?: IonRouterOutlet
 ) {
   this.platform.backButton.subscribeWithPriority(-1, () => {
     if (!this.routerOutlet.canGoBack()) {
--- a/docs/developing/hardware-back-button.md
+++ b/docs/developing/hardware-back-button.md
@@ -266,6 +266,7 @@ document.addEventListener('ionBackButton', (ev: BackButtonEvent) => {
 <TabItem value="angular">
 
 ```tsx
+import { Optional } from '@angular/core';
 import { IonRouterOutlet, Platform } from '@ionic/angular';
 import { App } from '@capacitor/app';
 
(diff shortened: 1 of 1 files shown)


docs/developing/hardware-back-button.md
```diff
@@ -266,14 +266,15 @@ document.addEventListener('ionBackButton', (ev: BackButtonEvent) => {
...
+import { Optional } from '@angular/core';
 import { IonRouterOutlet, Platform } from '@ionic/angular';
 import { App } from '@capacitor/app';
 
 ...
 
 constructor(
   private platform: Platform,
-  private routerOutlet: IonRouterOutlet
+  @Optional() private readonly routerOutlet?: IonRouterOutlet
```
sean-perkins, line 277:
```suggestion
  @Optional() private routerOutlet?: IonRouterOutlet
```

sean-perkins (changes requested):
Thanks for submitting this change! Can we update the provider to not be `readonly`? While semantically correct, this isn't a consistent pattern used in our examples and could be confusing to viewers to see differing patterns.

docs/developing/hardware-back-button.md
red1code: ok, no problem about not using `readonly`

sean-perkins (approved):

PR #2512, opened 2022-08-29
Use @capacitor/preferences in your-first-app

ImpulseDeer:
@capacitor/storage has been renamed @capcitor/preferences with the relase of version 4.
This is not update in this tutorial which will throw an ERR when trying to install the plugins.


jcesarmobile:
Covered here with all other usages of storage plugin
https://github.com/ionic-team/ionic-docs/pull/2499

PR #2515, opened 2022-08-30
docs(checkbox): prep for playgrounds

liamdebeasi:
- Remove inline TOC
- Remove phone demo
- Remove inline usage
- Add sections for upcoming playgrounds


sean-perkins (approved):
~ We could optionally delete the imports for `Tabs` and `TabItem` as well.

amandaejohnston (approved):

PR #2516, opened 2022-08-30
docs(toast): add playgrounds

amandaejohnston:
- Removes the phone demo and Usage section
- Moves TOC to the sidebar
- Adds all playgrounds

I elected to do an Icons playground instead of the Complex Toast Structure one planned in the design doc, because in hindsight I felt the original was encouraging a design pattern that would be better suited for `ion-alert`.


static/usage/toast/buttons/demo.html
```diff
@@ -0,0 +1,65 @@
...
+
+    .container p {
+      margin-bottom: 0;
+    }
+  </style>
+</head>
+
+<body>
+  <ion-app>
+    <ion-content>
```
liamdebeasi, line 26: Can we add a header to the demo? I know it's not needed for toast to work, it just looks a bit odd without a header in the device frame.

static/usage/toast/presenting/controller/vue.md
```diff
@@ -0,0 +1,26 @@
+```html
+<template>
+  <ion-button expand="block" @click="() => presentToast('top')">Present Toast At the Top</ion-button>
+  <ion-button expand="block" @click="() => presentToast('middle')">Present Toast At the Middle</ion-button>
+  <ion-button expand="block" @click="() => presentToast('bottom')">Present Toast At the Bottom</ion-button>
```
liamdebeasi, line 5: Our previous Vue code samples have not used arrow functions, so it would be good to stay consistent.

```suggestion
  <ion-button expand="block" @click="presentToast('top')">Present Toast At the Top</ion-button>
  <ion-button expand="block" @click="presentToast('middle')">Present Toast At the Middle</ion-button>
  <ion-button expand="block" @click="presentToast('bottom')">Present Toast At the Bottom</ion-button>
```

static/usage/toast/styling/javascript.md
```diff
@@ -0,0 +1,37 @@
+```html
+<ion-button onclick="presentToast()">Click Me</ion-button>
+
+<script>
+  async function presentToast() {
+    const toast = document.createElement('ion-toast');
```
liamdebeasi, line 6: While this works, we should be using the toast controller so the toast gets added as a child of `ion-app`.

static/usage/toast/icon/react.md
```diff
@@ -0,0 +1,24 @@
+```tsx
+import React, { useState } from 'react';
```
liamdebeasi, line 2:
```suggestion
import React from 'react';
```

liamdebeasi (changes requested):
Code looks good. Any chance we could spruce up the "Styling" section? Maybe we could use the toasts in Apple News as a reference? I'm more referring to the background/padding/box shadow/text color than the actual inner contents of the toast.

image

image

liamdebeasi (approved):
This looks great! Two small things, otherwise this is good on my end:

1. Can we make the "Hello Styled World" text darker? It's a bit hard to read: 
image


2. Can we change the "Styling" header to "Theming" based on the team discussion earlier today?

sean-perkins (approved):

PR #2530, opened 2022-09-02
docs(button): add components playgrounds

brandyscarney:
Question: should we remove the tables / descriptions under some of the headings?

I left them there to keep the additional information, but maybe we should either have them or not?


sean-perkins (approved):
The playground examples look great to me 👍 

Thoughts on eliminating the table sections and moving that information into the comment blocks in Ionic Framework, to be autogenerated as part of the "Properties" section?

This could help us prevent docs from becoming stale, since when introducing or removing property values in a feature, hopefully we would catch that the JSDoc comment was not updated with the new context.

brandyscarney:
@sean-perkins That's a good idea. I will look into adding that into the framework before removing from here & merging.

brandyscarney:
I removed the tables from the docs here and that info has been added to the properties themselves here: https://github.com/ionic-team/ionic-framework/pull/25940

PR #2538, opened 2022-09-08
docs(item): add component playgrounds

brandyscarney:
Adds the following playgrounds for item:

- Basic Usage
- Clickable Items 
- Detail Arrows
- Item Lines
- Media Items
- Buttons in Items
- Icons in Items
- Item Inputs
- Helper & Error Text
- Item Counter
- Theming
  - Colors
  - CSS Custom Properties
  - CSS Shadow Parts
  - Input highlighting
    - ✅ (MERGED) Depends on this PR https://github.com/ionic-team/ionic-framework/pull/25905

The following sections were removed:
- Item Placement
  - Slotted Content
  - Text Alignment
  
  https://ionic-docs-git-fw-1286-ionic1.vercel.app/docs/api/item


docs/api/item.md
```diff
@@ -70,2058 +82,91 @@ The below chart details the item slots and where it will place the element insid
...
-  <ion-button slot="start">
-    <ion-icon slot="icon-only" name="navigate"></ion-icon>
-  </ion-button>
-  <ion-label>Icon only Buttons</ion-label>
-  <ion-button slot="end">
-    <ion-icon slot="icon-only" name="star"></ion-icon>
-  </ion-button>
-</ion-item>
-```
+<HelperError />
```
amandaejohnston, line 128: Is the `ion-touched` logic needed here?

static/usage/item/helper-error/react.md
```diff
@@ -0,0 +1,41 @@
+```tsx
+import React, { useState } from 'react';
+import { IonInput, IonItem, IonLabel, IonNote } from '@ionic/react';
+
+function Example() {
+  const [isTouched, setIsTouched] = useState(false);
+  const [isValid, setIsValid] = useState(false);
+  const [isInvalid, setIsInvalid] = useState(false);
```
amandaejohnston, line 8: Since the input can't be valid and invalid at the same time, you could do this with just `isValid` and add the `ion-invalid` class if it's false.

static/usage/item/helper-error/react.md
```diff
@@ -0,0 +1,41 @@
...
+    setIsTouched(true);
+  }
+
+  return (
+    <>
+      <IonItem fill="solid" className={`${isValid && "ion-valid"} ${isInvalid && "ion-invalid"} ${isTouched && "ion-touched"}`}>
+        <IonLabel position="floating">Email</IonLabel>
+        <IonInput type="email" onIonInput={(event) => validate(event)} onIonBlur={() => markTouched()}></IonInput>
+        <IonNote slot="helper">Enter a valid email</IonNote>
+        <IonNote slot="error">Invalid email</IonNote>
```
amandaejohnston, line 35:
```suggestion
      <IonLabel position="floating">Email</IonLabel>
      <IonInput type="email" onIonInput={(event) => validate(event)} onIonBlur={() => markTouched()}></IonInput>
      <IonNote slot="helper">Enter a valid email</IonNote>
      <IonNote slot="error">Invalid email</IonNote>
```

amandaejohnston (changes requested):

static/usage/item/helper-error/react.md
brandyscarney: This indentation is needed since these are all children of the `IonItem`

docs/api/item.md
brandyscarney: Yeah so without adding `ion-touched` to the JavaScript, Vue, and React examples the label does not change to the proper color for an `ion-invalid` state. This results in the following focused & unfocused states:

Screen Shot 2022-09-13 at 7 18 02 PM
Screen Shot 2022-09-13 at 7 18 30 PM

Angular handles these states automatically so it looks like this for both focused & unfocused states:

Screen Shot 2022-09-13 at 7 19 23 PM

At first I wasn't adding `ion-touched` but I added them in after realizing this because Angular has the correct behavior. I figured we should match that for all of the other examples so users know what classes the validation / error text requires since this isn't well documented anywhere.

Also, when you first start typing in any of the examples, you'll see the error text / highlight turn red as soon as you start typing, but the label does not. This is actually a [framework bug](https://github.com/ionic-team/ionic-framework/issues/21643). The label has the correct behavior, the error text and highlight color should only display / change color after the `ion-touched` class has been added (so on blur of the input).

docs/api/item.md
```diff
@@ -58,8 +59,19 @@ See the [theming documentation](/docs/theming/css-variables) for more informatio
...
+<Lines />
+
+
 ## Item Placement
 
+### Slotted Content
+
 Item uses named [slots](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot) in order to position content. This logic makes it possible to write a complex item with simple, understandable markup without having to worry about styling and positioning the elements.
 
 The below chart details the item slots and where it will place the element inside of the item:
```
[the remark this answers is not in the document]
brandyscarney: I'm good with removing the section on slots, agree that it is redundant. Should we also remove the `Text Alignment` section under it? I believe we originally added this section because people were having trouble with wrapping the text in an item, so it was a way to link them to the docs on `.ion-text-wrap`, but I did use this class in the basic usage example so maybe that is sufficient now? The only other components where we link to the CSS Utilities are the content & grid components, and I can't see any other classes in there that are specifically helpful to item.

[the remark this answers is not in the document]
amandaejohnston: Good call, I agree with removing that section as well 👍 If we want to be safe, we could always add a callout for the text wrapping class above the Basic playground. That should make it easier to find while still cleaning things up.

static/usage/item/helper-error/react.md
amandaejohnston: Whoops, my bad, misread the HTML 😆 In that case you can remove the `<>`/`</>` here too.

amandaejohnston (approved):
LGTM once the comments are addressed 👍

PR #2545, opened 2022-09-13
Fix Capacitor link

RichardJECooke:


brandyscarney:
Thanks for the PR! This is actually in a lot of places in the docs, we should probably update all of them. cc @jaredcbaum

PR #2551, opened 2022-09-14
docs(item-sliding): add component playgrounds

brandyscarney:

- Basic Usage
- Icon Options
- Expandable Options

Removes the following sections:
- Options Layout (some of this section text was moved into the Icon Options section above)

------------------------------------------------

There is a note under Expandable Options which I kept in from the old documentation that says:

> This can be combined with the `ionSwipe` event on the [item options](https://ionicframework.com/docs/api/item-options) to call a method when the item is fully swiped.

However, we don't have any usage for the `ionSwipe` event (and didn't before) but I can add it if we think it would be helpful.


liamdebeasi (approved):
Changes look good 👍  Optional, but may be nice to have the phone frame for this so users can swipe less (due to phone frame being narrower than the width of the playground)

amandaejohnston (approved):

brandyscarney:
I messed around with putting them in a device preview but they just looked really empty with no header. If you feel strongly about it we can always update it!

PR #2552, opened 2022-09-14
docs(spinner): add component playgrounds

brandyscarney:
Adds playgrounds for:

- Basic Usage
- Theming
  - Colors
  - CSS Custom Properties

At first I had both playgrounds with the spinners in a list, but I changed the colors one to be more simple. If you think the first one shouldn't use a list either, or they both should, let me know. 

Corrects the usage section which stated the default for Material Design was `crescent` as it is now `circular`.

https://ionic-docs-git-fw-1265-ionic1.vercel.app/docs/api/spinner

### DIFF
--- a/static/usage/spinner/theming/css-properties/react.md
+++ b/static/usage/spinner/theming/css-properties/react.md
@@ -1,0 +1,13 @@
+```tsx
+import React from 'react';
+import { IonSpinner } from '@ionic/react';
+
+function Example() {
+  return (
+    <>
+      <IonSpinner></IonSpinner>
+    </>
+  );
+}
+export default Example;
+```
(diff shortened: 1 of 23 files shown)


amandaejohnston (approved):

brandyscarney:
I added a section for CSS Custom Properties to show `--color` being used and noticed that the documentation incorrectly stated `crescent` as the default for Material Design so I updated that and added the other missing spinners to Basic Usage.

static/usage/spinner/theming/css-properties/react.md
```diff
@@ -0,0 +1,13 @@
+```tsx
+import React from 'react';
+import { IonSpinner } from '@ionic/react';
+
+function Example() {
+  return (
+    <>
+      <IonSpinner></IonSpinner>
+    </>
```
amandaejohnston, line 9:
```suggestion
    <IonSpinner></IonSpinner>
```

amandaejohnston (approved):

PR #2569, opened 2022-09-20
docs(segment): add component playgrounds

brandyscarney:

- Basic Usage
- Scrollable Segments
- Toolbar Segments
- Theming
  - Colors
  - CSS Custom Properties
  
  https://ionic-docs-git-fw-1268-ionic1.vercel.app/docs/api/segment


liamdebeasi (commented):
Can we tweak the "CSS Custom Properties" container so the segment does not span the entire width of the container? 
image


Otherwise looks great!

amandaejohnston (approved):

PR #2571, opened 2022-09-21
docs(action-sheet): component playground examples

sean-perkins:
Introduces component playground examples for `ion-action-sheet`:
- Basic: Usage with many buttons, data and roles
- Theming > Styling: Example targeting specific elements to customize the appearance
- Theming > CSS Custom Properties: Example using CSS variables to customize the appearance

Quick preview link: https://ionic-docs-git-fw-1309-ionic1.vercel.app/docs/api/action-sheet


static/usage/action-sheet/theming/styling/demo.html
```diff
@@ -0,0 +1,91 @@
...
+  <script type="module" src="https://cdn.jsdelivr.net/npm/@ionic/core@6/dist/ionic/ionic.esm.js"></script>
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@ionic/core@6/css/ionic.bundle.css" />
+
+  <style>
+    ion-action-sheet.my-custom-class .action-sheet-group {
+      background: #F58840;
+    }
+
+    ion-action-sheet.my-custom-class .action-sheet-title {
+      color: #EADEDE;
```
liamdebeasi, line 19: Can we keep the title/subtitle as white? The gray is a bit hard to read on the orange background.

Also looks like some of the active indicators are not styled: 
image

liamdebeasi: Also why is the "Delete" button white on iOS and black on MD when the rest of the buttons are always black?

static/usage/action-sheet/basic/vue.md
```diff
@@ -0,0 +1,70 @@
...
+
+  code {
+    white-space: pre-wrap;
+  }
+</style>
+
+<template>
+  <div class="container">
+    <ion-button @click="presentActionSheet">Open</ion-button>
+    <code>{{result}}</code>
```
liamdebeasi, line 19:
```suggestion
    <code>{{ result }}</code>
```

static/usage/action-sheet/basic/vue.md
```diff
@@ -0,0 +1,70 @@
...
+  import { IonButton, actionSheetController } from '@ionic/vue';
+
+  export default {
+    components: { IonButton },
+    data() {
+      return {
+        result: '',
+      };
+    },
+    methods: {
```
liamdebeasi, line 33: Any particular reason why the Options API was chosen over the Composition API?

static/usage/action-sheet/theming/css-properties/demo.html
```diff
@@ -0,0 +1,66 @@
...
+      },
+      {
+        text: 'Cancel',
+        role: 'cancel',
+        data: {
+          action: 'cancel'
+        }
+      }];
+
+      document.body.appendChild(actionSheet);
```
liamdebeasi, line 60: This should be attached to `ion-app`. Alternatively, you can use the alertController which will do this for you. (same goes for the other instances of `document.body.appendChild`)

liamdebeasi (changes requested):

static/usage/action-sheet/theming/css-properties/demo.html
liamdebeasi: Hmm looks like alert also does `document.body`: https://ionicframework.com/docs/api/alert

I'd be fine doing `document.body` for now to be consistent and then revisiting the best practice later.

static/usage/action-sheet/basic/vue.md
sean-perkins: I thought we were waiting until all component playgrounds exist, before moving to the Composition API usage?

static/usage/action-sheet/theming/css-properties/demo.html
sean-perkins: Honestly - I referenced `ion-alert` examples when creating these examples. I can update both to match a specific usage pattern, if we have one.

static/usage/action-sheet/basic/vue.md
liamdebeasi: I think that was for using `script setup`, not the Composition API. Our current usage looks to be inconsistent:

Composition API: https://ionicframework.com/docs/api/accordion#toggle-accordions
Options API: https://ionicframework.com/docs/api/alert#controller

sean-perkins: Ahh good distinction - I'll get that fixed here + get `ion-alert` updated as well.

static/usage/action-sheet/theming/styling/demo.html
sean-perkins: Updated based on feedback.

The "Delete" button was white to show how to override the color opinionation we make for iOS mode. I removed this from the example, as the styles do not apply on MD & we do not currently support a CSS variable for developers to easily customize that appearance.

docs/api/action-sheet.md
```diff
@@ -1,12 +1,8 @@
 ---
 title: "ion-action-sheet"
-hide_table_of_contents: true
-demoUrl: "/docs/demos/api/action-sheet/index.html"
-demoSourceUrl: "https://github.com/ionic-team/ionic-docs/tree/main/static/demos/api/action-sheet/index.html"
 ---
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
```
amandaejohnston, line 5: `Tabs` and `TabItem` aren't used anymore here, so we can remove these imports.

docs/api/action-sheet.md
```diff
@@ -24,28 +20,26 @@ import EncapsulationPill from '@components/page/api/EncapsulationPill';
...
-  toc={toc}
-  maxHeadingLevel={2}
-/>
 
+An Action Sheet is a dialog that displays a set of options. It appears on top of the app's content, and must be manually dismissed by the user before they can resume interaction with the app. Destructive options are made obvious in `ios` mode. There are multiple ways to dismiss the action sheet, including tapping the backdrop or hitting the escape key on desktop.
 
+import Basic from '@site/static/usage/action-sheet/basic/index.md';
 
-An Action Sheet is a dialog that displays a set of options. It appears on top of the app's content, and must be manually dismissed by the user before they can resume interaction with the app. Destructive options are made obvious in `ios` mode. There are multiple ways to dismiss the action sheet, including tapping the backdrop or hitting the escape key on desktop.
+<Basic />
```
amandaejohnston, line 28: The result display doesn't seem to be working in the Vue Stackblitz 🤔

amandaejohnston (changes requested):

liamdebeasi (commented):
A few things I noticed:

1. In the "CSS Custom Properties" demo, the "Delete" button is red on iOS, but black on MD:

| ios | md |
| - | - |
| image | image |

Maybe all the button texts should be white?

2. In the "Styling" demo, the "Delete" button is white but all the other buttons are black. Maybe both the "Styling" and "CSS Custom Properties" demos should be aligned on button colors?

sean-perkins:
@liamdebeasi 

1. That is the default experience for destructive role buttons. On iOS we render the destructive role as red to align with the iOS design spec. MD does not have that opinionation. We do not expose a CSS variable to customize this appearance (other than overriding the danger color, which is only applied in iOS).
2. I can adjust this one, as we can target the selectors easily.

amandaejohnston (approved):

liamdebeasi:
> That is the default experience for destructive role buttons. On iOS we render the destructive role as red to align with the iOS design spec. MD does not have that opinionation. We do not expose a CSS variable to customize this appearance (other than overriding the danger color, which is only applied in iOS).

We already override the danger color in the "Styling" demo:

```css
ion-action-sheet.my-custom-class .action-sheet-button,
ion-action-sheet.my-custom-class .action-sheet-button.ion-focused {
  color: #000000;
}
```

image

The red button on the orange background is hard to read, so we should either make the button black or change the background color of the action sheet.

| red text | black text |
| - | - |
| Screen Shot 2022-09-26 at 9 53 48 AM | Screen Shot 2022-09-26 at 9 53 39 AM |

PR #2576, opened 2022-09-22
docs(toolbar): add component playgrounds

brandyscarney:

- Basic Usage (a toolbar in a header and footer)
- Buttons in Toolbars (button, back-button, and menu-button shown inside of toolbars)
  -  :warning: In order for this test to work in Angular, I had to add the `RouterModule` to the Angular Stackblitz default code. Otherwise the back button errors. I didn't think it made sense to create an entirely new app module for this test.
- Searchbars in Toolbars
- Segments in Toolbars
- Progress Bars in Toolbars
- Theming
  - Colors
  - CSS Custom Properties
  
  Preview: https://ionic-docs-git-fw-1256-ionic1.vercel.app/docs/api/toolbar

Removes the playground from:
- Segment / toolbar
- Searchbar / toolbar

And adds a playground pointing to the "Progress Bars in Toolbars" to Progress Bar docs


liamdebeasi (approved):

docs/api/toolbar.md
```diff
@@ -24,941 +17,76 @@ import EncapsulationPill from '@components/page/api/EncapsulationPill';
...
-| `start`      | Positions to the `left` of the content in LTR, and to the `right` in RTL.                                |
-| `end`        | Positions to the `right` of the content in LTR, and to the `left` in RTL.                                |
 
+## Buttons in Toolbars
 
-## Borders
+Buttons placed in a toolbar should be placed inside of the [buttons](./buttons) component. The buttons component can be positioned inside of the toolbar using a named [slot](#slots). The `"primary"` and `"secondary"` slots behave differently in `ios` and `md` mode. The buttons component can wrap a standard [button](./button), [back button](./back-button), a [menu button](./menu-button), or several of any of them. A button in a toolbar is styled to be clear by default, but this can be changed using the [`fill`](./button#fill) property on the button. Back buttons are linked to the app's navigation and will only show up when there is a previous page in the navigation stack, but we can bypass this by setting the `defaultHref` property. The menu button is set to only show when a linked menu is active, but this can be overridden by setting the `autoHide` property to `false`.
```
amandaejohnston, line 37:
```suggestion
Buttons placed in a toolbar should be placed inside of the [buttons](./buttons) component. The buttons component can be positioned inside of the toolbar using a named [slot](#slots). The `"primary"` and `"secondary"` slots behave differently in `ios` and `md` mode.

The buttons component can wrap a standard [button](./button), [back button](./back-button), a [menu button](./menu-button), or several of any of them. A button in a toolbar is styled to be clear by default, but this can be changed using the [`fill`](./button#fill) property on the button.
```
The descriptions of how `back-button` and `menu-button` work aren't needed since those should be covered by the respective component docs. (Also, the line break is optional but I think it makes this part easier to read 😄)

docs/api/toolbar.md
```diff
@@ -24,941 +17,76 @@ import EncapsulationPill from '@components/page/api/EncapsulationPill';
...
+import Buttons from '@site/static/usage/toolbar/buttons/index.md';
+
+<Buttons />
+
+
+## Searchbars in Toolbars
+
+A [searchbar](./searchbar) can be placed inside of a toolbar to search through the content. It should be the only child component of the toolbar, and will take up the full width and height.
+
+import Searchbars from '@site/static/usage/toolbar/searchbars/index.md';
```
amandaejohnston, line 48: Would it make sense to import the toolbar playground from `ion-searchbar` instead? (Or have searchbar import this one.) The two are more or less identical, so that would cut down on code we're maintaining. I could see it being confusing though, so I'm open to keeping both.

Ditto for the segments playground below, though you may need to merge the latest `main` into this branch first to grab those.

docs/api/toolbar.md
brandyscarney: Thanks! So I added that explanation in for the properties in menu button & back button because I had to use that in the code usage to avoid adding navigation or a menu to the examples. I thought adding all of that would be too much for the toolbar usage but wanted the back button & menu button to still show:

```
<ion-toolbar>
  <ion-buttons slot="start">
    <ion-back-button default-href="#"></ion-back-button>
  </ion-buttons>
  <ion-title>Back Button</ion-title>
</ion-toolbar>

<ion-toolbar>
  <ion-buttons slot="start">
    <ion-menu-button auto-hide="false"></ion-menu-button>
  </ion-buttons>
  <ion-title>Menu Button</ion-title>
</ion-toolbar>
```

Basically I was trying to let users know that they don't need to add this to their apps. Maybe there is a better way to get this point across? It might be bad to have because of people copy/pasting it without knowing why it's there.

docs/api/toolbar.md
brandyscarney: Oh that's a good idea. I'm fine with sharing them unless we think it will be confusing to maintain. Maybe a code comment above the import is sufficient to avoid confusion?

docs/api/toolbar.md
amandaejohnston: Maybe something like "for details on the properties used on `back-button` and `menu-button`, see their respective docs"? Reworded a little to be able to include inline links without it feeling awkward 😆 I just want to avoid duplicate info in case we change the props later.

docs/api/toolbar.md
```diff
@@ -24,941 +17,78 @@ import EncapsulationPill from '@components/page/api/EncapsulationPill';
...
-| `end`        | Positions to the `right` of the content in LTR, and to the `left` in RTL.                                |
 
+## Buttons in Toolbars
 
-## Borders
+Buttons placed in a toolbar should be placed inside of the [buttons](./buttons) component. The buttons component can be positioned inside of the toolbar using a named [slot](#slots). The `"primary"` and `"secondary"` slots behave differently in `ios` and `md` mode.
 
-In `md` mode, the `<ion-header>` will receive a box-shadow on the bottom, and the `<ion-footer>` will receive a box-shadow on the top.  In `ios` mode, the `<ion-header>` will receive a border on the bottom, and the `<ion-footer>` will receive a border on the top.
+The buttons component can wrap a standard [button](./button), [back button](./back-button), a [menu button](./menu-button), or several of any of them. A button in a toolbar is styled to be clear by default, but this can be changed using the [`fill`](./button#fill) property on the button. The properties included on back button and menu button in this example are for display purposes, see their respective documentation for proper usage.
```
amandaejohnston, line 39:
```suggestion
The buttons component can wrap a standard [button](./button), [back button](./back-button), a [menu button](./menu-button), or several of any of them. A button in a toolbar is styled to be clear by default, but this can be changed using the [`fill`](./button#fill) property on the button. The properties included on back button and menu button in this example are for display purposes; see their respective documentation for proper usage.
```

amandaejohnston (approved):

PR #2584, opened 2022-09-28
docs(card): add component playgrounds

brandyscarney:
Adds component playgrounds for:
- Basic Usage
- Media Cards
- Card Buttons
- List Card
- Theming
  - Colors
  - CSS Custom Properties

The first 3 are based on the Material Design card docs: https://material.io/components/cards#anatomy
The "List Card" is based on the iOS App Store card

https://ionic-docs-git-fw-1299-ionic1.vercel.app/docs/api/card

I had Ben review this one to make sure the mostly-default CSS styles were good enough and he approved. He requested the CSS to flip the title order between ios and md, but if we don't want it in the Stackblitz demos I could remove it.


sean-perkins (approved):

static/usage/card/basic/demo.html
```diff
@@ -0,0 +1,46 @@
...
+    .container {
+      max-width: 400px;
+      margin: 0 auto;
+    }
+
+    /* iOS places the subtitle above the title */
+    ion-card-header.ios {
+      display: flex;
+      flex-flow: column-reverse;
+    }
```
liamdebeasi, line 23: I know Ben requested this, but this seems wrong. If the default iOS layout in Ionic is incorrect, shouldn't we fix that internally instead?

brandyscarney: I thought about this too, but I was wondering if it would be too "magical" if we automatically switched them on users. Should we open discussion for it?

brandyscarney: I opened a pull request for implementing this in the framework: https://github.com/ionic-team/ionic-framework/pull/26084

The alternative is we continue to do as we did in past examples and switch the subtitle & title so it looks right in iOS and wrong in Material Design.

liamdebeasi (approved):

liamdebeasi:
Can we also make a tech debt ticket to remove the flex workarounds once https://github.com/ionic-team/ionic-framework/pull/26084 ships?

brandyscarney:
@liamdebeasi Done! FW-2578

PR #2590, opened 2022-09-30
docs(shadow-parts): remove errant '>' char

rwaskiewicz:
this commit removes a '>' that doesn't appear to be used/needed following an internal link


brandyscarney:
That was a feature

brandyscarney (approved):

PR #2593, opened 2022-09-30
+ docs(intro): change uninstall cli

cesvelasquez22:
Before really doesn't uninstall the cli.


jcesarmobile:
ionic was the name of the old CLI, was replaced with @ionic/cli long ago, but there are so much tutorials out there still using ionic as the name of the CLI that a lot of people is still installing and using it and that's why this intro tells you to uninstall it first and then install the newer @ionic/cli.

PR #2599, opened 2022-10-06
feat(playground): add support for opening in StackBlitz in ios mode

brandyscarney:
This PR passes the mode from the `ios` / `md` toggle on the playground to StackBlitz so you can see the examples in the same mode you have selected.

I know we discussed doing this by setting it on `ion-app` but I found two issues with that approach:

1) It doesn't work at all for React 
2) The styling is actually slightly off with it set on `ion-app`, see below

Buttons with `<ion-app mode="ios">`:
Screen Shot 2022-10-06 at 5 24 31 PM

Buttons with mode set via the config:
Screen Shot 2022-10-06 at 5 24 52 PM

So that's why I decided to set it via the config for all of the examples. Let me know if you think there's a better way.

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -249,6 +249,7 @@ export default function Playground({
       title,
       description,
       includeIonContent,
+      mode: isIOS ? 'ios' : 'md'
     };
 
     let codeBlock;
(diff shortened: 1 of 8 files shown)


sean-perkins (approved):
Looks good to me 👍

amandaejohnston (approved):

PR #2604, opened 2022-10-10
docs(keyboard): update typo in inputmode description

Shimin-Zhang:
fix a typo about inputmode description.


brandyscarney:
Thank you for the PR! Upon further review the original text is correct. If you look at the [inputmode](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode) documentation for `email` you will see it is not appropriate to use when an email address is required:

> A virtual keyboard optimized for entering email addresses. Typically includes the `@` character as well as other optimizations. Inputs that require email addresses should typically use `<input type="email">` instead.

PR #2609, opened 2022-10-12
chore(playground): update playground file names and imports to match best practices

brandyscarney:
Updates the Angular, React and Vue files to follow our best practices guide:

> *Angular:* Prefer the following format: `[name of component]_component_[type of file].md`
> 
> Example: `example_component.html.md`
> 
> *React:* Prefer the following format: `main_[type of file].md`
> 
> Example: `main_css.md`
> 
> *Vue*: Prefer the following format `[name of component]_vue.md`
> 
> Example: `modal_vue.md`


sean-perkins:
This may completely bust the JP build. Let's hold on merging until we can vet that. Otherwise we need to open an identical PR against that branch with the same file renames.

sean-perkins (approved):
I did more exploration in this and I think we are safe here! 

The main problem we run into with the JP build, is renaming the folder structure of where the Playground component is imported from in the docs page. Since we are changing the lower level files, I _think_ it'll be fine.

liamdebeasi (approved):

amandaejohnston (approved):

brandyscarney:
@sean-perkins Should I merge it and see if it breaks all of the things? 😬

sean-perkins:
@brandyscarney go for it! I can fix the JP build again over JEDI day if I was wrong 😆

PR #2613, opened 2022-10-17
feat(playground): disable specific modes

sean-perkins:
Adds the ability to disable specific modes for a playground. Currently showing this feature on the button basic and expanded examples. 

Example usage:

**Only available for iOS**
```tsx
<Playground mode="ios" />
```

**Only available for MD**
```tsx
<Playground mode="md" />
```

Quick preview: https://ionic-docs-git-fw-2662-ionic1.vercel.app/docs/api/button

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -13,11 +13,23 @@ import TabItem from '@theme/TabItem';
 
 import { IconHtml, IconTs, IconVue, IconDefault, IconCss, IconDots } from './icons';
 
-const ControlButton = ({ isSelected, handleClick, title, label }) => {
-  return (
+const ControlButton = ({
+  isSelected,
+  handleClick,
+  title,
+  label,
+  disabled,
+}: {
+  isSelected: boolean;
+  handleClick: () => void;
+  title: string;
+  label: string;
+  disabled?: boolean;
+}) => {
+  const controlButton = (
     <button
-      type="button"
-      title={title}
+      title={disabled ? undefined : title}
+      disabled={disabled}
       className={`playground__control-button ${isSelected ? 'playground__control-button--selected' : ''}`}
       onClick={handleClick}
       data-text={label}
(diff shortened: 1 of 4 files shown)


src/components/global/Playground/index.tsx
```diff
@@ -13,18 +13,39 @@ import TabItem from '@theme/TabItem';
...
+}: {
+  isSelected: boolean;
+  handleClick: () => void;
+  title: string;
+  label: string;
+  disabled?: boolean;
+}) => {
+  const controlButton = (
     <button
-      type="button"
```
amandaejohnston, line 19: Did you mean to remove the `type="button"` here?

src/components/global/Playground/index.tsx
```diff
@@ -105,20 +133,27 @@ export default function Playground({
     console.warn('No code usage examples provided for this Playground example.');
     return;
   }
+  if (typeof mode !== 'undefined' && mode !== 'ios' && mode !== 'md') {
+    console.warn(`Invalid mode provided: ${mode}. Accepted values are: "ios" or "md".`);
+    return;
```
amandaejohnston, line 138: Do we want to return here? Seems like we could fall back to default behavior to fail more gracefully instead.

src/components/global/Playground/index.tsx
sean-perkins: Yes. When not inside of a `<form>` element, there is no need to specify the type. I was being verbose in the original implementation and forgot to clean this up.

src/components/global/Playground/index.tsx
sean-perkins: I think returning makes the most sense so that we do not commit bad playground examples that use an invalid mode. 

The intention of using `mode` would be to restrict the example to one or the other, so falling back would likely result in showing an example that is confusing to users.

With returning early, the playground would not render (with an error logged to the console). This would hopefully be caught in local development and for sure would be caught in a PR.

amandaejohnston: Makes sense to me 👍

amandaejohnston (approved):

brandyscarney (approved):
Nice work! Looks good to me.

brandyscarney:
One request: can we add a ticket for going through to see which playgrounds need to use this? I know I have a few in header/footer that could use it. 🙂

PR #2620, opened 2022-10-19
docs(reorder): add component playgrounds

brandyscarney:
Adds component playgrounds for the following:

- Basic Usage (A reorder slotted inside of an item) 
- Toggling Reorder (Toggling disabled on reorder group)
- Custom Reorder Icon (A custom icon inside of the slotted reorder inside of an item)
- Reorder Wrapper (Reorder as a wrapper around an item)
- Updating Data (Rendering the items based on an array and updating that array on complete)
- Usage with Virtual Scroll (The basic reorder with a custom scroll target)

This also removes all usage examples from reorder group and links to reorder instead.

Preview Links: 
https://ionic-docs-git-fw-1272-ionic1.vercel.app/docs/api/reorder
https://ionic-docs-git-fw-1272-ionic1.vercel.app/docs/api/reorder-group


amandaejohnston:
Rather than use `console.log`, would it be possible to show that info in the playground itself? Here are a couple existing examples:
https://ionicframework.com/docs/api/action-sheet
https://ionicframework.com/docs/api/alert#buttons

I think it would also be sufficient to show the `event.detail.from` and `to` stuff in just the basic usage playground, or maybe a separate playground just to demonstrate that if you'd rather call it out more.

amandaejohnston (approved):

PR #2627, opened 2022-10-28
docs(menu): add component playgrounds for side & multiple

brandyscarney:
While working on the [menu playwright migration](https://github.com/ionic-team/ionic-framework/pull/26187) I noticed there were no usage examples in the docs on how to call open a menu when you have multiple menus.

This PR adds the following two playgrounds:
- Menu Sides
- Multiple Menus

The first one just shows an example of how to change the side to `end` which is more of a buildup to the next one. Multiple menus has two menus on the `start` side and one on the `end` side to show how you can use `MenuController` to open a menu by id or side. 

⚠️ The Angular & JavaScript StackBlitz are currently not working for multiple menus, investigating why.

### DIFF
--- a/docs/api/menu.md
+++ b/docs/api/menu.md
@@ -18,26 +18,25 @@ import EncapsulationPill from '@components/page/api/EncapsulationPill';
 <EncapsulationPill type="shadow" />
 
 
-The Menu component is a navigation drawer that slides in from the side of the current view.
-By default, it slides in from the left, but the side can be overridden.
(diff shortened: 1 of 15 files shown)


liamdebeasi:
If the multiple menu issue is a blocker, we can probably ship only the side playground.

thetaPC:
Verifying that it's still an issue on Angular & JavaScript StackBlitz.

brandyscarney:
Closing in favor of https://github.com/ionic-team/ionic-docs/pull/3049

PR #2629, opened 2022-10-28
docs(buttons): add component playgrounds

brandyscarney:
Adds the following component playgrounds:

- Basic Usage (a single `ion-buttons` component with an `ion-button` in a toolbar)
- Buttons Placement (`ion-buttons` with slots `start`, `end`, `secondary` and `primary`)
- Types of Buttons (a playground of default, outline, and solid buttons, buttons with icons, a back button & a menu button)
- Collapsible Buttons (this reuses the playground from https://ionicframework.com/docs/api/title#collapsible-buttons)

A few things to note:
- The [Buttons in Toolbars](https://ionicframework.com/docs/api/toolbar#buttons-in-toolbars) playground is essentially the buttons placement + types of buttons playground above. I wanted to split these though to really showcase the different positions buttons can have. We could just pull both in to the Toolbar docs if we wanted to reuse them. Lmk any thoughts.
- The Collapsible Buttons playground could be updated to use the word `Button` instead of `Click Me` if it would make it more clear.
- 
Preview Link: https://ionic-docs-git-fw-1300-ionic1.vercel.app/docs/api/buttons


liamdebeasi (commented):
I'm noticing that the ellipsis icon in "Types of Buttons" is horizontal in Angular/Vanilla JS and vertical in Vue/React despite the mode being the same.

docs/api/buttons.md
```diff
@@ -22,340 +17,56 @@ import EncapsulationPill from '@components/page/api/EncapsulationPill';
...
+:::info
+
+This feature is only available for iOS.
 
+:::
 
+<!-- Reuse the playground from the Title directory -->
+import CollapsibleLargeTitleButtons from '@site/static/usage/title/collapsible-large-title/buttons/index.md';
 
+<CollapsibleLargeTitleButtons />
```
amandaejohnston, line 68: Yeah, I think "Button" would be better here for the text; "Click Me" implies something should happen when you click it.

amandaejohnston (changes requested):

amandaejohnston (approved):

brandyscarney:
After investigation, the issue @liamdebeasi found where switching between the ellipsis icons was not working in React & Vue was found to be a bug. I created the following framework issue for this: https://github.com/ionic-team/ionic-framework/issues/26207

liamdebeasi (approved):

PR #2630, opened 2022-10-30
docs(react): provide unique key instead of index

ThejasKiranPS:
As providing index as keys is a bad practice, this commit modifies the key value to be `photo.filepath` which will be unique and consistent while adding new photos.

[Keys - react docs](https://reactjs.org/docs/lists-and-keys.html#keys) 
[An example for why it is not recommended](https://robinpokorny.com/blog/index-as-a-key-is-an-anti-pattern/)


sean-perkins (approved):

sean-perkins:
Thank you for fixing this!

PR #2633, opened 2022-10-31
docs(back-button): add component playgrounds

brandyscarney:
Adds the following playgrounds for both v6 and v7:

- Basic Usage (A back button with an `ion-nav`)
- Custom Back Button (A back button with an `ion-nav` and custom `text` and `icon`)

Known issue: in Angular and Vue, opening in the selected playground mode doesn't work due to a custom `app.module.ts` & `main.ts`.


Youngblood99 (approved):

amandaejohnston:
We elected to skip the `defaultHref` playground for now due to an issue that seems specific to Stackblitz. I'll be opening a separate PR to add that back in so we can use it to report the issue to them as needed.

liamdebeasi (approved):
Code looks good, though I noticed that the Angular stackblitz always starts in MD mode (even if iOS mode is selected in the playground)

amandaejohnston:
I mentioned that in the PR description as a known issue. It's something we've run into with other playgrounds too.

liamdebeasi:
🤦 Sorry totally missed that. Ok cool sounds good!

PR #2640, opened 2022-11-10
docs(datetime): confirmation buttons use the confirm api

### LINKED ISSUE #2637: content: Datetime doc mentions "done" and "cancel" methods
### URL

https://ionicframework.com/docs/api/datetime#showing-confirmation-buttons

### Issue Description

The link above mentions `done` and `cancel` methods triggered by the buttons, but there is no other mention of a `done` method in the documentation.   I played with the stackblitz a little bit, and couldn't get a public `cancel()` method on the component to be called, either.

Either this documentation isn't clear (what methods are Done and Cancel preconfigured to call), or it's inaccurate (I don't think they're actually preconfigured to do anything).

sean-perkins:
The `ion-datetime` documentation refers to the `done` API, which does not exist. This was a typo and was meant to reference the `confirm` API. I've additionally linked to these APIs to help developers navigating the docs. 

Resolves #2637

### DIFF
--- a/docs/api/datetime.md
+++ b/docs/api/datetime.md
@@ -265,7 +265,7 @@ By default, `ionChange` is emitted with the new datetime value whenever a new da
 
 ### Showing Confirmation Buttons
 
-The default Done and Cancel buttons are already preconfigured to call the `confirm` and `cancel` methods, respectively.
+The default Done and Cancel buttons are already preconfigured to call the [`confirm`](#confirm) and [`cancel`](#cancel) methods, respectively.
 
 <ShowingConfirmationButtons />
 
--- a/docs/api/datetime.md
+++ b/docs/api/datetime.md
@@ -265,7 +265,7 @@ By default, `ionChange` is emitted with the new datetime value whenever a new da
 
 ### Showing Confirmation Buttons
 
-The default Done and Cancel buttons are already preconfigured to call the `done` and `cancel` methods, respectively.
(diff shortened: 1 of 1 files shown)


amandaejohnston (commented):
Should we also specify that the default button behavior is to leave the overlay open?

sean-perkins:
@amandaejohnston the default behavior of the "Done" button is to dismiss the overlay on click: https://github.com/ionic-team/ionic-framework/blob/main/core/src/components/datetime/datetime.tsx#L1364

Are you referring to something else?

amandaejohnston:
Ah whoops, I was thinking the `cancel` method, since the issue poster was confused about that.

sean-perkins:
Personally I think we should be careful mixing the public API documentation with the UI implementation documentation. For example, clicking "Cancel" in the datetime implementation with `show-default-buttons`, does dismiss the overlay, but directly using the public API like the OP, will not unless you pass the additional argument to dismiss the overlay.

This is why I decided to link out to the API documentation from that section, so developers can decide if they need to use the public API method manually and if so, if they need to additional close the overlay. 

Let me know if you have ideas on how to better reflect that.

amandaejohnston:
Ahh okay, nah I just skimmed their issue too fast; I thought they were using the datetime's built-in implementation and didn't think to double-check that 😆 Sorry for the confusion, haha.

amandaejohnston (approved):

PR #2650, opened 2022-11-16
fix(docs): redirect for layout to grid

### LINKED ISSUE #2648: content: Vertical Grid Alignment page missing
### URL

https://ionicframework.com/docs/layout/grid#vertical-alignment

### Issue Description

Noticed this page is missing from the docs

sean-perkins:
The 301 redirect in nginx handles the server request when directly accessing the URL, but when using a client-side feature such as search; the redirect for `/docs/layout/grid` does not correctly redirect to `/docs/api/grid`. 

Resolves #2648

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -309,6 +309,17 @@ module.exports = {
         },
       };
     },
+    [
+      '@docusaurus/plugin-client-redirects',
+      {
+        redirects: [
+          {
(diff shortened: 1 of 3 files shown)


sean-perkins:
This does not appear to solve the issue (needed a live environment to verify). Will continue to dig in more as I have additional time.

PR #2653, opened 2022-11-17
docs(split-pane): add css properties file

liamdebeasi:
- Adds a playground for using css properties on split-pane
- Also removed a stray `angular.md` file in the `spinner` demo

Preview: https://ionic-docs-git-1264-css-ionic1.vercel.app/docs/api/split-pane#css-custom-properties


static/usage/split-pane/theming/css-properties/demo.html
```diff
@@ -0,0 +1,48 @@
...
+    <ion-app>
+      <ion-split-pane when="xs" content-id="main">
+        <ion-menu content-id="main">
+          <ion-header>
+            <ion-toolbar color="tertiary">
+              <ion-title>Menu</ion-title>
+            </ion-toolbar>
+          </ion-header>
+          <ion-content class="ion-padding">
+            Menu Content is 350px wide and has a blue dashed border
```
amandaejohnston, line 31: The text here is different from what the example code uses, which looks odd.

amandaejohnston (approved):
Looks good aside from the one comment 👍

static/usage/split-pane/theming/css-properties/demo.html
liamdebeasi: Good catch! Fixed in https://github.com/ionic-team/ionic-docs/pull/2653/commits/5acf12b8751ce529d78787fa0ed994ee29938d06.

PR #2656, opened 2022-11-24
Update 2-taking-photos.md

derf974:
Sorry I'm beginner with ionic ,
Tell me if i'm wrong but when i tried to do  the "Getting  Started", it's raised me an error  :
Type 'string | undefined' is not assignable to type 'string'.
[ng]   Type 'undefined' is not assignable to type 'string'.

I found this fix.

### DIFF
--- a/docs/angular/your-first-app/2-taking-photos.md
+++ b/docs/angular/your-first-app/2-taking-photos.md
@@ -108,7 +108,7 @@ Over in the `addNewToGallery` function, add the newly captured photo to the begi
 
   this.photos.unshift({
     filepath: "soon...",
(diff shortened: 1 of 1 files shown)


amandaejohnston:
Thank you for the PR! Looking into this further, it looks like the better approach is to mark `webviewPath` as optional in the `UserPhoto` interface, so you avoid the error without having to assert a variable that may not actually be defined. The React and Vue versions of the guide take this approach.

I've created a separate PR that makes this change instead: https://github.com/ionic-team/ionic-docs/pull/2926 You'll be given co-author credit when the PR is merged. Closing this one out in favor of the newer one.

PR #2659, opened 2022-11-28
docs(native): add Dynamsoft Barcode Scanner

xulihang:
Add Dynamsoft Barcode Scanner to the list of native plugins.


jcesarmobile:
Thanks for the pull request, but @awesome-cordova-plugins are no longer displayed on Ionic website

PR #2661, opened 2022-11-29
docs(theming): update vendor prefixed line height

web-padawan:
The [Vendor Prefixes](https://ionicframework.com/docs/theming/css-shadow-parts#vendor-prefixed-pseudo-elements) section in the docs has incorrect line height, apparently because it starts with `<a>` tag:

paragraph

Added the markup for the paragraph using `<p>` elements, hopefully this helps to fix it.


liamdebeasi:
Thanks for the PR! It looks like having an `<a>` as the first element on line causes the `<p>` to not get added. Manually adding the tag fixes the issue, and so I've merged this.

PR #2668, opened 2022-12-07
fix(playground): warn if version not specified

sean-perkins:
- Statically assigns the `version` for the v6 playground examples
- Removes the default `version` value for the `<Playground />` component
  - Not providing a version will result in a warning in the console.

Note: In the previous PR I incorrectly had the `version` typed applied to the `UsageTargetOptions`, instead of the playground properties.


amandaejohnston:
@sean-perkins Any blockers here? I forgot to add the `version` to my datetime playgrounds in a previous PR, so this warning would be useful to have.

PR #2669, opened 2022-12-08
docs(loading): inline playground usage

sean-perkins:
Adds an inline overlay example using triggers for `ion-loading`. Additionally updates the link references for stylesheets in the HTML example to use the dev-build.

Updates previous (non-controller) examples to use inline loading overlays. 

Note: Angular and JS Stackblitz examples do not load the correct npm version (this is an issue with Stackblitz). When viewing those examples, the overlay will not be presented when clicking the trigger element. 

Quick preview: https://ionic-docs-git-fw-2934-b-ionic1.vercel.app/docs/api/loading


sean-perkins:
Note to self: Before merging (after approval), update the dev-build for this example to match the existing dev-build from `main`. That way we can continue to do find/replace once v7 ships.

docs/api/loading.md
```diff
@@ -37,69 +34,9 @@ import Controller from '@site/static/usage/v7/loading/controller/index.md';
...
-    return { isOpenRef, setOpen }
-  }
-});
-</script>
-```
-</TabItem>
-</Tabs>
+import Inline from '@site/static/usage/v7/loading/inline/index.md';
+
+<Inline />
```
amandaejohnston, line 39: Do the React and Vue examples still need `ion-content`?

amandaejohnston, line 39: Could we also update the existing non-controller examples to show inline usage? Unless controller usage is our recommended strategy for `ion-loading`. We may also want an example for `isOpen`.

Worth noting that the overlays we've migrated so far have also included language on when to use inline vs. controller; example: https://ionicframework.com/docs/v7/api/action-sheet#inline-action-sheets-recommended

amandaejohnston: `ion-content` is still in the Vue example here 👀

amandaejohnston (approved):
LGTM aside from that last comment.

PR #2683, opened 2022-12-22
docs(angular): add missing non-null assertion operator

dfeliz:
## What

This PR adds a missing non-null assertion operator (it's being used earlier in the tutorial).

## Why

The non-null assertion operator is used in the `readAsBase64` method creation (`Saving Photos to the Filesystem` step) but is missing in the `Adding Mobile` step. Typescript complains about this.

image


mapsandapps (approved):
Thank you for the contribution!

PR #2685, opened 2022-12-23
docs(tabs): vue usage includes router outlet

NyllRE:
this isn't an edit about wrong stuff or not, but it's not clear why is it missing the ion-router-outlet while if I added it as-is it warns me of the lack of an ion router outlet. and it's confusing for why it's not referenced here. in addition to that it's not clear where should I add the tabs component and the schedule component etc

### DIFF
--- a/docs/api/tabs.md
+++ b/docs/api/tabs.md
@@ -393,6 +393,7 @@ import {
   IonIcon, 
   IonLabel, 
   IonPage,
+  IonRouterOutlet,
   IonTabBar, 
   IonTabButton, 
   IonTabs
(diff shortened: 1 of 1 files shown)


liamdebeasi (approved):
Thanks! I updated your PR to import `IonRouterOutlet` and provide it to Vue component, otherwise you will get an error on compilation.

liamdebeasi:
Merged. Thank you!

NyllRE:
Welcome, also I wanted to ask if this is also necessary in other frameworks? if so I hope people experienced with those frameworks implement it too so no further confusion is left with the ionic tabs

PR #2691, opened 2023-01-06
docs(tabs): icons are hidden in usage examples

liamdebeasi:
The icons in each tab serve as decoration as each tab button has a visible label. As a result, they should not be announced by screen readers. This PR adds aria-hidden="true" to the icons used in the usage examples.


amandaejohnston (approved):
Looks like the React usage was missed (I'm guessing you ctrl+f'd for `ion-icon` and missed the `IonIcon`s, since that's what I almost did too 😉) but once that's fixed up, this LGTM.

sean-perkins (approved):

PR #2695, opened 2023-01-10
Cleanup ionic cli commands, tables and deprecations

dtarnawsky:
- Commands that do not have options wont have examples like `ionic signup [options]`
- Option for --lab has been removed from description of `ionic serve`
- Option for --lab-port has been removed from `ionic serve`
- Tables for options and advanced options replaced with standard headers and text
- Any inputs to commands are shown as an arguments. Eg `ionic cordova build [platform]` rather than `ionic cordova build [inputs]`
- doctor commands now have a deprecation warning

### DIFF
--- a/scripts/data/cli.json
+++ b/scripts/data/cli.json
@@ -1920,10 +1920,10 @@
     {
       "name": "ionic doctor check",
       "namespace": ["ionic", "doctor"],
-      "summary": "\n\n:::warning\nThis command has been deprecated and will be removed in an upcoming major release of the Ionic CLI.\n:::\nCheck the health of your Ionic project",
+      "summary": "Check the health of your Ionic project",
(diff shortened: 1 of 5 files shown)


scripts/data/cli.json
```diff
@@ -1920,10 +1920,10 @@
     {
       "name": "ionic doctor check",
       "namespace": ["ionic", "doctor"],
-      "summary": "\n\n:::warning\nThis command has been deprecated and will be removed in an upcoming major release of the Ionic CLI.\n:::\nCheck the health of your Ionic project",
+      "summary": "Check the health of your Ionic project",
```
dotNetkow, line 1923: opportunity to mention the VS code extension?

dtarnawsky: I've got a bit of work to do to understand the checks that `doctor` does and make sure that the VS Code extension covers those cases. When I get around to that I think it would be relevant to update the docs to say that it supersedes the `doctor` command.

PR #2699, opened 2023-01-10
docs(action-sheet): add inline usage 

amandaejohnston:
- Adds playgrounds for inline usage, including through `trigger` and `isOpen`.
- Converts existing `css-properties` and `styling` playgrounds to use inline syntax.
- Renames `basic` playground to `controller`.
- Adds new `role-info-on-dismiss` playground, which shows how to capture the `data` and `role` of the dismiss event. Originally this was part of the controller playground, but it's been pulled out into its own to simplify everything.


static/usage/v7/action-sheet/theming/styling/vue.md
```diff
@@ -29,47 +29,46 @@
 </style>
 
 <template>
-  <ion-button @click="presentActionSheet">Open</ion-button>
+  <ion-button id="open-action-sheet">Open</ion-button>
+  <ion-action-sheet
+    trigger="open-action-sheet"
+    css-class="my-custom-class"
```
liamdebeasi, line 35: Since we are using the component directly, we can just specify `class`. This applies to other framework usages (though react should use `className`)
```suggestion
   class="my-custom-class"
```

static/usage/v7/action-sheet/inline/trigger/javascript.md
```diff
@@ -0,0 +1,34 @@
+```html
+<ion-button id="open-action-sheet">Open</ion-button>
+<ion-action-sheet
+  trigger="open-action-sheet"
+  header="Actions"
+></ion-action-sheet>
+
+<script>
+  const actionSheet = document.querySelector('ion-action-sheet');
```
liamdebeasi, line 9: cc @sean-perkins does this need to be `var` still or has the StackBlitz bug been fixed?

liamdebeasi (changes requested):

static/usage/v7/action-sheet/inline/trigger/javascript.md
amandaejohnston: Went ahead and changed it to `var` just in case.

liamdebeasi (approved):

PR #2700, opened 2023-01-10
fix(docs): Remove orphaned Ionic Studio page

dtarnawsky:


kensodemann (approved):
I 1000% agree that this page should not exist.

PR #2702, opened 2023-01-12
fix(playground): angular and html examples use web containers

sean-perkins:
**Problem**

Stackblitz's `html` and `angular-cli` templates use CDN links for resolving dependencies. There is a known bug where the CDN links always load the latest published version of a package, regardless of the version you specify in the `package.json`: https://twitter.com/seanfperkins/status/1575157474281283587

**Solution**

This PR migrates the HTML and Angular example base templates to use web containers, so we have absolute control over the dependency version. The HTML project templates uses a minimal express server to serve the static HTML assets. The Angular template uses the Angular CLI to serve the project.


sean-perkins:
Couple of challenges I am running into that I need to work through:
- [ ] Exposing the controllers (e.g. `modalController`) through window does not make it a global. This causes those examples to fail.
- [ ] Angular web containers will run out of memory 💀. Need to check if `--max-old-space-size=4096` can be passed to workaround this.
  - > RangeError: WebAssembly.Instance(): Out of memory: Cannot allocate Wasm memory for new instance

liamdebeasi:
I'm going to close this for now since there hasn't been any movement since January. Feel free to re-open if you'd still like to merge.

PR #2705, opened 2023-01-17
docs(upgrade-guides): add v7 migration, new upgrade guide links

liamdebeasi:
This PR does a few things:

1. Remove the "Upgrading to Ionic 6" and "Migration" links in favor of a dedicated "Upgrade Guides" section on the sidebar. This improves the visibility of the migration guides. It also scales better than the previous approach as we can add an `8-0.md` file to the `updating` directory when it comes time to write the Ionic 8 migration guide.
2. Creates the Ionic 7 migration. I only included the changes that are required in order for an app to compile. (This is what we did with the Ionic 6 migration guide too)
3. Adds a note to the v5, v6, and v7 upgrade guides reminding users to update to the latest version of the previous major release before proceeding with upgrading.

~Note: I still need to create 301 redirects so the old migration guide links do not break. This needs to happen in https://github.com/ionic-team/ionic-nginx/blob/main/config/redirects.conf.~

Added 301 redirects: https://github.com/ionic-team/ionic-nginx/pull/22

v7 guide: https://ionic-docs-git-v7-migration-ionic1.vercel.app/docs/updating/7-0

You can also access other guides in the "Upgrade Guides" sidebar.

### DIFF
--- a/docs/updating/7-0.md
+++ b/docs/updating/7-0.md
@@ -1,16 +1,18 @@
-# Upgrading to Ionic 7
+---
+title: Updating to 7.0
+---
 
-Follow this guide to upgrade your Ionic 6 apps to Ionic 7.
+# Updating from Ionic 6 to Ionic 7
 
 :::note
 Ionic 7 is in beta. Please report any issues on [the Ionic Framework GitHub Repo](https://github.com/ionic-team/ionic-framework).
 :::
 
-## Getting Started
-
 :::note
-This guide assumes that you have already updated your app to the latest version of Ionic 6. Make sure you have followed the [Upgrading to Ionic 6 Guide](./upgrading-to-ionic-6) before starting this guide.
+This guide assumes that you have already updated your app to the latest version of Ionic 6. Make sure you have followed the [Upgrading to Ionic 6 Guide](./6-0) before starting this guide.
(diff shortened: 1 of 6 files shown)


docs/updating/7-0.md
```diff
@@ -0,0 +1,122 @@
...
+
+1. Rename any usages of the `--background` and `--background-checked` CSS Variables to `--track-background` and `--track-background-checked`, respectively.
+
+### Virtual Scroll
+
+1. Remove `ion-virtual-scroll` and any associated types. This component has been removed in favor of using virtual scroll solutions provided by JavaScript Frameworks. The guides below contain more information about this migration:
+
+[Angular Migration Guide](https://ionicframework.com/docs/angular/virtual-scroll)<br />
+[React Migration Guide](https://ionicframework.com/docs/react/virtual-scroll)<br />
+[Vue Migration Guide](hhttps://ionicframework.com/docs/vue/virtual-scroll)
```
amandaejohnston, line 115:
```suggestion
[Vue Migration Guide](https://ionicframework.com/docs/vue/virtual-scroll)
```

amandaejohnston (approved):
Couple minor things, but otherwise LGTM.

PR #2710, opened 2023-01-19
docs(icon): add dedicated component page and playground

amandaejohnston:
Currently, our docs just link to https://ionic.io/ionicons for info on `ion-icon`. This PR adds a dedicated page for the component, so that we can include a playground in an easy-to-find spot showing how to import and use the icon names with each framework. It also edits the links in the sidebar and main components index page to point to this new dedicated page instead of the Ionicons site.

Some things to note:
- While this PR includes docs for both v6 and v7, it's off the v7 branch so no extra work is needed given the structural changes to the docs repo. There is a separate PR at https://github.com/ionic-team/ionic-docs/pull/2710 which backports the updates to `main`.
- I left off the autogenerated props, methods, etc. because the docs site pulls this info from a file auto-generated by Stencil over on the framework repo, and we're trying to keep the framework and ionicons repos decoupled as much as possible. I figure a link to the dedicated Ionicons site covers our bases there since it already has that info.
- I also didn't include more than a basic usage playground since all the props and such seemed self-explanatory to me, and I didn't want to duplicate too much info between this and the Ionicons site, but LMK if you'd like an additional playground on any particular feature.

### DIFF
--- a/docs/api/icon.md
+++ b/docs/api/icon.md
@@ -1,0 +1,21 @@
+---
+title: 'ion-icon'
+---
+
+<head>
+  <title>ion-icon: Icon Component for Ionic Framework Apps</title>
+  <meta
+    name="description"
+    content="Ion-icon is a component for displaying premium designed icons with support for SVG and web font."
+  />
+</head>
+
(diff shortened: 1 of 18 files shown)


liamdebeasi (approved):
This is great! Nice work 👍 

Could we backport this to the v6 docs?

amandaejohnston:
Backport PR created: https://github.com/ionic-team/ionic-docs/pull/2716

sean-perkins (approved):
Nicely done!

PR #2713, opened 2023-01-22
Add IonRouterOutlet

xclusive36:
Updated tab-button for react to add IonRouterOutlet. Otherwise documentation when applied throws an error when used in an Ionic Framework App with React.

Source: https://github.com/ionic-team/ionic-react-conference-app/blob/master/src/pages/MainTabs.tsx#L17

### DIFF
--- a/docs/api/tab-button.md
+++ b/docs/api/tab-button.md
@@ -126,30 +126,51 @@ See the [tabs documentation](tabs.md) for more details on configuring tabs.
 
 ```tsx
 import React from 'react';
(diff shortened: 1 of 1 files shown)


amandaejohnston:
Thank you for the PR! In the time since this was made, we've added a live demo that shows example usage inluding `ion-router-outlet`: https://ionicframework.com/docs/api/tabs#usage-with-router As such, I'm going to close this out, but we appreciate the work you put in.

PR #2715, opened 2023-01-23
docs(react): adding react to an existing app

### LINKED ISSUE #2714: bug: Getting Started Docs for Existing React users is wrong
### Describe the Bug

This bug report was requested by Sean Perkins, in this tweet: https://twitter.com/seanfperkins/status/1617373232176168960

The documentation does not work for existing React users that want to integrate Ionic. Users will either end up with a blank screen, or with a bunch of gray buttons for every date, year, month, strewn all over the page.

### Expected Behavior

See a date time input, or don't use that for the first example.

### Steps to Reproduce

Go here: https://ionicframework.com/docs/
Click "React" and land here: https://ionicframework.com/docs/react
Scroll down to Resources and click "Add Ionic to Existing React App":
https://dev.to/ionic/adding-ionic-react-to-an-existing-react-project-4kib

In the first section, these are the first four commands given. They don't work as stated under Describe the Bug. See Additional Informatioin below for the solution.
```
npm i @ionic/react
import '@ionic/react/css/core.css';
import { IonButton, IonDatetime } from '@ionic/react';
<IonDatetime displayFormat="MM/DD/YYYY" placeholder="Select Date"></IonDatetime>
```

### Screenshots

_No response_

### Operating System

Linux

### Browser

Chrome

### Version

109

### Additional Information

A correct example using modern ionic and React would look like the last entry shown under "working example" here:
https://github.com/ionic-team/ionic-framework/issues/24917
Specifically that page links to this code sandbox with the solution:
https://codesandbox.io/s/cool-fog-u3by0v?file=/src/App.js

If you run into a blank screen, the cause and solution is described here:
https://forum.ionicframework.com/t/ionic-4-global-scss-cause-blank-screen-only-to-browser/169677/3

sean-perkins:
The Ionic React resources is linking out to an outdated blog post for adding `@ionic/react` to an existing React application. This PR migrates that content into the Ionic Docs and updates the content to reflect the required changes as of Ionic 6. 

Resolves #2714

### DIFF
--- a/docs/react.md
+++ b/docs/react.md
@@ -128,7 +128,7 @@ $ ionic serve █
 
 <DocsCard
   header="Add Ionic to Existing React App"
-  href="https://dev.to/ionic/adding-ionic-react-to-an-existing-react-project-4kib"
+  href="/docs/react/adding-ionic-react-to-an-existing-react-project"
   icon="/icons/logo-react-icon.png"
 >
   <p>Use individual components or the complete app experience.</p>
--- a/docs/react/adding-ionic-react-to-an-existing-react-project.md
+++ b/docs/react/adding-ionic-react-to-an-existing-react-project.md
@@ -12,13 +12,35 @@ To get started with using components install the `@ionic/react` package:
 npm i @ionic/react
 ```
 
-Next, import the stylesheets from Ionic in your main app file:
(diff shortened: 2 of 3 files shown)


liamdebeasi (approved):
One changes otherwise looks good

Issue #2580, opened 2022-09-27
content: add framework documentation for getting Ionic mode


brandyscarney:
### URL

https://ionicframework.com/docs

### Issue Description

I can't find anywhere we document how to get the mode, for example in JavaScript:

```typescript
const mode = Ionic.mode;
```

in Angular:

```typescript
import { Config } from '@ionic/angular';

constructor(
  public config: Config
) { }
  
ngOnInit() {
  this.mode = this.config.get('mode');
}
```

PR #2727, opened 2023-01-26
docs(config): move framework config guides to unified config guide

liamdebeasi:
The config guides on the docs are currently split out into Angular, React, and Vue guides. However, they contain a lot of the same information. The main thing that differs is the code snippets.

I would like to create a consolidated config page that has per-framework code snippets that is located in the "Developing" section of the docs. This has a few benefits:

1. Makes the config options in Ionic more visible. Currently they are buried in each JS Framework section.
2. Makes it possible to link to a single config doc. Several pages on the docs link to the "Ionic config" docs but then only link to the Angular-specific config page.
3. Reduces the amount of content we need to maintain. Currently, we need to make any config copy updates 3 times (Angular, React, and Vue)

Nginx redirects PR: https://github.com/ionic-team/ionic-nginx/pull/23

Preview: https://ionic-docs-git-config-consolidate-ionic1.vercel.app/docs/v7/developing/config


docs/developing/config.md
```diff
@@ -202,19 +101,6 @@ class AppComponent {
...
-import { Config } from '@ionic/angular';
-
-@Component(...)
-class AppComponent {
-  constructor(config: Config) {
-    const keyboardHeight = config.getNumber('keyboardHeight');
-  }
-}
-```
-
```
amandaejohnston, line 217: Was this example removed intentionally? I know the info is fairly redundant, but it seems strange to have an example on the other two but not this one. Maybe one example for the whole provider would be easier?

docs/developing/config.md
```diff
@@ -2,158 +2,57 @@
 title: Config
```
amandaejohnston, line 2: Could we also backport the change to the v6 docs?

amandaejohnston (changes requested):

liamdebeasi:
This isn't as needed now that base components are on pause.

amandaejohnston:
I would argue it's still useful 👀 The reasons you outlined in the description aren't at all base components-specific.

liamdebeasi:
I agree it's still useful, but I don't have time to make the changes this sprint 😄 . It's not as pressing ATM, so I'll try to get this pulled into the next sprint

docs/developing/config.md
liamdebeasi: This was moved to `config#getnumber`. React and Vue currently do not provide this kind of API which is why I made it Angular-only.

docs/developing/config.md
liamdebeasi: Done in https://github.com/ionic-team/ionic-docs/pull/2727/commits/5ca3b5075b6de157f6e1b49c72c1249b8d069f28

brandyscarney:
This also looks like it could close https://github.com/ionic-team/ionic-docs/issues/2580 if we added in the other frameworks under the "Reading the config" section.

amandaejohnston (approved):

liamdebeasi:
@brandyscarney We don't have mechanisms for reading the config in Ionic React and Ionic Vue which is why those docs aren't available.

PR #2728, opened 2023-01-26
chore(deps): update playgrounds to pull from next tag

liamdebeasi:


static/code/stackblitz/v7/react/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@ionic/react": "6.4.2-dev.11673981545.1e612253",
+    "@ionic/react": "next",
```
amandaejohnston, line 6: It seems like Stackblitz might have an issue with installing from the `next` tag? If I try the inline action sheet example (https://ionic-docs-git-playground-next-tag-ionic1.vercel.app/docs/v7/api/action-sheet#inline-action-sheets-recommended), Angular and JS behave as expected (i.e. they don't work due to that Stackblitz issue) but Vue opens the action sheet automatically and it can't be presented again, and React errors out entirely. This doesn't happen in the live example (https://ionicframework.com/docs/v7/api/action-sheet#inline-action-sheets-recommended) and if I change the `next` tag to `7.0.0-beta.0` manually in the Stackblitz, things start working again.

sean-perkins: Yeah looks like we are missing the package-lock updates as well here.

liamdebeasi: Ok I updated the StackBlitz deps to use the specific beta: https://github.com/ionic-team/ionic-docs/pull/2728/commits/b987cd41281f0b4ed1bc29aa6b082021addc20f2

The playgrounds iframes will still pull from next, but we'll have to make sure we update these deps on each beta.

amandaejohnston (approved):

sean-perkins (approved):

PR #2729, opened 2023-01-26
docs(tabs): component playground example

sean-perkins:
Adds a component playground example for `ion-tabs` for the v6 and v7 docs.


docs/api/tabs.md
```diff
@@ -42,448 +26,34 @@ Both `ion-tabs` and `ion-tab-bar` can be used as standalone elements. They don
...
-interface TabsCustomEvent extends CustomEvent {
-  detail: { tab: string };
-  target: HTMLIonTabsElement;
-}
-```
+Tabs can be used with the Ionic router to implement tab-based navigation. The tab bar and active tab will automatically resolve based on the url. This is the most common pattern for tabs navigation.
 
+import Router from '@site/static/usage/v7/tabs/router/index.md';
 
+<Router />
```
amandaejohnston, line 35: Would it be possible to move the `.example-content` CSS to a global file in the Angular code, or maybe remove it entirely (across all frameworks for consistency)? While I do think it looks nice to have the content centered like that, the Angular example's number of files is a bit overwhelming at the moment.

docs/api/tabs.md
```diff
@@ -42,448 +26,34 @@ Both `ion-tabs` and `ion-tab-bar` can be used as standalone elements. They don
...
-While not required, this interface can be used in place of the `CustomEvent` interface for stronger typing with Ionic events emitted from this component.
+## Usage with Router
 
-```typescript
-interface TabsCustomEvent extends CustomEvent {
-  detail: { tab: string };
-  target: HTMLIonTabsElement;
-}
-```
+Tabs can be used with the Ionic router to implement tab-based navigation. The tab bar and active tab will automatically resolve based on the url. This is the most common pattern for tabs navigation.
```
amandaejohnston, line 31: Could we mention that using either `ion-router-outlet` or `ion-nav` is required for `ion-tabs` to work?

docs/api/tabs.md
sean-perkins: I updated the Angular examples. Stackblitz is being unreliable today when launching new examples..

amandaejohnston (approved):

PR #2731, opened 2023-01-27
update type to fix build

dereklowlind:
update the UserPhoto interface as specific string type breaks the build in this step


sean-perkins:
@dereklowlind can you elaborate on this change? 

What are you assigning to `webviewPath` that isn't a `string` type?

liamdebeasi:
Hey there,

We have not heard back regarding our question in https://github.com/ionic-team/ionic-docs/pull/2731#issuecomment-1407074994. We are going to close the PR for now. If you have clarification for our question feel free to reply, and we will be happy to re-open the PR. Let me know if you have questions. Thanks!

PR #2733, opened 2023-01-28
docs: version ionic cli and native as v6

dtarnawsky:


mhartington:
My docusaurus knowledge is lacking a bit. Will this also handle the sidebar conten as well?

dtarnawsky:
Closing as Lars has generated the CLI and Native docs sections for v6/c7(beta) rather than version them.

PR #2738, opened 2023-01-31
Add missing import of IonBackButton in Vue example

jochensengier:


liamdebeasi:
Thanks for the PR! We recently released new `ion-back-button` documentation that makes use of our component playgrounds: https://ionicframework.com/docs/api/back-button

As part of this work we removed the old usage code snippets. The new component playground code correctly imports `IonBackButton` from `@ionic/vue` and provides it to the Vue component. I am going to close this, but let me know if you have any questions.

PR #2756, opened 2023-02-03
Update range.md

WessimH:
The expected type of Ionchange is not specified


liamdebeasi:
Thanks for the PR! Could you please clarify what issue this resolves? The interface for events emitted from `ion-range` is noted here: https://ionicframework.com/docs/api/range#rangecustomevent

liamdebeasi:
Hey there,

I have not heard back in a while so I am going to close. Let me know if you'd like to reopen and can provide more context on https://github.com/ionic-team/ionic-docs/pull/2756#issuecomment-1431653158.

PR #2758, opened 2023-02-05
update swiper/ionic7 Slides migration guide

mia-z:
If using Vite or similar build tools (that aren't Create-React-App) then the given import path is unclear and not documented anywhere that I could see so I added those import variations under the Pre-Processors section


liamdebeasi:
Thanks for the PR! It looks like this change was recently made: https://ionicframework.com/docs/react/slides#swiping-with-style

I am going to close this, but let me know if you have any questions.

PR #2762, opened 2023-02-06
docs(card): remove extra ios header styling from v7 playgrounds

amandaejohnston:
In Ionic v7, cards in `ios` mode will automatically display the title and subtitle in reverse order to match native behavior (see: https://github.com/ionic-team/ionic-framework/pull/26084). Our playgrounds currently have custom CSS to achieve this look, but this is no longer needed now that the fix is merged. (Note that this only applies to the v7 docs, as the fix was not included in v6.) This PR removes the extra CSS, and also updates all card playgrounds to use the latest beta version of Framework to include the fix.


sean-perkins (approved):
Nice work!

PR #2763, opened 2023-02-07
fix(datetime): align basic demo container to the start

sean-perkins:
When viewing the [basic usage](https://ionicframework.com/docs/api/datetime#basic-usage) example for `ion-datetime` in MD mode, toggling the month/year picker results in a UI shift. 

This isn't a behavior of Ionic, but the result of the container styles in the demo. 

Aligning the container to the start removes this layout shift.

Before:

<video src="https://user-images.githubusercontent.com/13732623/217121595-2ec22191-b2a2-4053-8c24-fb604dcd719d.mp4"></video>

After:

<video src="https://user-images.githubusercontent.com/13732623/217121683-c9738564-e661-4c56-a6b3-a3266224ebad.mp4"></video>

### DIFF
--- a/static/usage/v6/datetime/basic/demo.html
+++ b/static/usage/v6/datetime/basic/demo.html
@@ -13,6 +13,10 @@
     ion-datetime {
       width: 350px;
     }
+
+    .container {
+      align-items: flex-start;
+    }
   </style>
 </head>
 
--- a/static/usage/v7/datetime/basic/demo.html
+++ b/static/usage/v7/datetime/basic/demo.html
@@ -13,6 +13,10 @@
     ion-datetime {
       width: 350px;
     }
+
+    .container {
+      align-items: flex-start;
+    }
   </style>
 </head>
 


liamdebeasi (commented):
This is still happening on MD mode (may need to switch to dark mode for the difference to be noticeable)

amandaejohnston (approved):
I'm not seeing the shift on my end 🤷 Recording in dark mode: https://user-images.githubusercontent.com/90629384/217342385-a2cfcb1b-cddd-49ab-ba83-4f147a4eeeb3.mp4

liamdebeasi:
The datetime container resize in the video. Isn't that what this PR fixes? It was happening in iOS mode prior to the patch.

sean-perkins:
@liamdebeasi this doesn't adjust the container size. It's preventing the active click target from visually shifting. You can see in the first example, I click, but then the button moves down on the screen and I have to move the mouse to toggle it. This is a side effect of the demo styles we use to "visually center" the component in the playground viewport.

liamdebeasi:
Oh, that makes sense. Thanks!

liamdebeasi (approved):

PR #2769, opened 2023-02-13
docs(datetime): add highlightedDates playgrounds

amandaejohnston:
Adds playgrounds for the new `highlightedDates` property added here: https://github.com/ionic-team/ionic-framework/pull/26775 Demos for using the prop as both an array and a callback are included, as well as versions for both v6 and v7.

Dev build for feature: `6.5.3-dev.11676573595.1564fefe`

### DIFF
--- a/docs/api/datetime.md
+++ b/docs/api/datetime.md
@@ -290,6 +290,8 @@ Developers can provide their own buttons for advanced custom behavior.
 
 Using the `highlightedDates` property, developers can style particular dates with custom text or background colors. This property can be defined as either an array of dates and their colors, or a callback that receives an ISO string and returns the colors to use.
 
+To maintain a consistent user experience, the style of selected date(s) will always override custom highlights.
+
 :::note
 This property is only supported when `preferWheel="false"`, and using a `presentation` of either `"date"`, `"date-time"`, or `"time-date"`.
 :::
--- a/docs/api/datetime.md
+++ b/docs/api/datetime.md
@@ -298,9 +298,13 @@ This property is only supported when `preferWheel="false"`, and using a `present
 
 ### Using Array
 
+An array is better when the highlights apply to fixed dates, such as due dates.
+
 <HighlightedDatesArray />
 
 ### Using Callback
+
+A callback is better when the highlighted dates are recurring, such as birthdays or recurring meetings.
 
 <HighlightedDatesCallback />
 
--- a/docs/api/datetime.md
+++ b/docs/api/datetime.md
@@ -290,6 +290,8 @@ Developers can provide their own buttons for advanced custom behavior.
 
 Using the `highlightedDates` property, developers can style particular dates with custom text or background colors. This property can be defined as either an array of dates and their colors, or a callback that receives an ISO string and returns the colors to use.
 
+When specifying colors, any valid CSS color format can be used. This includes hex codes, rgba, [color variables](../theming/colors), etc.
+
 To maintain a consistent user experience, the style of selected date(s) will always override custom highlights.
 
 :::note
--- a/docs/api/datetime.md
+++ b/docs/api/datetime.md
@@ -31,6 +31,9 @@ import ShowingConfirmationButtons from '@site/static/usage/v7/datetime/buttons/s
 import CustomizingButtons from '@site/static/usage/v7/datetime/buttons/customizing-buttons/index.md';
 import CustomizingButtonTexts from '@site/static/usage/v7/datetime/buttons/customizing-button-texts/index.md';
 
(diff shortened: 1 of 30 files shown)


liamdebeasi (commented):
The demo in MD mode does not work for some reason. Also every time I click it switches months. Interestingly, when I reset the demo it works:

https://user-images.githubusercontent.com/2721089/219681495-8e6881d9-df5a-4a73-9e60-eb0ae13ff470.mov

docs/api/datetime.md
```diff
@@ -283,6 +286,22 @@ Developers can provide their own buttons for advanced custom behavior.
 
 <CustomizingButtons />
 
+## Highlighting Specific Dates
+
+Using the `highlightedDates` property, developers can style particular dates with custom text or background colors. This property can be defined as either an array of dates and their colors, or a callback that receives an ISO string and returns the colors to use.
```
liamdebeasi, line 291: Can we call out that the default selected date style will always override custom styles? Would be good to add some reasoning around that decision too.

docs/api/datetime.md
```diff
@@ -283,6 +286,22 @@ Developers can provide their own buttons for advanced custom behavior.
 
 <CustomizingButtons />
 
+## Highlighting Specific Dates
```
liamdebeasi, line 289: Can we add some verbiage around when you might want to use an array versus a callback? (i.e. array is good when you have fixed dates, callback is good when you have repeating dates)

static/usage/v6/datetime/highlightedDates/array/angular/example_component_ts.md
```diff
@@ -0,0 +1,32 @@
+```ts
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'app-example',
+  templateUrl: 'example.component.html',
+})
+export class ExampleComponent {
+  public highlightedDates = [
```
liamdebeasi, line 9: Angular style guides say not to add the explicit `public` 
```suggestion
  highlightedDates = [
```

static/usage/v6/datetime/highlightedDates/array/angular/example_component_ts.md
```diff
@@ -0,0 +1,32 @@
...
+    {
+      date: '2023-01-05',
+      textColor: '#800080',
+      backgroundColor: '#ffc0cb',
+    },
+    {
+      date: '2023-01-10',
+      textColor: '#09721b',
+      backgroundColor: '#c8e5d0',
```
liamdebeasi, line 18: Can we have some of these samples use CSS variables to indicate that you don't always have to use hex values?

docs/api/datetime.md
liamdebeasi, line 291: Can we call out that you can place any valid CSS color value in here? (i.e. rgba, hsl, variables with colors, etc)

static/usage/v7/datetime/highlightedDates/array/javascript.md
```diff
@@ -0,0 +1,29 @@
+```html
+<ion-datetime presentation="date" value="2023-01-01"></ion-datetime>
+
+<script>
+  const datetime = document.querySelector('ion-datetime');
```
liamdebeasi, line 5:
```suggestion
  var datetime = document.querySelector('ion-datetime');
```

Const does not work because of a StackBlitz issue (same change needed for the other JS samples that run in StackBlitz as well)

liamdebeasi (changes requested):

amandaejohnston:
> The demo in MD mode does not work for some reason. Also every time I click it switches months.

Looks like this is a separate bug; I'm able to replicate it in the live v6 and v7 docs too. (It might take a few refreshes.) I'm unsure whether the resolution will be on the framework or the playground component itself, but it's most obviously manifesting in the docs, so I filed an issue there: https://github.com/ionic-team/ionic-docs/issues/2778

static/usage/v7/datetime/highlightedDates/array/index.md
```diff
@@ -0,0 +1,24 @@
+import Playground from '@site/src/components/global/Playground';
+
+import javascript from './javascript.md';
+import react from './react.md';
+import vue from './vue.md';
+
+import angular_example_component_html from './angular/example_component_html.md';
+import angular_example_component_ts from './angular/example_component_ts.md';
+
+<Playground
```
sean-perkins, line 10: The playground component currently defaults to version `6`. We will need to specify the explicit version on playground examples:

```tsx
<Playground version={7}
```

v6 playground examples will also need to be:
```tsx
<Playground version={6}
```

This flag is used to determine which stackblitz template files are used.

liamdebeasi (approved):
Great job!

sean-perkins (approved):
🎉

PR #2791, opened 2023-02-24
chore: version api reference docs

sean-perkins:
Updates the API script to version component API reference information into versioned subdirectories, so that the main markdown source files can import the correct version.

e.g.: v6 docs for components now reference v6 APIs and v7 docs for components now reference the `next` tag (or v7) APIs.

Issue: FW-3411


scripts/api.js
```diff
@@ -18,26 +33,36 @@ const fs = require('fs');
...
-  components.forEach((comp) => {
-    const compTag = comp.tag.slice(4);
-    writeAutoGeneratedPage(compTag, 'props', renderProperties(comp));
-    writeAutoGeneratedPage(compTag, 'events', renderEvents(comp));
-    writeAutoGeneratedPage(compTag, 'methods', renderMethods(comp));
-    writeAutoGeneratedPage(compTag, 'parts', renderParts(comp));
-    writeAutoGeneratedPage(compTag, 'custom-props', renderCustomProps(comp));
-    writeAutoGeneratedPage(compTag, 'slots', renderSlots(comp));
-  });
+  for (let { npmTag, version } of versions) {
```
liamdebeasi, line 36: I believe this file runs for each docusaurus version. Does this mean that it's going to re-generate v6 APIs twice and then v7 APIs twice?

liamdebeasi: I was hoping there was a way to pass the version in the command line so that it naturally works with docusaurus versions

sean-perkins: This script runs once, before the docusaurus build starts. The docusaurus build process includes all versions. For example when you run `docusaurus build`, it is statically building the markdown contents for the latest version and all legacy versions to HTML (this is why the build for static sites takes longer and longer with each version you add). 

If we are wanting these files to be co-located within the `versioned_docs` folder, I can make that change 👍 it will just write to the versioned folder directory and use a relative path for importing, instead of using the alias path.

PR #2804, opened 2023-03-02
pin the versions for angular to v5

skofgar:
setting `@ionic/angular@5` and `@ionic/angular-toolkit@5` is important, otherwise this will skip the migration to v5 and go straight to v6 or any other, newer packages.

### DIFF
--- a/versioned_docs/version-v6/reference/migration.md
+++ b/versioned_docs/version-v6/reference/migration.md
@@ -31,7 +31,7 @@ For a complete list of breaking changes from 4.x to 5.x, please refer to [the br
 For Angular based projects, you can simply run:
 
 ```shell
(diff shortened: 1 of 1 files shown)


amandaejohnston:
Thank you for the PR! There are a few other places where the `npm install` commands needed updating, but I noticed that the migration guides hosted on the v7 version of the docs have properly pinned versions ([example](https://ionicframework.com/docs/updating/5-0)). So, I created https://github.com/ionic-team/ionic-docs/pull/2871 to edit the versioned pages to link to these latest versions instead, side-stepping the issue and narrowing things to one source of truth.

I'm going to close this PR in lieu of the aforementioned one, but I've given you co-author credit when merging. Thanks again!

PR #2808, opened 2023-03-06
docs(layout): update css-utilities to float properly

### LINKED ISSUE #265: CSS utilities, float elements code snippet doesn't actually leverage float classes for layout due to grid
**I'm submitting a ...**
[x] bug report
[ ] feature request

**URL:**
https://beta.ionicframework.com/docs/layout/css-utilities#element-placement

**Current behavior:**
The markup does not actually leverage `float-*` attributes since it uses a grid.

**Expected behavior:**
A relevant markup, without a grid.

**Steps to reproduce:**
- Copy sample markup,
- Change with responsive float attributes that should reverse the float positions, for instance:
```html
<ion-grid>
  <ion-row>
    <ion-col float-left float-md-right>
      <div>
        <h3>float-left (should be float-right above md breakpoint)</h3>
        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed ac vehicula lorem.
      </div>
    </ion-col>

    <ion-col float-right float-md-left>
      <div>
        <h3>float-right (should be float-left above md breakpoint)</h3>
        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed ac vehicula lorem.
      </div>
    </ion-col>
  </ion-row>
</ion-grid>
```
- Notice that it doesn't work as advertised,
- Debug and see that Flexbox is handling the actual layout.

**Other information:**
Can be fixed with the following markup and CSS that both mimic the grid structure:

_Template_
```html
<div class="ion-grid">
  <div class="ion-row">
    <div class="ion-col ion-col-33" float-left float-md-right>
      <div>
        <h3>float-left (should be float-right above md breakpoint)</h3>
        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed ac vehicula lorem.
      </div>
    </div>

    <div class="ion-col ion-col-33" float-right float-md-left>
      <div>
        <h3>float-right (should be float-left above md breakpoint)</h3>
        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed ac vehicula lorem.
      </div>
    </div>
  </div>
</div>
```

_CSS_
```css
.ion-col {
  padding: 5px;
}

.ion-col-33 {
  width: 33%;
}

.ion-grid {
  padding: 0 0 20px 0;
}

.ion-col div {
  background-color: #f5f7fa;
  border: 1px solid #e4e8f0;
  padding: 5px;
}

.ion-row:after {
  content: "";
  clear: both;
  display: block;
}
```

brandyscarney:
I kept the grid because this saves us from having to clear the floats between examples, but I moved the float css to an image so you can properly see it 

Code preview: https://1vutb4.stackblitz.io

Site preview: https://ionic-docs-git-fw-3055-ionic1.vercel.app/docs/v7/layout/css-utilities#float-elements

closes #265

### DIFF
--- a/docs/layout/css-utilities.md
+++ b/docs/layout/css-utilities.md
@@ -134,17 +134,37 @@ The float CSS property specifies that an element should be placed along the left
 ```html
 <ion-grid>
   <ion-row>
-    <ion-col class="ion-float-left">
-      <div>
-        <h3>float-left</h3>
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed ac vehicula lorem.
-      </div>
+    <ion-col>
+      <h3>no float</h3>
+      <img
+        alt="Silhouette of a person's head"
+        src="https://ionicframework.com/docs/img/demos/avatar.svg"
+        height="50px"
+      />
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed ac
+      vehicula lorem.
     </ion-col>
-    <ion-col class="ion-float-right">
-      <div>
-        <h3>float-right</h3>
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed ac vehicula lorem.
-      </div>
+    <ion-col>
+      <h3>float-left</h3>
+      <img
+        alt="Silhouette of a person's head"
+        src="https://ionicframework.com/docs/img/demos/avatar.svg"
+        height="50px"
+        class="ion-float-left"
+      />
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed ac
+      vehicula lorem.
+    </ion-col>
+    <ion-col>
+      <h3>float-right</h3>
+      <img
+        alt="Silhouette of a person's head"
+        src="https://ionicframework.com/docs/img/demos/avatar.svg"
+        height="50px"
+        class="ion-float-right"
+      />
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed ac
+      vehicula lorem.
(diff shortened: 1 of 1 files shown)


docs/layout/css-utilities.md
```diff
@@ -134,17 +134,37 @@ The float CSS property specifies that an element should be placed along the left
...
-    <ion-col class="ion-float-left">
-      <div>
-        <h3>float-left</h3>
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed ac vehicula lorem.
-      </div>
+    <ion-col>
+      <h3>no float</h3>
+      <img
+        alt="Silhouette of a person's head"
+        src="https://ionicframework.com/docs/img/demos/avatar.svg"
```
liamdebeasi, line 141: Can we use the local asset?

```jsx
import useBaseUrl from '@docusaurus/useBaseUrl';

...

<img src={useBaseUrl('img/demos/avatar.svg')} />
```

That way when developing locally you get the local SVG instead of the one that is in production (in case the SVG ever changes)

Same thing for the other images.

liamdebeasi (changes requested):

docs/layout/css-utilities.md
brandyscarney: How would that work inside of a code block? All of the current uses like that are for actual rendered videos or images.

liamdebeasi: 🤦 Sorry I thought this was a playground by mistake. What you have now is correct

liamdebeasi (approved):

Issue #2778, opened 2023-02-17
bug: datetime sometimes doesn't init properly after switching to MD mode


averyjohnston:
### Describe the Bug

https://ionicframework.com/docs/api/datetime#basic-usage

Sometimes, when switching any datetime playground to `md` mode, the datetime doesn't appear to be initializing properly. The "today" date isn't highlighted, and clicking any date will cause the datetime to jump back to the previous month. Fiddling with the next/prev months buttons will fix it, as will clicking the "Reset demo" button to reload the iframe.

If the bug doesn't trigger at first, you have to refresh the page to try again. It's fairly consistent on my end though, and happens in both v6 and v7.

### Expected Behavior

The datetime should work properly in either mode of the playground.

### Steps to Reproduce

1. Scroll to any datetime playground.
2. Click the "MD" button.
3. If bug doesn't appear, refresh the entire page and try again. Seems to take 1-3 refreshes max.

### Screenshots

https://user-images.githubusercontent.com/90629384/219694426-6507a918-e8a4-48ac-b627-395953e1e819.mp4

### Operating System

Windows

### Browser

Chrome

### Version

110.0.5481.77 (Official Build) (64-bit)

PR #2809, opened 2023-03-06
fix(playground): iframe is hidden using display

liamdebeasi:
resolves https://github.com/ionic-team/ionic-docs/issues/2778

This issue was introduced in https://github.com/ionic-team/ionic-docs/pull/2623 when switching from `display: none` to visibility/width. The IntersectionObserver in `ion-datetime` does not fire when the width/visibility change. As a result, the datetime was attempting to center its contents with the width was `0px`. When the width was reset when the iframe shows, datetime does not adjust again.

The visibility/width code was added because the fullscreen offset variables were not being set correctly on `ion-content` with `display: none`. However, I fixed an issue in https://github.com/ionic-team/ionic-framework/pull/26782 which causes the content to recalculate the offset variables when the `resize` event is dispatched on the window. When the iframe is un-hidden, the resize event is dispatched so we are able to safely revert to using `display: none` while ensuring that the offset variables are calculated correctly.

https://ionic-docs-git-fw-3500-ionic1.vercel.app/docs/api/datetime

### DIFF
--- a/src/components/global/Playground/playground.css
+++ b/src/components/global/Playground/playground.css
@@ -204,8 +204,7 @@
 }
 
 .playground .frame-hidden {
-  visibility: hidden;
-  width: 0%;
+  display: none;
 }
 
 @media only screen and (max-width: 600px) {


brandyscarney (approved):
I couldn't reproduce the datetime issue that this fixes, but I jumped on your branch (to enable MD mode on the footer playground) and tested out the previous issue we had that PR https://github.com/ionic-team/ionic-docs/pull/2623 solved and the issue didn't come back, so this looks good!

amandaejohnston (approved):
LGTM!

PR #2817, opened 2023-03-09
docs(gestures): remove browser support section

sean-perkins:
Removes the browser support section from the v6 & v7 docs for the "Gestures" page.


liamdebeasi (approved):
Will removing the note from https://ionicframework.com/docs/reference/browser-support#desktop-browsers be done in a separate PR?

sean-perkins:
Good call out, I missed that. I'll remove that in this PR 👍

liamdebeasi:
Did we want to remove this for the v5 docs too?

PR #2827, opened 2023-03-14
docs(input, textarea): update migration guide for proper metadata usage

liamdebeasi:
As a result of the team's recent discussion on item/list I have updated the migration guide to note that metadata on inputs should be used outside of lists. This will require a more in-depth usage guide for items/lists, but this at least removes the incorrect usage.

### DIFF
--- a/static/usage/v7/input/migration/index.md
+++ b/static/usage/v7/input/migration/index.md
@@ -53,15 +53,21 @@ import TabItem from '@theme/TabItem';
 </ion-item>
 
 <!-- After -->
-<ion-item>
-  <ion-input
-    label="Email:"
-    counter="true"
-    maxlength="100"
-    helper-text="Enter an email"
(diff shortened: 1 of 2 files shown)


brandyscarney (commented):
I know this isn't the point of this PR, but the lack of visual breaks between before & after examples is kind of confusing: https://ionic-docs-git-updateusage-input-ionic1.vercel.app/docs/v7/api/input#using-the-modern-syntax

Could we do one code block with a before of all 3 examples, then the after in a different code block?

liamdebeasi:
I think that's going to make it hard to compare. I could break each before/after into its own code block?

liamdebeasi:
Sure thing. I can handle that in a separate PR -- would like to get this changes out first.

amandaejohnston (approved):

PR #2830, opened 2023-03-14
feat: docusaurus plugin to generate component api docs (jp)

sean-perkins:
Duplicates: https://github.com/ionic-team/ionic-docs/pull/2829 but targets the JP branch.


liamdebeasi:
Closing as we are looking into alternative infrastructure for host translations of the Ionic docs.

PR #2831, opened 2023-03-15
Set webviewPath property to be nullable

wonix:
Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.ts(2322)
photo.service.ts(30, 3): The expected type comes from property 'webviewPath' which is declared


sean-perkins:
Closing as I have not received a response. If this issue still exists, please open a new issue. Thanks!

PR #2832, opened 2023-03-15
Update modal-example_component_ts.md

BlackHakwDC:
fix failure in type-ahead::search


liamdebeasi:
Thanks for the PR! Can you please describe the issue that this resolves?

BlackHakwDC:
The search has 2 errors.

1) It searches in the value properties of the displayed objects. However, the user does not see the values but the label properties. This means that the search result is not comprehensible for him.

2) If you normalise the query, which is correct, you must also normalise the values to be searched, otherwise you will not find any results.

liamdebeasi:
Makes sense, thanks! This code is used in several other places since we use this for other JS frameworks too. I'll work on making these other changes.

PR #2836, opened 2023-03-20
chore: remove v7 beta banner

sean-perkins:
Removes the v7 beta banner as seen here: https://ionicframework.com/docs/v7/updating/7-0


liamdebeasi (approved):
Looks good. Note that this should not be merged until Ionic 7 GA is available.

PR #2841, opened 2023-03-21
Pwa react

liamdebeasi:


sean-perkins (commented):
It isn't in the diff code, but the React example for CRA is using React 17 usage. I don't know if we want to update that to React 18, since our starters and examples will also be on that:

```tsx
import React from 'react';
import ReactDOM from 'react-dom/client';
import App from './App';
import * as serviceWorker from './serviceWorker';

ReactDOM.createRoot(document.getElementById('root') as HTMLElement).render(
  <React.StrictMode>
    <App />
  </React.StrictMode>
);

// If you want your app to work offline and load faster, you can change
// unregister() to register() below. Note this comes with some pitfalls.
// Learn more about service workers: https://bit.ly/CRA-PWA
// serviceWorker.unregister();
serviceWorker.register();
```

sean-perkins (approved):
Changes look good 👍 React 17/18 comment is optional.

liamdebeasi:
Good thought. I updated this in https://github.com/ionic-team/ionic-docs/pull/2841/commits/b79fc3ad19a5c7562f733bfe12909a4222cc3fb3

PR #2842, opened 2023-03-21
docs(updating): clean up verbiage on updating documents

brandyscarney:
This PR updates the following:

- Removes the second "Ionic" in the upgrade guide titles to avoid wrapping title (`Updating from Ionic 5 to Ionic 6` -> `Updating from Ionic 5 to 6`)
- References in the sidebar change from 4.0 to v4 (`Ionic 4.0` -> `Ionic v4`)
- Breaking Changes callout 
    - Added to the top of every page (already existed on v4 and v5 pages, exists at the bottom for v6 and v7)
    - Becomes a "caution" block instead of a "note" which highlights it red, this could be changed to info instead which would highlight it blue, but I thought it should be highlighted differently
    - Updates broken links to the breaking changes documents
- Renames all mentions of versions as `4.0` to `4`, `3.0` to `3`, etc. 
- Change links to other guides from `Upgrades to` to `Updates to` for consistency with their titles (see the Note block at the top of each page) 
- Removes usage of 4.x, 5.x, etc. and instead just referred to it by major number. This was used very inconsistently, for example one sentence was: `Once the team is comfortable that the Ionic 4.0 app has become stable and has fulfilled a core set of features, you can then shut down the Ionic 1 app.`

Breaking change `caution` vs `info` block:

| caution | info |
| --| --|
| Screenshot 2023-03-21 at 3 24 37 PM | Screenshot 2023-03-21 at 3 24 24 PM |

My goal was to bring some consistency to the update guides, but I am open to feedback if anyone doesn't like any of these changes.

Preview link: https://ionic-docs-git-docs-upgrading-guides-ionic1.vercel.app/docs/v7/

### DIFF
--- a/docs/updating/4-0.md
+++ b/docs/updating/4-0.md
@@ -10,7 +10,7 @@ title: Updating to v4
 This guide assumes that you have already updated your app to the latest version of Ionic 3. If you are using Ionic 1 or 2, Make sure to follow the [Updating from Ionic 1 to 4 Guide](#updating-from-ionic-1-to-4) instead.
 :::
 
-:::caution Breaking Changes
+:::info Breaking Changes
 For a **complete list of breaking changes** from Ionic 3 to Ionic 4, please refer to [the breaking changes document](https://github.com/ionic-team/ionic-framework/blob/main/BREAKING_ARCHIVE/v4.md) in the Ionic Framework repository.
 :::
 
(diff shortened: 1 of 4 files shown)


sean-perkins (approved):
Looks good, no notes 👍

liamdebeasi (approved):
Nitpick: Could we use `info` instead of `caution`? The blue looks nicer whereas the red stole my attention away from the migration guide 😂 
Content looks good though

PR #2845, opened 2023-03-22
docs(config): move framework config guides to unified config guide

liamdebeasi:
This PR ports https://github.com/ionic-team/ionic-docs/commit/d28c8314438091c7f8137a223fc9d8671766ebdd to the Japanese documentation.


liamdebeasi:
Turns out this is not needed as the `main` docs are able to fold in the translations.

PR #2847, opened 2023-03-22
docs(v5): change CDN links to use core@v5

### LINKED ISSUE #2126: v5 sidebar demo shows v6 component
**URL**
[The URL at which the content is missing or inaccurate](https://ionicframework.com/docs/api/datetime)

Navigate to the page and select framework version v5 in the top right.

The component on the right side of the page shows the new calendar datetime component introduced in framework version 6.

thetaPC:
resolves #2126 

v5 demos were using the latest (v6) core version. This caused some components to render wrong (`ion-datetime`).

`ion-datetime` also had to be reverted to use v5 code structure.


sean-perkins (changes requested):
Tested through the migrated examples and they load and behave correctly 👍 

The `ion-accordion` and `ion-breadcrumbs` components were introduced in v6. We can delete those demo examples from:
- `static/demos/api/accordion/index.html`
- `static/demos/api/breadcrumbs/index.html`

They were introduced here: https://github.com/ionic-team/ionic-docs/pull/2087, but this was a flux time before we had component playgrounds for v6 and were applied to what we consider as the "v5 examples". 

Once those are removed, this PR should be good to go.

brandyscarney:
I found the same as Sean, but also we redirect the users to https://ionic.io/ionicons/ for the icon demo, so that one can be removed too:

```
static/demos/api/icon/index.html
```

This looks good though!

sean-perkins (approved):
Well done 👏

PR #2848, opened 2023-03-22
docs: update dev workflow instructions

thetaPC:
`npm install` fails due to:
1. having the wrong version of `npm` and `node`
2. peer dependencies

Instructions have been modified to overcome this until a permanent fix is implemented.


sean-perkins (approved):
Nice work, thanks for updating this 👍 

I created FW-3798 to track the work effort for fixing the ionic-docs project to not require legacy peer deps in a future sprint.

PR #2853, opened 2023-03-24
docs(input, textarea): clarify error text behavior

liamdebeasi:
Someone on the Ionic Discord noted that we do not state `.ion-touched` is needed. This was a bug we fixed in https://github.com/ionic-team/ionic-framework/commit/ef33270b55122574e0fdb32187410d8d8a4fa1ae, but we did not update the docs.


brandyscarney (approved):
LGTM

PR #2854, opened 2023-03-24
docs(toast): remove hardware back button dismissal

thetaPC:
Removed any mentions that `ion-toast` can use the hardware back button to dismiss. The change was added to v6 and v7 (beta).

### DIFF
--- a/docs/api/toast.md
+++ b/docs/api/toast.md
@@ -48,9 +48,11 @@ import ControllerExample from '@site/static/usage/v7/toast/presenting/controller
 
 ## Dismissing
 
-Toasts are intended to be subtle notifications and are not intended to interrupt the user.
+Toasts are intended to be subtle notifications and should not interrupt the user. As a result, user interaction should not be required to dismiss the toast.
 
-The toast can be dismissed automatically after a specific amount of time by passing the number of milliseconds to display it in the `duration` of the toast options. If a button with a role of `"cancel"` is added, then that button will dismiss the toast. To dismiss the toast after creation, call the `dismiss()` method on the instance. A hardware back button will not dismiss the toast.
+The toast can be dismissed automatically after a specific amount of time by passing the number of milliseconds to display it in the `duration` of the toast options. If a button with a role of `"cancel"` is added, then that button will dismiss the toast. To dismiss the toast after creation, call the `dismiss()` method on the instance.
+
+Pressing the hardware back button does not dismiss toasts since they are not supposed to interrupt the user.
 
 The following example demonstrates how to use the `buttons` property to add a button that automatically dismisses the toast when clicked, as well as how to collect the `role` of the dismiss event.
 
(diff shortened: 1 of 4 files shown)


versioned_docs/version-v6/api/toast.md
```diff
@@ -91,7 +91,7 @@ function Example() {
 
 ## Dismissing
 
-The toast can be dismissed automatically after a specific amount of time by passing the number of milliseconds to display it in the `duration` of the toast options. If a button with a role of `"cancel"` is added, then that button will dismiss the toast. To dismiss the toast after creation, call the `dismiss()` method on the instance.
+The toast can be dismissed automatically after a specific amount of time by passing the number of milliseconds to display it in the `duration` of the toast options. To dismiss the toast after creation, call the `dismiss()` method on the instance.
```
liamdebeasi, line 94: Why was this removed? Buttons in toasts with a `cancel` role dismiss the toast: https://github.com/ionic-team/ionic-framework/blob/f68644065c227b73fe49330b030e2cfdad7b603a/core/src/components/toast/toast.tsx#L238-L240

thetaPC: I misunderstood the wording and thought it was related to the hardware back button. I'll revert back.

liamdebeasi (approved):
This looks good. Is the update to the toast docs going to be a separate PR?

thetaPC:
> This looks good. Is the update to the toast docs going to be a separate PR?

Let me know if the wording needs to be rephrased. It's a combo from your words and mine.

docs/api/toast.md
```diff
@@ -48,7 +48,9 @@ import ControllerExample from '@site/static/usage/v7/toast/presenting/controller
 
 ## Dismissing
 
-The toast can be dismissed automatically after a specific amount of time by passing the number of milliseconds to display it in the `duration` of the toast options. If a button with a role of `"cancel"` is added, then that button will dismiss the toast. To dismiss the toast after creation, call the `dismiss()` method on the instance.
+Toasts are intended to be subtle notifications and are not intended to interrupt the user.
+
+The toast can be dismissed automatically after a specific amount of time by passing the number of milliseconds to display it in the `duration` of the toast options. If a button with a role of `"cancel"` is added, then that button will dismiss the toast. To dismiss the toast after creation, call the `dismiss()` method on the instance. A hardware back button will not dismiss the toast.
```
liamdebeasi, line 53:
```suggestion
Toasts are intended to be subtle notifications and should not interrupt the user. As a result, user interaction should not be required to dismiss the toast.

The toast can be dismissed automatically after a specific amount of time by passing the number of milliseconds to display it in the `duration` of the toast options. If a button with a role of `"cancel"` is added, then that button will dismiss the toast. To dismiss the toast after creation, call the `dismiss()` method on the instance.

Pressing the hardware back button does not dismiss toasts since they are not supposed to interrupt the user.
```

How about something like the above?

liamdebeasi (changes requested):

liamdebeasi (approved):
Thank you!

PR #2859, opened 2023-03-27
chore(many): update v7 playgrounds to point to Ionic 7.0.0

amandaejohnston:
Updates the `demo.html` for all v7 playgrounds, as well as the Stackblitz app files, to use Ionic 7.0.0 instead of v6, a dev build, the `next` tag, etc. Note that until v7 is officially released, this will cause all playgrounds to fail to load.

When reviewing, it may be easiest to look at the two commits separately since the playgrounds update was pretty much just a big find-and-replace.


static/code/stackblitz/v7/react/package-lock.json
liamdebeasi, line 1: We'll likely need to update this again once v7 is out since the package-lock still maps to an old dev build: https://github.com/ionic-team/ionic-docs/blob/a47e193ee22d2864074db7b13f9c29bd39e0e438/static/code/stackblitz/v7/react/package-lock.json#L2249-L2250

liamdebeasi: Same thing for the Vue package-lock

liamdebeasi (approved):

liamdebeasi:
FYI Your preview deployment will fail until the latest version of the docs is built. Needed to merge https://github.com/ionic-team/ionic-docs/pull/2829.

I'll ping you here when this is fixed.

PR #2873, opened 2023-03-30
Update slides.md

devsalimsulaiman:
the register() should be called inside constructor


amandaejohnston:
Thank you for the PR. The `register` function is called outside of the component, so the proposed change isn't syntactically correct. Is there a particular problem you're running into with the current code?

amandaejohnston:
I'm going to close this out given the above. If you're still running into trouble, please [open an issue](https://github.com/ionic-team/ionic-docs/issues/new/choose) and we can take another look.

PR #2875, opened 2023-03-31
chore(docs contributing): update code of conduct to link out to full …

johnwheeler-prod:
…company code of conduct


jaredcbaum (changes requested):
Can you remove the period after "here" and the noreferrer from the 'rel' attributes? We want to be able to capture referral traffic between our sites.

jaredcbaum (approved):

PR #2877, opened 2023-04-04
Add optimizeDeps exclusion note for vite config

mvanderlinde:
The current instructions for building one's first app using Vue + Vite does not work, and results in a `net::ERR_ABORTED 504 (Outdated Optimize Dep)` error.

Based on the discussion in https://github.com/ionic-team/pwa-elements/issues/109 — and confirmed by my own testing — excluding the `pwa-elements` loader from Vite's dependency optimization fixes the issue in development.

It seems there may be lingering issues in production builds even with this fix based on that same open issue, but at the very least, this removes a current blocker in the documentation for users creating their first Ionic app.

### DIFF
--- a/docs/vue/your-first-app.md
+++ b/docs/vue/your-first-app.md
@@ -109,6 +109,15 @@ import { defineCustomElements } from '@ionic/pwa-elements/loader';
 defineCustomElements(window);
 ```
 
+Since we're using Vite, you'll need to also add the following code to `vite.config.ts`.
+
+```tsx
+// Within the defineConfig block
+optimizeDeps: {
+  exclude: [`@ionic/pwa-elements/loader`],
+},
+```
+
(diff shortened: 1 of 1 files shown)


b4-io:
The lingering production problems are that vite doesn't include any of the pwa-elements libraries in the final build. 
I don't find this to be an acceptable change as this isn't needed at all if the problem is fixed from the core.

sean-perkins:
Hello @mvanderlinde thanks for this suggested docs change. 

Reviewing the active discussion on https://github.com/ionic-team/pwa-elements/issues/109, I do not believe this is the best way to resolve this issue. It may suffice as a workaround to avoid the error, but it does not have the same behavior that the developer expects. The Ionic team should prioritize resolving the issue in PWA elements that is causing this exception to occur.

PR #2879, opened 2023-04-05
docs(deployment): typo with windows requirements

robertegj:
Assuming the Windows section was copied from the macOS section. Should read four instead of two.


sean-perkins (approved):
Thanks for updating this!

PR #2883, opened 2023-04-06
chore(jp): merge recent update

rdlabo:
merge recent update and some translate. (A few commits are in there for https://ionicframework.jp/docs/, but they don't affect https://ionicframework.com/docs/ja/)


liamdebeasi (approved):

liamdebeasi:
Thanks for taking care of this!

PR #2861, opened 2023-03-28
docs(react): writing tests with testing-library

sean-perkins:
Updates the React testing documentation to use Vitest and testing-library.

Preview: https://ionic-docs-git-docs-react-testing-ionic1.vercel.app/docs/react/testing/introduction

Resolves: #2969

PR #2886, opened 2023-04-06
fix(styles): add sidebar indentation for nested sections

sean-perkins:
Updates the indentation for nested categories in the sidebar and the margin spacing between sections. 

You can see an example of a nested sidebar 3 levels deep with these same changes in this PR: https://github.com/ionic-team/ionic-docs/pull/2861 (React > Testing > Unit Testing)

Our goal is to maintain the spacing and indentation for the first-level categories, but allow the nested categories to be indented. 

CleanShot 2023-04-06 at 11 43 10

### DIFF
--- a/src/styles/components/_doc-sidebar.scss
+++ b/src/styles/components/_doc-sidebar.scss
@@ -103,9 +103,8 @@ html[data-theme='dark'] {
     }
   }
 
-  .theme-doc-sidebar-item-category-level-1 + .theme-doc-sidebar-item-category-level-1 {
-    /* adds margin between first level categories */
-    margin-block-start: 1.5rem;
+  .theme-doc-sidebar-item-category-level-1 > .menu__list > .menu__list-item:last-of-type {
+    padding-block-end: 1.5rem;
   }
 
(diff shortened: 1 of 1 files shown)


jaredcbaum (changes requested):
Screenshot 2023-04-06 at 4 02 49 PM

I think we'll still want margin at the end of top level categories to conform to the design.

sean-perkins:
@jaredcbaum good catch! I moved the styles down after this:

```css
&:not(:first-child) {
  margin-block-start: 0;
}
```

This should re-introduce the spacing between sections (and at the bottom of the last section).

jaredcbaum:
Screenshot 2023-04-10 at 8 42 32 AM

This still maintains the spacing when the items are collapsed. Not sure if you intended it to be this way, but figured I'd point it out just in case.

sean-perkins:
Ok third time may be the charm. I see why we were targeting the inner list for padding vs. margins between the top-level categories. 

Here is a preview against the other PR to show 3 levels deep:

CleanShot 2023-04-10 at 11 51 34

jaredcbaum (approved):
Looks great!

PR #2887, opened 2023-04-06
docs(select): popover can be used with multiple prop

liamdebeasi:
Multi-select popovers in `ion-select` were added in https://github.com/ionic-team/ionic-framework/pull/23474, but the docs still say you can't use popovers with `multiple="true"`.


thetaPC (approved):
LGTM

PR #2889, opened 2023-04-06
Revert "fix(jp): set the v7 document version to v7"

liamdebeasi:
Reverts ionic-team/ionic-docs#2884

This commit is causing the build to fail.

### DIFF
--- a/scripts/i18n.sh
+++ b/scripts/i18n.sh
@@ -2,7 +2,7 @@
 set -o errexit
 set -o nounset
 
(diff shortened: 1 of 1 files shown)


liamdebeasi:
@rdlabo Your previous commit caused the deployment to fail. I'd like to roll back your previous commit so the team can investigate why the build failed.

mapsandapps (approved):

rdlabo (approved):
Understood. The "current" path may be wrong now.

rdlabo:
@liamdebeasi What should I do about this problem?

PR #2891, opened 2023-04-07
fix(modal): controller playground uses v7 label syntax

sean-perkins:
Updates the v7 controller playground examples to use the modern syntax. Also updates the inconsistencies with the form label across the different frameworks.


amandaejohnston (approved):
The React and Vue examples just need the now-unneeded imports for `ion-label` removed, but otherwise this LGTM 👍

brandyscarney (approved):

PR #2894, opened 2023-04-10
Update lifecycle.md

FernetB:
Adds the `useLocation` problem into the docs

### DIFF
--- a/docs/react/lifecycle.md
+++ b/docs/react/lifecycle.md
@@ -145,3 +145,9 @@ Below are some tips on use cases for each of the life cycle events.
 - `ionViewDidEnter` - If you see performance problems from using `ionViewWillEnter` when loading data, you can do your data calls in `ionViewDidEnter` instead. This event won't fire until after the page is visible to the user, however, so you might want to use either a loading indicator or a skeleton screen, so content doesn't flash in un-naturally after the transition is complete.
 - `ionViewWillLeave` - Can be used for cleanup, like unsubscribing from data sources. Since `componentWillUnmount` might not fire when you navigate from the current page, put your cleanup code here if you don't want it active while the screen is not in view.
 - `ionViewDidLeave` - When this event fires, you know the new page has fully transitioned in, so any logic you might not normally do when the view is visible can go here.
+
+## Passing state between pages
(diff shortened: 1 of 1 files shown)


docs/react/lifecycle.md
```diff
@@ -145,3 +145,9 @@ Below are some tips on use cases for each of the life cycle events.
...
 - `ionViewWillLeave` - Can be used for cleanup, like unsubscribing from data sources. Since `componentWillUnmount` might not fire when you navigate from the current page, put your cleanup code here if you don't want it active while the screen is not in view.
 - `ionViewDidLeave` - When this event fires, you know the new page has fully transitioned in, so any logic you might not normally do when the view is visible can go here.
+
+## Passing state between pages
+You can pass states between pages like normal with `history.push()` but need to be **aware** that ionic keep the pages mounted and with that if you navigate from a page A with `stateA` to a page B with `stateB`, the page A will be hidden and the state of that page will no longer be `stateA` instead will be `stateB` and it may crash your app, for that you need to add an undefined check `?` whenever you want to access to the properties of an `state`
+
+If you want to avoid forgetting to add undefined check you could do something like 
+`useCustomLocation<T> = useLocation<Partial<T>>`
```
liamdebeasi, line 153:
```suggestion
Since Ionic React manages the lifetime of a page, state on previous pages may update as users navigate your application. Developers should include the appropriate checks to ensure that previous pages only access defined states.

For example, the following code will error if `testObject` is not defined: `{ state.testObject.childKey }`

Instead, developers should access `childKey` only if `testObject` is defined: `{ state.testObject?.childKey }`
```

I changed some of the wording and added an example so developers have a reference for how to account for this.

liamdebeasi (changes requested):

docs/react/lifecycle.md
FernetB: Nice! But I thing that needs to be clarify that the state of the first page will become the state of the second page
Why is this important? Because if you go from A that has for a example a state with `person.id = 1` to page B with state `person.id = 2`, then if you have some sort of conditional fetching that fetches when the `person.id` changes (like swr) or an `useEffect` that has that in the array of dependecies this may have undesired side effects. That's why I put special emphasis on 
>the page A will be hidden and the state of that page will no longer be `stateA` instead will be `stateB`

liamdebeasi: I changed the first paragraph to have the following: 

> Since Ionic React manages the lifetime of a page, state on previous pages may update as users navigate your application. This can impact state that is determined using `useEffect` from React or `useLocation` from React Router. For example, if `PageA` calls `useLocation`, the state of `useLocation` will change when the user navigates from `PageA` to `PageB`.


How does that sound?

liamdebeasi (approved):

liamdebeasi:
Merged. Thank you!

PR #2896, opened 2023-04-10
refactor(docs): unversion the CLI portion of the docs

liamdebeasi:
This PR does a few things:

1. Removes versioned copies of the CLI docs.
2. Moves the CLI content from the `docs` directory up a level.
3. Updates `docusaurus.config.js` to reference the CLI given this new structure.
4. Updates the `crowdin.yml` file to allow the CLI documentation (minus the auto generated commands) to be translated.

Note 1: The Guide will have several broken links for now. The Guide is versioned and is trying to referenced versioned pages of The CLI. Once Guide is un-versioned this should be resolved.
Note 2: The version dropdown has not been moved yet. This will be moved in a future PR.


amandaejohnston (approved):
I noticed that if you visit any of the CLI pages, and then use the version dropdown to switch to an earlier version of Ionic, you get booted to the docs homepage. I assume that'll be fixed/avoided once the version dropdown is moved in a future PR?

liamdebeasi:
Yeah I plan on moving the version selector to only show up on the Components/API pages.

PR #2901, opened 2023-04-10
docs(nav): clearly state that it's not used for routing

thetaPC:

`nav` doesn't clearly state that it's not used for routing.

## What is the new behavior?

For v7 and v6:
- Provides an example
- Provides an alternative

- [ ] Yes
- [x] No

## Other information

N/A


amandaejohnston (changes requested):
`ion-router` is only used for pure JS/Stencil projects (see [the docs](https://ionicframework.com/docs/api/router)). Could we instead do something like this?

```
Instead, see the routing guides for [Angular](../angular/navigation), [React](../react/navigation), and [Vue](../vue/navigation), or [`ion-router`](./router) for pure JavaScript projects.
```

amandaejohnston (approved):

Issue #1507, opened 2020-06-25
Color Generator not updating main color when replacing with a value without hashtag


noahhorlacher:
**Describe the bug**
On the [ionic color generator page](https://ionicframework.com/docs/theming/color-generator), the corresponding '-shade' and '-tint' sub-variables don't update when replacing a hexcode in the main variable input fields with a value missing the hashtag.

**To reproduce**
1. Go to 'https://ionicframework.com/docs/theming/color-generator'
2. Click on the arrow to the right of any color input field such as 'Primary' to see the sub-variables
3. Select the current value (click on the input field and press CTRL+A) including the hashtag
4. Paste a hex code value missing the hashtag e.g. ff0000
5. The site auto-inserts the hashtag at the beginning, but doesn't auto-update the colors and values of the sub-variables, until you type in another character. It also doesn't update if it's not in focus anymore.

**Expected behavior**
After the site prepends a hashtag to the value, it's expected to auto-update the sub-variables.

**Screenshots**
![screenshot](https://user-images.githubusercontent.com/32394110/85674474-27c11800-b6c5-11ea-8a21-a6bef0ce2a25.png)

**Browser and OS**
 - OS: Ubuntu 20.04 LTS
 - Browser: Firefox, Google Chrome
 - Version: 77.0.1, 83.0.4103.116

acran:
I tracked down the reason for this to this block:
https://github.com/ionic-team/ionic-docs/blob/bdf3a610a72783567e5c8868101dbd1bcf214176/src/components/color-gen/variable-selector/color-gen-variable-selector.tsx#L48-L51

The fix should be to simply drop the `return` since it will still ensure that the input starts with a `#` and if it not yet a valid input (i.e. 7 characters long) it will be skipped by the subsequent check anyway:
https://github.com/ionic-team/ionic-docs/blob/bdf3a610a72783567e5c8868101dbd1bcf214176/src/components/color-gen/variable-selector/color-gen-variable-selector.tsx#L58

It would also need to update `val`, too, and declaring it with `let` instead of `const`. So the fixed code would be:
~~~ts
    let val = input.value.trim();

    if (input.matches('[type="text"]') && val[0] !== '#') {
      input.value = '#' + val;
      val = input.value;
    }
~~~

amandaejohnston:
[...]

PR #2904, opened 2023-04-11
fix(ColorGenerator): always prepend entered colors with pound sign

amandaejohnston:
In the color generator at https://ionicframework.com/docs/theming/color-generator, if you enter a color without a `#`, the generated shade and tint will be properly updated, but the color dot to the left of the main color won't update. This PR ensures there's always a `#` at the start of the input values.

Resolves https://github.com/ionic-team/ionic-docs/issues/1507


brandyscarney (approved):
Looks great!

PR #2910, opened 2023-04-12
docs(angular): add type assertion for your first app code snippet

josh33:
Was getting an error: "Type 'string | undefined' is not assignable to type 'string'." Adding an exclamation point at the end of line 111 solves it.


sean-perkins (approved):
Thanks for submitting this fix 👍

PR #2913, opened 2023-04-13
docs(searchbar): debounce uses ionInput

liamdebeasi:
The debounce demo for v7 uses `ionChange` when it should use `ionInput`: https://github.com/ionic-team/ionic-docs/pull/2913

> The debounce property has been updated to control the timing in milliseconds to delay the event emission of the ionInput event after each keystroke. Previously it would delay the event emission of ionChange.

More info: https://github.com/ionic-team/ionic-framework/issues/27192#issuecomment-1506908563

```
Co-authored-by: guillaumesmo <guillaumesmo@users.noreply.github.com>
```

### DIFF
--- a/docs/api/searchbar.md
+++ b/docs/api/searchbar.md
@@ -66,7 +66,7 @@ import Toolbar from '@site/static/usage/v7/toolbar/searchbars/index.md';
 
 ## Debounce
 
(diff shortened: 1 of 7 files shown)


thetaPC (approved):
LGTM

PR #2915, opened 2023-04-14
docs(your-first-app): switch to Date.now()

Biskweet:
I replaced `new Date().getTime()` with the static method `Date.now()` because the previous was over-engineering and had worse performance.


thetaPC (changes requested):
The change needs to be applied:
- to the other steps of React: Your First App
- to the other frameworks (Angular: Your First App & Vue: Your First App)
- v6 for all the frameworks (React: Your First App, Angular: Your First App, & Vue: Your First App)

thetaPC (approved):
LGTM

PR #2918, opened 2023-04-15
Fix typo

RemcoSimonides:


brandyscarney:
Thank you for the PR! However, the current documentation is correct. The `ion-tabs` component should be used with Angular, Vue and React. You can see the usage information for those [here](https://ionicframework.com/docs/api/tabs#usage-with-router). You will notice that `ion-tab` is only used in the `JavaScript` example. I am going to close this since no updates need to be made. 🙂

PR #2920, opened 2023-04-18
Update 3-saving-photos.md

22017498uhi:
typo on line 53. it should be "webviewpath" instead of "webviewPath" ?


liamdebeasi:
Thanks for the PR! I double checked the tutorial, and the current spelling of `webviewPath` is correct. That property is accessed when building the "Displaying Photos" functionality: https://ionicframework.com/docs/angular/your-first-app/taking-photos#displaying-photos. I am going to close this, but let me know if you have questions.

PR #2922, opened 2023-04-19
fix(playground): allow automatic mode selection when relevant base file is overwritten

amandaejohnston:
Example playgrounds that exhibit this bug:
- Angular (replaces `src/app/app.module.ts`): https://ionicframework.com/docs/api/nav#using-navlink
- Javascript (replaces `index.html`): https://ionicframework.com/docs/api/modal#animations
- React (replaces `src/app.tsx`): none yet
- Vue (replaces `src/main.ts`): https://ionicframework.com/docs/api/tabs#usage-with-router

### DIFF
--- a/src/components/global/Playground/stackblitz.utils.ts
+++ b/src/components/global/Playground/stackblitz.utils.ts
@@ -79,7 +79,7 @@ const openHtmlEditor = async (code: string, options?: EditorOptions) => {
 }
 
 const openAngularEditor = async (code: string, options?: EditorOptions) => {
-  let [main_ts, app_module_ts, app_component_ts, app_component_css, app_component_html, example_component_ts, styles_css, global_css, variables_css, angular_json, tsconfig_json, package_json] = await loadSourceFiles([
+  const defaultFiles = await loadSourceFiles([
     'angular/main.ts',
     'angular/app.module.ts',
     'angular/app.component.ts',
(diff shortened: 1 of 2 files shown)


src/components/global/Playground/stackblitz.utils.ts
```diff
@@ -92,18 +108,40 @@ const openAngularEditor = async (code: string, options?: EditorOptions) => {
...
+  const package_json = defaultFiles[11];
 
   if (options.angularModuleOptions) {
     if (options.angularModuleOptions.imports) {
-      app_module_ts = `${options.angularModuleOptions.imports.join('\n')}\n${app_module_ts}`;
+      files[appModule] = `${options.angularModuleOptions.imports.join('\n')}\n${files[appModule]}`;
     }
     if (options.angularModuleOptions.declarations) {
-      app_module_ts = app_module_ts.replace('/* CUSTOM_DECLARATIONS */', options.angularModuleOptions.declarations.map(d => `\n  ${d}`).join(','));
+      files[appModule] = files[appModule].replace('/* CUSTOM_DECLARATIONS */', options.angularModuleOptions.declarations.map(d => `\n  ${d}`).join(','));
```
amandaejohnston, line 140: I'm not really sure what this replace is for; the `/* CUSTOM_DECLARATIONS */` piece wasn't present in the template even before my changes. I left it as-is just in case. Functionality should be the same as before.

sean-perkins: This is a legacy feature that we never ended up needing. Before the multi-file playgrounds were a thing, I had experimented with a configuration object that would just append the import statements and declarations to your existing playground code, instead of having to re-create the `app.module.ts` yourself. 

I think we could kill this feature (it isn't used in any playground to date).

amandaejohnston: Does that include the whole `angularModuleOptions` feature? Or just this `CUSTOM_DECLARATIONS` piece? If the former, I'll whip up a separate PR for it later 👀

sean-perkins: Yes, it would be the entire `angularModuleOptions` configuration/feature.

sean-perkins (commented):
The base files used for the v7 stackblitz are modified, but I don't see diffs for the v6 assets or the `withContent` variant files, e.g.: https://github.com/ionic-team/ionic-docs/blob/eb94947667ed56176388ff8232db36f1875a960b/static/code/stackblitz/v6/react/app.withContent.tsx#L25-L27

Do those files need to be additionally updated to support this?

amandaejohnston:
Good catch! Got those updated too.

sean-perkins (approved):

PR #2949, opened 2023-05-01
docs(breadcrumbs): update vue example

mapsandapps:
Because of a bug ([now fixed](https://github.com/ionic-team/ionic-framework/pull/27040)), the Vue example for breadcrumbs was setting the color on the individual `ion-breadcrumb`s. We can now set it on the `ion-breadcrumbs` component.

### DIFF
--- a/static/usage/v7/breadcrumbs/theming/colors/vue.md
+++ b/static/usage/v7/breadcrumbs/theming/colors/vue.md
@@ -1,10 +1,10 @@
 ```html
 <template>
-  <ion-breadcrumbs>
-    <ion-breadcrumb color="primary" href="#home">Home</ion-breadcrumb>
-    <ion-breadcrumb color="primary" href="#electronics">Electronics</ion-breadcrumb>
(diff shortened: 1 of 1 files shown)


sean-perkins (changes requested):
The Stackblitz example for Vue is continuing to show the bug: https://stackblitz.com/edit/pmidpz?file=package-lock.json

You will need to additionally update the `package-lock.json` file that is used as the template for the Stackblitz project. It unfortunately doesn't load the latest automatically (we have to include a lock file or the boot time is super slow). So when we are patching a playground with a bug fix or a new feature, we need to manually update the lock file for that framework's project. 

Steps:
- Navigate into this folder: https://github.com/ionic-team/ionic-docs/tree/main/static/code/stackblitz/v7/vue
- `npm install`
- Validate the lock file is referencing the latest version of `@ionic/vue`.

sean-perkins (approved):
Changes look great, nice work 👍

PR #2955, opened 2023-05-09
docs(range): update docs to use label prop

amandaejohnston:
Feature PR: https://github.com/ionic-team/ionic-framework/pull/27408

Rather than add a playground for the new label prop, this PR makes the prop the default by updating the `labelPlacement` playground and modern syntax migration guide to use it, then adds a playground and migration example for the slot.

### DIFF
--- a/docs/api/range.md
+++ b/docs/api/range.md
@@ -116,7 +116,9 @@ Developers can perform this migration one range at a time. While developers can 
 
 ### Using the Modern Syntax
 
-Using the modern syntax involves removing the `ion-label` and passing the label directly inside of `ion-range` using `slot="label"`. The placement of the label can be configured using the `labelPlacement` property on `ion-range`.
+Using the modern syntax involves removing the `ion-label` and passing the label to `ion-range` using the `label` property. The placement of the label can be configured using the `labelPlacement` property.
+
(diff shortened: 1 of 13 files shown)


mapsandapps (approved):

docs/api/range.md
```diff
@@ -34,6 +34,14 @@ import LabelsPlayground from '@site/static/usage/v7/range/labels/index.md';
 
 <LabelsPlayground />
 
+## Label Slot
```
liamdebeasi, line 37: Threading for discussion:

I wonder if it makes more sense to have a "Labels" heading with Basic, Placement, and Custom HTML playgrounds.

We could also combine the Basic and Placement playgrounds into 1 playground that shows how to use the `label` property with placement options. If we went with the 2nd approach, it looks like we'd just need to change some headings on the page and keep the playgrounds as-is.

liamdebeasi: As added context: My initial reaction was that it seems strange to have a Label Placement and Label Slot, but no "Label" section. Placement and Slot seem like they are sub-features for the overall "Label" feature.

amandaejohnston: Rearranged things 👍

liamdebeasi (approved):
One comment, but otherwise looks good!

PR #2956, opened 2023-05-10
docs(vue): streamline usage examples with script setup syntax

treighmawaka:
Problem:
Many usage examples in the current docs use an outdated syntax containing unnecessary bloat.

Solution:
This PR streamlines code snippets by using [Vue's officially recommended](https://vuejs.org/api/sfc-script-setup.html) `<script setup>` syntax, making usage examples more digestible.


mapsandapps (approved):
Thanks for your contribution! This is helpful.

PR #2957, opened 2023-05-10
fix(tabs): correct router playground src

thetaPC:
Issue URL: N/A

Tabs has a playground for router that was using v6's `demo.hmtl`

## What is the new behavior?

Switched it to v7's `demo.hmtl`

- [ ] Yes
- [x] No

## Other information

N/A


amandaejohnston (approved):
Thank you!

PR #2960, opened 2023-05-13
fix position <ion-router-outlet> on html

skytern:
Because the position it's wrong, the position inside <ion-tabs> create duplication <ion-router-outlet> and we don't change tab.


sean-perkins:
Hello @skytern thanks for submitting this fix!

The `ion-router-outlet` isn't needed in Angular with `ion-tabs`, that is a mistake on our end. I updated the docs for v6 and v7 in this PR: https://github.com/ionic-team/ionic-docs/pull/2965 and gave you co-authored-by credit.

Thanks for helping us improve the documentation!

PR #2961, opened 2023-05-15
fix: Ionic CLI links to latest docs page 

zakuru:
https://github.com/ionic-team/ionic-docs/assets/171174/d009bcad-aacf-452e-a135-f53b3e087c2f


sean-perkins (approved):
Thanks for correcting this!

PR #2966, opened 2023-05-19
docs(spinner): add docs for resizing

mapsandapps:
In https://github.com/ionic-team/ionic-framework/issues/18115, a user recommended adding an example to the docs for the functionality added in https://github.com/ionic-team/ionic-framework/pull/27424.

### DIFF
--- a/docs/api/spinner.md
+++ b/docs/api/spinner.md
@@ -43,6 +43,14 @@ import CSSProps from '@site/static/usage/v7/spinner/theming/css-properties/index
 
 <CSSProps />
 
+## Styling
+
+### Resizing
+
+import Resizing from '@site/static/usage/v7/spinner/theming/resizing/index.md';
+
+<Resizing />
+
 
 ## Properties
 <Props />
--- a/static/usage/v7/spinner/theming/resizing/demo.html
+++ b/static/usage/v7/spinner/theming/resizing/demo.html
@@ -1,0 +1,29 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Spinner</title>
+    <link rel="stylesheet" href="../../../../common.css" />
+    <script src="../../../../common.js"></script>
+    <script type="module" src="https://cdn.jsdelivr.net/npm/@ionic/core@6/dist/ionic/ionic.esm.js"></script>
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@ionic/core@6/css/ionic.bundle.css" />
+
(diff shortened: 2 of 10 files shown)


docs/api/spinner.md
```diff
@@ -43,6 +43,14 @@ import CSSProps from '@site/static/usage/v7/spinner/theming/css-properties/index
 
 <CSSProps />
 
+## Styling
+
+### Resizing
+
```
mapsandapps, line 49: I wasn't sure what to name/where to put this. I'm also not sure if we want to add this to the docs at all. Feel free to reject if that's appropriate 🙂

thetaPC: I lean towards following the structure that buttons has for the [size section](https://ionicframework.com/docs/v6/api/button#size).

static/usage/v7/spinner/theming/resizing/demo.html
```diff
@@ -0,0 +1,29 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Spinner</title>
+    <link rel="stylesheet" href="../../../../common.css" />
+    <script src="../../../../common.js"></script>
+    <script type="module" src="https://cdn.jsdelivr.net/npm/@ionic/core@6/dist/ionic/ionic.esm.js"></script>
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@ionic/core@6/css/ionic.bundle.css" />
```
thetaPC, line 10: These need to point to `core@7`

thetaPC (changes requested):

docs/api/spinner.md
mapsandapps: Hmm, I had looked at that one, but for button, `size` is an attribute, and it seems like most of the top-level items like that are attributes.

thetaPC: Good point regarding the attributes.

I also agree that it follows the action sheet more.

docs/api/spinner.md
```diff
@@ -37,19 +37,19 @@ import Colors from '@site/static/usage/v7/spinner/theming/colors/index.md';
 
 <Colors />
 
-### CSS Custom Properties
+### Styling
 
-import CSSProps from '@site/static/usage/v7/spinner/theming/css-properties/index.md';
+You may use custom CSS to style the spinner. For example, you can resize the spinner by setting the width and height.
```
thetaPC, line 42: Suggestion to change: `You can resize the spinner by setting the width and height.`

thetaPC (approved):
LGTM

brandyscarney (approved):
Looks great!

PR #2970, opened 2023-05-23
feat(playground): inline dependencies per framework example

sean-perkins:
This PR adds support for inline dependencies for individual framework examples. For example, this allows you to specify required dependencies that your component playground may need, without having to ship that dependency with every single component playground for all Ionic docs examples.

Example usage:
```tsx
<Playground
  version="7"
  code={{
    javascript: {
      files: {
        'index.html': javascript_index_html,
        'index.ts': javascript_index_ts,
      },
      dependencies: {
        '@maskito/core': '^0.11.0',
      },
    },
  }}
  src="usage/v7/input/mask/demo.html"
  size="300px"
/>
```

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -85,6 +85,20 @@ interface UsageTargetOptions {
   files: {
     [key: string]: MdxContent;
   };
+  /**
+   * The list of dependencies to use in the Stackblitz example.
+   * The key is the package name and the value is the version.
+   * The version must be a valid semver range.
+   *
+   * For example:
+   * ```ts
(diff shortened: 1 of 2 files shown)


liamdebeasi (approved):
A couple typos but otherwise looks good 👍

brandyscarney (approved):
Tested it out locally. It works great when you pass a valid dependency, and only applies the dependency to the `package.json` of the specified framework. StackBlitz errors and drops all dependencies if you pass an invalid one, but I think this is okay since it should indicate rather quickly where the problem code is:

Screenshot 2023-05-23 at 10 28 24 AM

PR #1776, opened 2020-12-04
docs(update-page): add a link to vue pwa (including vue-svg logo) in deployment/pwa page

VAS000:
- Added vue svg logo ==> /src/assets/img/frameworks/vue.svg
- Added a card with link/logo to pwa page for vue

PR #2974, opened 2023-05-25
docs(PWAs): add Vue to list of PWA deployment guides

amandaejohnston:
The PWA deployment index [here](https://ionicframework.com/docs/deployment/progressive-web-app) already has links to guides for all three frameworks, but the text only lists Angular and React.

Essentially the same as https://github.com/ionic-team/ionic-docs/pull/1776, but the link and logo asset had already been added at some point; only the text needed updating. Co-author credit has been given.


liamdebeasi (approved):
Thanks!

PR #613, opened 2019-04-04
docs(navigation): support relative URLs.

rdlabo:
remove note of not support relative URLs, and add sample using relative URLs.

ref: https://github.com/ionic-team/ionic/commit/d9a7c63

PR #2980, opened 2023-05-30
docs(angular): relative routing is supported

sean-perkins:
Relative routing in Angular is supported in Ionic Framework: https://github.com/ionic-team/ionic-framework/commit/d9a7c63

This PR is in favor of: https://github.com/ionic-team/ionic-docs/pull/613


amandaejohnston (commented):
It looks like a fair amount of Prettier changes unrelated to the fix have snuck in here. Would it be possible to move those to a different PR?

amandaejohnston (approved):
Thanks!

PR #2982, opened 2023-05-31
docs(datetime): add playground for styling wheel pickers

amandaejohnston:
Feature PR: https://github.com/ionic-team/ionic-framework/pull/27529
Shortcut: https://ionic-docs-git-fw-2254-ionic1.vercel.app/docs/api/datetime#wheel-pickers

Moves the existing Theming playground to a subdirectory, and adds a new Wheel Styling playground. Dev build for testing: `7.0.7-dev.11685554390.10c2ca9b` (You'll need this for the new features to show up in the iframes.)

I intentionally didn't include the new `time-button` shadow part, as I feel this wheel styling pattern will be common enough to warrant its own playground, and anything else felt like clutter. We don't always include every single shadow part and CSS variable in playgrounds, and the `time-button` one seems self-explanatory to me.


mapsandapps (approved):
I tried it out. The explanation and examples are clear. Thanks!

Issue #2986, opened 2023-06-02
content: your-first-app, readAsBase64 function has type error in Angular


implicitdefcncdragneel:
### URL

https://github.com/ionic-team/ionic-docs/blob/main/docs/angular/your-first-app/5-adding-mobile.md

### Issue Description

**What is missing or inaccurate about the content on this page?**

Missing  : **fileName**  
Description : path: photo.path generating error as Type 'string | undefined' is not assignable to type 'string'.

Modified Code : 

```
private async readAsBase64(photo: Photo) {
  // "hybrid" will detect Cordova or Capacitor
  const fileName = new Date().getTime() + '.jpeg';
  
  if (this.platform.is('hybrid')) {
    // Read the file into base64 format
    const file = await Filesystem.readFile({
      path: fileName
    });

    return file.data;
  }
  else {
      // Fetch the photo, read as a blob, then convert to base64 format
      const response = await fetch(photo.webPath!);
      const blob = await response.blob();

      return await this.convertBlobToBase64(blob) as string;
    }
  }
```

PR #2987, opened 2023-06-02
docs(angular):updated component photo service (#2986)

implicitdefcncdragneel:
Resolves [#2986](https://github.com/ionic-team/ionic-docs/issues/2986)

- Added fileName in readAsBase64

### DIFF
--- a/docs/angular/your-first-app/5-adding-mobile.md
+++ b/docs/angular/your-first-app/5-adding-mobile.md
@@ -35,7 +35,6 @@ First, we’ll update the photo saving functionality to support mobile. In the `
 ```tsx
 private async readAsBase64(photo: Photo) {
   // "hybrid" will detect Cordova or Capacitor
-
   if (this.platform.is('hybrid')) {
     // Read the file into base64 format
     const file = await Filesystem.readFile({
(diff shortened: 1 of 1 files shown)


amandaejohnston:
Thank you for the PR! `photo.path` is actually the correct variable to use here. You can see the details for it on the Capacitor docs: https://ionicframework.com/docs/native/camera#photo

Rather than change the variable, we should assert it -- `photo.path` should become `photo.path!` instead. This would keep consistency with the other frameworks ([React](https://ionicframework.com/docs/react/your-first-app/adding-mobile#platform-specific-logic), [Vue](https://ionicframework.com/docs/vue/your-first-app/adding-mobile#platform-specific-logic)), which already have this change.

Could you update your PR accordingly? Alternatively, I can put up a separate PR with this new change instead.

implicitdefcncdragneel:
Hi @amandaejohnston, I have updated the PR! accordingly, Could you please review the changes made?

amandaejohnston:
Thank you! I went ahead and removed the extra newline; otherwise this LGTM.

amandaejohnston (approved):

PR #2988, opened 2023-06-03
Update hardware-back-button.md to clarify usage for Capacitor apps.

nomadsiv:
The original text only points to the capacitor app plugin, but the event listener registering does not work as described here. The clarification added ought to help.

### DIFF
--- a/docs/developing/hardware-back-button.md
+++ b/docs/developing/hardware-back-button.md
@@ -22,7 +22,26 @@ The hardware back button refers to the physical back button on an Android device
 ## Hardware Back Button in Capacitor and Cordova
 
 :::note
-The `@capacitor/app` package must be installed in Capacitor apps to use the hardware back button.
+The [`@capacitor/app`](https://capacitorjs.com/docs/apis/app) package must be installed in Capacitor apps to use the hardware back button.
+
+Registering `ionBackButton` does not work in Capacitor apps.
+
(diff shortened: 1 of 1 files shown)


liamdebeasi (commented):
Thanks for the PR! I'm not sure I understand why this change is necessary. If the `@capacitor/app` plugin is installed then Capacitor will emit the `backbutton` event. If you are using Ionic Framework, we listen for `backbutton` and emit `ionBackButton` and dispatch it on the document: https://github.com/ionic-team/ionic-framework/blob/b8113cbd794faff08bc8786bd60aedcdf8403d0c/core/src/utils/hardware-back-button.ts#LL37C14-L37C14

liamdebeasi:
Hey there,

I have not heard back, so I am going to close this. If you would still like to get this merged feel free to reply. Thanks!

nomadsiv:
Hi Liam,
Apologies, for the delayed response. I was a bit unsure how to
respond/explain. But let me try:

>If the @capacitor/app plugin is installed then Capacitor will emit the
backbutton event.

When I looked at this documentation this is not mentioned there, i.e., to
listen for the backbutton event. there is the Note at the top saying if
using capacitor, the @capacitor/app plugin is needed. Then the rest of the
page only talks about the ionic part.

My PR is meant to clarify the backbutton event etc. for capacitor.

I still think it adds value, for what it's worth :)

Let me know if this makes sense.

Regards,
Madhu

PR #2989, opened 2023-06-05
chore(deps): update react and vue playground dependencies

sean-perkins:
Updates the component playground dependencies for React and Vue.


liamdebeasi (approved):
Tested the React and Vue playgrounds and verified they are using Ionic v7.0.10

PR #3001, opened 2023-06-20
4409 docs

liamdebeasi:


brandyscarney (approved):
Same case here where the JavaScript StackBlitz example isn't floating the label while the other frameworks are but if this is caused by npm / StackBlitz then LGTM. 👍

PR #3012, opened 2023-06-29
docs(react): clarify IonTabs behavior with IonPage

liamdebeasi:
As noted in https://github.com/ionic-team/ionic-framework/issues/27705, it turns out that IonTabs in Ionic React automatically renders IonPage for you, so adding duplicate IonPage instances can cause rendering issues. This behavior is not aligned with Vue, so we should look into making this consistent in the future. For now, I've updated the docs to make this clear.


mapsandapps (approved):
LGTM. I didn't see any other instances of this other than vue ones.

PR #3014, opened 2023-07-03
chore(docusaurus): update to latest

thetaPC:
Issue URL: N/A

Docusaurus was on an old version.

## What is the new behavior?

Node 18 can now be used to run the project without failing.

Docusaurus has been updated to the latest version.

- `docusaurus-theme-classic/` was [removed](https://github.com/ionic-team/ionic-docs/pull/3014/commits/fb2e41945dc74260b1a03fe6be4879d2992ea7df) in order to rely on the official [`docusaurus-theme-classic` theme](https://docusaurus.io/docs/api/themes/@docusaurus/theme-classic) via the official [`docusaurus-preset-classic` preset](https://docusaurus.io/docs/using-plugins#docusauruspreset-classic).

`docusaurus-theme-classic/` was originally added in order to create custom navbar item types ([cta](https://github.com/ionic-team/ionic-docs/blob/03022cbb5c0b806d8a5879854752fe7abfc4f142/docusaurus.config.js#L101), [icon link](https://github.com/ionic-team/ionic-docs/blob/03022cbb5c0b806d8a5879854752fe7abfc4f142/docusaurus.config.js#L201), and [separator](https://github.com/ionic-team/ionic-docs/blob/03022cbb5c0b806d8a5879854752fe7abfc4f142/docusaurus.config.js#L183)). However, this approach would require a need to manually check differences between the original and the custom. 

During the update, I decided to remove it in order to make it easier to upgrade **and** because the custom theme would lead to errors after updating. There are currently no documents that provide a clear way to keep the old approach. There is a current feature request to allow custom navbar items.

In the meantime, the cta and icon links are now being rendered through an [`html` type](https://docusaurus.io/docs/api/themes/configuration#navbar-with-custom-html) or [`doc` type](https://docusaurus.io/docs/api/themes/configuration#navbar-doc-link). The functionality is the same. However, the styling had to be updated to prevent any visual changes between the before and after update.

The icon links are set up the same way that the official Docusaurus sets its social links. Instead of passing the SVG through a component, it's being set up through CSS. In order for the SVG to change colors based on light/dark mode, the SVG is being passed through a mask.

- [`docusaurus-preset-classic` preset](https://docusaurus.io/docs/using-plugins#docusauruspreset-classic) is being used as the project's theme

The preset already ships multiple packages like `@docusaurus/plugin-content-docs`. I uninstalled the shipped packages since they are not required to be installed as dependencies. It will also relay each option entry to the respective plugin/theme through the `docusaurus.config.js` presets.

- `@ionic-internal/ionic-ds` is still being used. This is great because we won't need to do anything extra when a new update is released.

- `isDarkTheme` is deprecated. It's been updated to the latest code (`colorMode`)

- CSS selectors were updated since the HTML structure has changed

- Theme components are up to date with the official [`docusaurus-theme-classic` theme](https://docusaurus.io/docs/api/themes/@docusaurus/theme-classic) components.

There are some components that were [swizzled (ejected)](https://docusaurus.io/docs/swizzling#ejecting). This allows us to override the original component's file. There are certain components that take advantage of this to display a different view. For example, the navbar is originally across the entire screen. However, Ionic only wants it inside the docs page.

Swizzling is considered unsafe since the developers would need to maintain it vs the original theme. For example, we have swizzled `Icon/Edit` because we want to show a GitHub logo instead of a pencil icon. We need make sure that our file syncs up to the original as much as possible. Otherwise, updates can be a pain.

So in short, we want to swizzle as little as possible and follow the original theme's file structure as much as possible. When we do swizzle, comment what was done to make maintenance easier.

- Added a swizzled `LocaleDropdownNavbarItem` in order to visually hide the language text.
- Added a swizzled `Icon/Edit` to change the svg
- Added a swizzled `Icon/Language` to change the svg

- [ ] Yes
- [x] No

## Other information

I also applied the same changes to a branched version of [translation/jp](https://github.com/ionic-team/ionic-docs/tree/translation/jp). The changes were successful and can be viewed at PR https://github.com/ionic-team/ionic-docs/pull/3054.


package.json
```diff
@@ -2,6 +2,9 @@
   "name": "ionic-docs",
   "version": "0.0.0",
   "private": true,
+  "engines": {
+    "node": ">=18.0.0"
+  },
```
thetaPC, line 7: This can be removed once the project settings on Vercel are updated to Node v18.

src/components/global/PlaygroundTabs/index.tsx
```diff
@@ -7,7 +7,8 @@
 
 import React, { useState, cloneElement, isValidElement, type ReactElement, useRef, createRef, useEffect } from 'react';
 import useIsBrowser from '@docusaurus/useIsBrowser';
-import { useScrollPositionBlocker, duplicates } from '@docusaurus/theme-common';
+import { duplicates } from '@docusaurus/theme-common';
+import { useScrollPositionBlocker } from '@docusaurus/theme-common/internal';
```
thetaPC, line 11: `useScrollPositionBlocker` is needed to [maintain page position](https://github.com/facebook/docusaurus/pull/5618) for clicked grouped tabs. However, Docusaurus [does not encourage](https://docusaurus.io/community/release-process#theming-public-api) for users to import anything from ` '@docusaurus/theme-common/internal'` unless it's a swizzled component.

I propose to leave this for now for the following reasons:
Keep in mind PlaygroundTabs is a swizzled component from Docusaurus's Tabs.

The PlaygroundTabs currently fails web accessibility because it has buttons ([left arrow](https://github.com/ionic-team/ionic-docs/blob/71e57ed60f06be63c883963deb54af0c751278ae/src/components/global/PlaygroundTabs/index.tsx#L143-L166) and [right arrow](https://github.com/ionic-team/ionic-docs/blob/71e57ed60f06be63c883963deb54af0c751278ae/src/components/global/PlaygroundTabs/index.tsx#L190-L213)) inside a list (`<ul>`). A list can only have `li` as their children. In order to fix this accessibility issue, we should try to find a way to move the buttons outside of the tabs. If this is possible, then we won't need to use Docusaurus' Tabs logic and we can replace it with:

```tsx
// src/components/global/PlaygroundTabs/index.tsx
<div>
   <LeftButton />
   <Tabs className="unique-tabs">
     <TabItem value="X">src/app/example.component.ts</TabItem>
     <TabItem value="X">src/app/example.component.html</TabItem>
     <TabItem value="X">src/app/example.component.css</TabItem>
   </Tabs>
   <RightArrow />
<div>
``` 

Thoughts? @liamdebeasi @sean-perkins

src/styles/components/_navbar.scss
```diff
@@ -106,6 +111,39 @@ html[data-theme='dark'] {
         }
       }
 
+      .icon-link {
+        &-mask {
+          &::before {
+            content: '';
+            width: 20px;
```
sean-perkins, line 118: I changed this to `19px` in the dev tools and it reproduces the icon sizing closer to production. 

The challenge that we have is in production the sizes are:
- Twitter icon: 19px
- Discord icon: 20px
- Github icon: 17px

So applying a one size fits all to them causes slight stretching compared to production.

sean-perkins:
https://github.com/ionic-team/ionic-docs/assets/13732623/4ec1be85-92c7-466b-a631-d4ad6cdf1490

I'm getting horizontal overflow in the preview link on a Macbook Air screen.

thetaPC:
> 

> Kapture.2023-08-08.at.11.04.28.mp4
> 
> I'm getting horizontal overflow in the preview link on a Macbook Air screen.

The navbar is longer because the language dropdown now shows the language. In this case, "English" is in the navbar.

Here are some options:
1. We shrink the font size of the navbar
2. We swizzle the language anchor and visually hide the language text
3. Hide the text via css but it might not be web acccessible

I'm leaning towards option 2 but that's another component that we would have to maintain. Option 1 would be an easy solution but the text is already small that any decrease might make it hard for users to read.

sean-perkins:
I believe we are doing a combination of option 2 & 3 today (or the structure for the menu has changed).

In the production docs we are setting the icon for the language menu with a `::before` selector and hiding the `span` which includes the SVG and text content. 

I'm open to either approach.

thetaPC:
> I believe we are doing a combination of option 2 & 3 today (or the structure for the menu has changed).
> 
> In the production docs we are setting the icon for the language menu with a `::before` selector and hiding the `span` which includes the SVG and text content.
> 
> I'm open to either approach.

No, the current docs site is only using option 3. The menu structure has changed since then. For example, the anchor wrapped the text and the svg in a container at the time that the site branched off. 
Screenshot 2023-08-09 at 9 47 04 AM
The site has been able to hide the text by applying the style via span.

However, the container (span) is no longer available in the latest Docusaurus.
Screenshot 2023-08-09 at 9 47 59 AM
I've added the icon via `::before` (for the purpose of the screenshot example) but we can't target only the text and the svg. If we tried hiding it via CSS then it would also hide the `::before`.

That's why I'm thinking that swizzling would be a better solution. We can add a container and hide it visually but leave it available for screen readers.

sean-perkins:
Sounds good to me, let's swizzle it! 🚀

jaredcbaum (approved):

jaredcbaum:
Not a huge deal, but I'm seeing some navbar layout weirdness on initial page load. The icons are initially laid out vertically and then quickly switch to the correct horizontal layout.

src/styles/components/_doc-sidebar.scss
```diff
@@ -28,7 +28,7 @@ html[data-theme='dark'] {
 
 //overrides
 #__docusaurus {
-  [class^='docSidebarContainer'] {
+  aside > div {
```
liamdebeasi, line 31: The sidebar border is incorrect

| main | branch |
| - | - |
| image | image |

Should be using `var(--sidebar-border-c);` instead of `var(--ifm-toc-border-color);`. This should also be using `border-inline-end` instead of `border-right` for RTL.

This selector change may be to blame here. The correct border is applied to the div inside of `aside`, but the `aside` itself also has a border.

CONTRIBUTING.md
```diff
@@ -23,12 +23,10 @@ Thanks for your interest in contributing to Ionic's documentation! :tada: Check
 In order to run the documentation locally, install the dependencies and run the development server:
```
liamdebeasi, line 23: The Command+K icon in the search field has the wrong ratio:


| main | branch |
| - | - |
| image | image |

The command symbol should roughly be the same size as the K

CONTRIBUTING.md
```diff
@@ -23,12 +23,10 @@ Thanks for your interest in contributing to Ionic's documentation! :tada: Check
...
 
 ```sh
-$ npm install --legacy-peer-deps
-$ npm start
+npm install
```
liamdebeasi, line 26: The translation icon in the top menu bar appears smaller/thinner than main

| main | branch |
| - | - |
| image | image |

CONTRIBUTING.md
```diff
@@ -23,12 +23,10 @@ Thanks for your interest in contributing to Ionic's documentation! :tada: Check
...
+npm start
```
liamdebeasi, line 27: The margins seem larger now:

(top is branch, bottom is main. I added red guidelines for emphasis)

image

Also looks like the default arrow icon changed too.

CONTRIBUTING.md
liamdebeasi, line 26: The sidebars are much smaller than they are in main

| main | branch |
| - | - |
| image | image |

CONTRIBUTING.md
```diff
@@ -23,12 +23,10 @@ Thanks for your interest in contributing to Ionic's documentation! :tada: Check
...
 ```
```
liamdebeasi, line 28: The "Note" fields are no longer yellow. Is this an intentional change?

Example:

branch: https://ionic-docs-git-docusaurus-update-ionic1.vercel.app/docs/cli#help
main: https://ionicframework.com/docs/cli#help

CONTRIBUTING.md
```diff
@@ -23,12 +23,10 @@ Thanks for your interest in contributing to Ionic's documentation! :tada: Check
...
 
-> **Note**: certain versions of npm (5-8) and Node.js (10-16) are required to run certain scripts.
```
liamdebeasi, line 30: The Ionic logo doesn't show up in the header on the "Native" page anymore: https://ionic-docs-git-docusaurus-update-ionic1.vercel.app/docs/native

CONTRIBUTING.md
```diff
@@ -23,12 +23,10 @@ Thanks for your interest in contributing to Ionic's documentation! :tada: Check
...
-
```
liamdebeasi, line 31: The hover state looks a bit odd since there's no padding on the left: 
image


This looks like a new behavior in the latest version of Docusuarus

CONTRIBUTING.md
```diff
@@ -23,12 +23,10 @@ Thanks for your interest in contributing to Ionic's documentation! :tada: Check
...
 ---
```
liamdebeasi, line 30: I can only click on the sidebar item text to expand/collapse despite the entire line being highlighted.

First part of the video shows `main` second part shows branch.

https://github.com/ionic-team/ionic-docs/assets/2721089/b845d67c-0373-426f-8407-b340ec2a867a

CONTRIBUTING.md
```diff
@@ -23,12 +23,10 @@ Thanks for your interest in contributing to Ionic's documentation! :tada: Check
...
 
 ## Using VS Code on Windows
```
liamdebeasi, line 32: The page on mobile goes outside the viewport due to the header:

| main | branch |
| - | - |
| image | image |

liamdebeasi (changes requested):
I didn't see any major issues, just a bunch of small layout bugs. Great work so far!

CONTRIBUTING.md
thetaPC: Managed to fix it. I did notice that some words were getting cutoff behind the scroll so the width was updated. Expect the new code to show a few pixels wider.

CONTRIBUTING.md
thetaPC: I had to hide the default icon that Docusaurus renders and [import a new icon via CSS](https://github.com/ionic-team/ionic-docs/pull/3014/files#diff-035ec64dbe3b95d97cbc279267dee0159d93b22d15e30970d9c53c16ba799a56R296-R298). The new icon has a bolder setup. This feels a bit weird. Do we want to continue with this approach or use the default?

CONTRIBUTING.md
thetaPC: Can you check if this is still an issue? I'm seeing the logo.

CONTRIBUTING.md
thetaPC: Managed to fix it. I did notice that some words were getting cutoff behind the scroll so the width was updated. Expect the new code to show a few pixels wider.

CONTRIBUTING.md
thetaPC: I've removed the hover styling to reflect main. Since it's no longer showing highlighted, would you still expect to click anywhere on the line or would you expect to click where the cursor is a pointer?

CONTRIBUTING.md
liamdebeasi: Logo looks correct now, though I see some duplicate icons: 
image

CONTRIBUTING.md
liamdebeasi: I would expect the behavior to match what's on the production website. Note that you can click anywhere on the child links.

CONTRIBUTING.md
thetaPC: The thinner icon is the default that Docusaurus provides. Main also opts out of using the default and uses the import via CSS. The reason seems to be due to wanting a different icon at that time. Here's a screenshot of the imported icon and the default icon on main side by side.

Screenshot 2023-09-06 at 8 13 51 AM (main on left, Docusaurus on right)

Fast forward to today, Docusaurus has changed it's icon. It matches the one shown in main but the SVG is thinner.

Screenshot 2023-09-06 at 8 13 56 AM  (main on left, Docusaurus on right)

I tried to match the default to main but couldn't get it so I opted to import main's icon via CSS. Realizing it now, I might be able to import the language icon from Ionicons. It might match the thickness. I'll give this a shot.

CONTRIBUTING.md
thetaPC: If you do see some differences. It's due to Docusaurus changing how they calculate the columns.

docusaurus.config.js
```diff
@@ -25,6 +25,53 @@ module.exports = {
   favicon: 'img/meta/favicon-96x96.png',
   organizationName: 'ionic-team',
   projectName: 'ionic-docs',
+  presets: [
```
liamdebeasi, line 28: The main body alignment is incorrect on smaller viewports:

| production | branch |
| - | - |
| image | image |

docusaurus.config.js
```diff
@@ -25,6 +25,53 @@ module.exports = {
...
+    [
+      '@docusaurus/preset-classic',
```
liamdebeasi, line 30: There seems to be an additional gutter on the right (notice the right margin is larger than the left margin in the `branch` image)

Looks to be a similar issue on an individual component's docs page too


| production | branch |
| - | - |
| image | image |

docusaurus.config.js
```diff
@@ -25,6 +25,53 @@ module.exports = {
...
+      /** @type {import('@docusaurus/preset-classic').Options} */
+      {
```
liamdebeasi, line 32: The command looks really big when compared with the "K"

| production | branch |
| - | - |
| image | image |

docusaurus.config.js
```diff
@@ -25,6 +25,53 @@ module.exports = {
...
+        // Will be passed to @docusaurus/plugin-content-docs (false to disable)
+        docs: {
```
liamdebeasi, line 34: The margin between the accordion header and the icon is incorrect in RTL (LTR is correct, so maybe we need to use `margin-inline` instead): 

| production | branch |
| - | - | 
| image | image |

liamdebeasi (changes requested):
These are the final things I noticed! I'll do another round of review just to make sure fixing the remaining issues didn't break anything else, but this is looking great!

docusaurus.config.js
thetaPC: Do you remember what width and browser you saw this? I'm not able to replicate it.

liamdebeasi: Sorry! I had my screenshots backwards. It looks like the problem exists in production, but your branch fixes it 🎉

src/components/global/PlaygroundTabs/index.tsx
liamdebeasi: Can we make a bug ticket with this information (include what the desired behavior is and ideas for how to implement)? We can prioritize it so it gets fixed quickly.

src/styles/components/_doc-item.scss
```diff
@@ -38,19 +38,27 @@
...
+            @media (max-width: 1364px) {
+              display: none;
+            }
+          }
+
+          // Table of contents
+          &.col--3 {
+            @media (max-width: 996px) {
+              display: none;
+            }
```
liamdebeasi, line 60: Can we add comments describing what this does and why it is necessary?

src/styles/components/_doc-sidebar.scss
```diff
@@ -100,6 +109,9 @@ html[data-theme='dark'] {
           padding-inline-start: calc(var(--dropdown-icon-width) + var(--dropdown-icon-gap));
         }
       }
+      &-collapsible {
+        display: block;
+      }
```
liamdebeasi, line 114: Can we add comments describing what this does?

src/styles/components/_doc-sidebar.scss
```diff
@@ -147,6 +159,8 @@ html[data-theme='dark'] {
     flex-direction: row-reverse;
     justify-content: flex-end;
 
+    flex: initial;
```
liamdebeasi, line 162: Can we add comments describing what this does?

liamdebeasi (approved):
A few code commenting changes, but otherwise this looks great! Let's sync prior to merging this so we can make the rest of the team aware of the upcoming changes. Great work!

src/components/global/PlaygroundTabs/index.tsx
thetaPC: Created the ticket. 

Adding the following comment for records:
I do want to point out that my proposed solution might not completely remove the need to use `useScrollPositionBlocker` if this [PR](https://github.com/ionic-team/ionic-docs/pull/3169/files#diff-448a7530c1f7d78c814f204783575fb780358f357f6eee5a69f50ff4b7f0f0b6R16) is approved.

thetaPC: Resolving since we determined to continue using it until an issue arises.

src/styles/components/_doc-sidebar.scss
thetaPC: Removed since it seems to be no longer needed.

src/styles/components/_doc-item.scss
thetaPC: Removed the `col--3` since I noticed that it's being removed completely under `max-width` so it doesn't need to be hidden via CSS. I did add a comment for `col--4` since it can be removed by JS.

liamdebeasi (approved):
Great work!

docusaurus.config.js
liamdebeasi: Hmm this still doesn't look quite right: 
image

The command logo and the letter seem to be different sizes/misaligned.

liamdebeasi (changes requested):
Requesting changes just in case you didn't see my search icon follow up comment (everything else looks good)

docusaurus.config.js
thetaPC: does it still look weird? I made changes last Friday: https://github.com/ionic-team/ionic-docs/pull/3014/commits/72a0ccf06021475f1b43a8f764666e92b1dab9c4

liamdebeasi: It looks a lot better now! 
image

liamdebeasi (approved):

PR #3015, opened 2023-07-05
docs(list): use dark mode aware color

liamdebeasi:


liamdebeasi:
I added co-author credit when enabling auto-merge.

PR #3016, opened 2023-07-05
chore: enable renovate for component playground deps

sean-perkins:
Enables the Renovate bot for managing the dependencies used by the component playground stackblitz project templates.

This PR is in favor of: https://github.com/ionic-team/ionic-docs/pull/3017

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -5,14 +5,10 @@
     {
       "packageNames": ["*"],
       "directoryNames": [
-        "static/code/stackblitz/v6/angular",
-        "static/code/stackblitz/v7/angular",
-        "static/code/stackblitz/v6/html",
-        "static/code/stackblitz/v7/html",
-        "static/code/stackblitz/v6/vue",
-        "static/code/stackblitz/v7/vue",
-        "static/code/stackblitz/v6/react",
-        "static/code/stackblitz/v7/react"
+        "static/code/stackblitz/*/angular",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": ["config:base", ":github>ionic-team/ionic-docs>"],
+  "packageRules": [
+    {
+      "packageNames": ["*"],
+      "directoryNames": [
```
liamdebeasi, line 7: This is pretty cool! Will this update any and all dependencies, or just certain ones?

liamdebeasi: Also, will these deps be updated in groups or will we get 1 PR per dep?

sean-perkins: This will update any and all dependencies within the Stackblitz directories. At the moment, it would create a separate PR per dependency.

We could enable [`groupName`](https://docs.renovatebot.com/noise-reduction/#package-grouping) to batch similar dependencies (like `@angular/*`, `@ionic/`, etc.). 

I didn't do it to start, but I would be open to those 2 cases if you agree.

liamdebeasi: It's not required, but I think it would be nice especially when the Angular deps get updated

liamdebeasi (approved):
Thanks for doing this!

PR #3030, opened 2023-07-11
chore(renovate): limit include paths

sean-perkins:
Based on: https://github.com/ionic-team/ionic-docs/issues/3029, Renovate is still detecting `package.json` files at the root and within the docusaurus plugin. 

This PR limits the `includePaths` to only the Stackblitz directories.

Additionally this PR tries to limit the allowed version bumps for the v6 folder paths to not attempt to bump to Ionic v7.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -11,7 +11,7 @@
       "groupName": "angular"
     },
     {
-      "matchPackageNames": ["^@ionic/"],
+      "matchPackagePatterns": ["^@ionic/"],
       "allowedVersions": "^6.0.0",
       "matchFileNames": ["static/code/stackblitz/v6/**/package.json"]
     }
--- a/renovate.json
+++ b/renovate.json
@@ -9,6 +9,11 @@
     {
       "matchPackagePatterns": ["^@angular/"],
       "groupName": "angular"
+    },
+    {
+      "matchPackageNames": ["^@ionic/"],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -9,12 +9,17 @@
     {
       "matchPackagePatterns": ["^@angular/"],
       "groupName": "angular"
+    },
+    {
+      "matchPackageNames": ["^@ionic/"],
```
brandyscarney, line 14: I'm not real familiar with renovate but [`matchPackageNames`](https://docs.renovatebot.com/configuration-options/#matchpackagenames) says:

> `matchPackageNames`: Use this field if you want to have one or more exact name matches in your package rule. See also `excludePackageNames`. Example: `"matchPackageNames": ["angular"],`

Should this also be [`matchPackagePatterns `](https://docs.renovatebot.com/configuration-options/#matchpackagepatterns)? 

> `matchPackagePatterns`: Use this field if you want to have one or more package names patterns in your package rule. See also `excludePackagePatterns`. Example: `"matchPackagePatterns": ["^angular"],`

sean-perkins: Yes! Good catch, updated here: https://github.com/ionic-team/ionic-docs/pull/3030/commits/abb75395213714abd2e23626901925e29ed6502f

brandyscarney (approved):

PR #3035, opened 2023-07-12
docs(animations): add playground for before and after hooks

brandyscarney:
This is an alternative implementation of https://github.com/ionic-team/ionic-docs/pull/1702. It replaces the "Before and After Hooks" samples with a playground. 

I copied the existing animation playgrounds including the "play", "pause" and "stop" button (see https://github.com/ionic-team/ionic-docs/pull/3026). I changed the styles from the existing docs to use a `filter` and `box-shadow` instead of opacity, making it easier to see the styles on dark mode. 

:warning: Give co-author credit to [dillionmegida](https://github.com/dillionmegida) :warning:

Preview: https://ionic-docs-git-fw-3378-ionic1.vercel.app/docs/utilities/animations#before-and-after-hooks

### DIFF
--- a/docs/utilities/animations.md
+++ b/docs/utilities/animations.md
@@ -326,118 +326,13 @@ import Group from '@site/static/usage/v7/animations/group/index.md';
 
 Ionic Animations provides hooks that let you alter an element before an animation runs and after an animation completes. These hooks can be used to perform DOM reads and writes as well as add or remove classes and inline styles.
 
-### Usage
-
-````mdx-code-block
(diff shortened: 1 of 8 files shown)


sean-perkins:
Since the playground component has a built in reset action, I'm unsure if we need a similar action within the demo example. 

I agree the opacity style is not visible on dark mode. Can we use a `filter` instead and maybe blur or use another filter style?

sean-perkins (approved):
I like the changes 👍

PR #3037, opened 2023-07-12
style(animations): buttons in StackBlitz demos do not need id

brandyscarney:
While implementing https://github.com/ionic-team/ionic-docs/pull/3035 I noticed in the other examples the `id` was defined on the buttons for the Angular, Javascript, React, and Vue examples. These do not need an `id` since they call click events directly, unlike the demo file.


liamdebeasi (approved):
Thank you!

thetaPC:
Thank you!!!

PR #3060, opened 2023-07-26
feat(playground): add prop to display console messages from demo

amandaejohnston:
Adds a new `showConsole` prop to the playground, which displays a console UI underneath the iframe demo. Any calls to `console.log`, `console.warn`, or `console.error` made from inside the iframe will be mirrored in the console UI.

I also converted the

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -265,18 +265,20 @@ export default function Playground({
   }, [renderIframes]);
 
   useEffect(() => {
-    if (frameiOS.current) {
-      frameiOS.current.contentWindow.addEventListener('console', (ev: CustomEvent) => {
-        setiOSConsoleItems((oldConsoleItems) => [...oldConsoleItems, ev.detail]);
-        consoleRef.current.scrollTo(0, consoleRef.current.scrollHeight);
-      });
-    }
+    if (showConsole) {
+      if (frameiOS.current) {
+        frameiOS.current.contentWindow.addEventListener('console', (ev: CustomEvent) => {
+          setiOSConsoleItems((oldConsoleItems) => [...oldConsoleItems, ev.detail]);
+          consoleRef.current.scrollTo(0, consoleRef.current.scrollHeight);
+        });
+      }
 
-    if (frameMD.current) {
-      frameMD.current.contentWindow.addEventListener('console', (ev: CustomEvent) => {
-        setMDConsoleItems((oldConsoleItems) => [...oldConsoleItems, ev.detail]);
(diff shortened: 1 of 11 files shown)


brandyscarney:
This looks great! Should we indicate somewhere that this is a console? Quick mockup:

Group 1

amandaejohnston:
I like that idea a lot :eyes: Added in https://github.com/ionic-team/ionic-docs/pull/3060/commits/93a36ba85470023a4e699044562eb9b087b02dde.

src/components/global/Playground/playground.css
```diff
@@ -213,6 +241,79 @@
   }
 }
 
+.playground__console {
```
sean-perkins, line 244: Everything works great! I think we should reduce the font size of the console however. It is larger than any other UI within the playground (including the code snippets). 

The code's font size is `.813rem`. Perhaps we start there to align visually?

src/components/global/Playground/index.tsx
```diff
@@ -444,11 +468,31 @@ export default function Playground({
...
   }
 
+  function renderConsole() {
+    return (
+      <div className="playground__console">
+        <div className="playground__console-header">
+          <code>Console</code>
+        </div>
+        <div className="playground__console-body" ref={consoleBodyRef}>
+          {(ionicMode === Mode.iOS ? iosConsoleItems : mdConsoleItems).map((consoleItem, i) => (
```
sean-perkins, line 478: This isn't part of the design document, so free feel to push back and I can log a separate task to track this.

When the console UI is initially blank, it is a little unclear to the user what to expect. I think a placeholder message could be a nice improvement over this. Something simple like:
> Console messages will appear here when logged from the example above.

sean-perkins (changes requested):
Only blocking comment is the font size.

src/components/global/Playground/index.tsx
amandaejohnston: I like that idea 👀 Will go ahead and add it since it should be simple.

brandyscarney:
This might be outside of the scope of this, but do we want to clear the console when the user presses the "reset demo" button? or have a separate clear button?

amandaejohnston:
That actually made me notice a bug where if you reset the demos, the console stops getting updated. Will address both :eyes:

amandaejohnston:
Alright, resetting the demo now empties the console, and the console continues to work 👍 @sean-perkins and/or @brandyscarney This is ready for review again.

sean-perkins (approved):
Fantastic work, this is great 🌟

PR #3063, opened 2023-07-27
Update adding-ionic-react-to-an-existing-react-project.md

AdamRaf:
added yarn command and line space for a prettier UI


sean-perkins:
Hello @AdamRaf thanks for submitting this pull request!

We prioritize showing `npm `commands across our documentation. It is a trend in the ecosytem that developers using an alternative package manager (yarn, pnpm, bun, etc.) should be able to navigate the differences for commands (such as installing a package). Otherwise it would be extremely tedious for projects to show instructions for every alternative tool that exists. 

As such we are going to close out this pull request. 

I appreciate you taking the time to suggest an improvement to the docs!

PR #3064, opened 2023-07-27
docs(api): add accessibility section for new htmlAttributes property

brandyscarney:
This PR adds the following to different API documentation:

[Action Sheet](https://ionic-docs-git-fw-4240-docs-ionic1.vercel.app/docs/api/action-sheet#accessibility):
- Puts all of the `Accessibility` text under subheading `Screen Readers` to match `Toast`
- Adds a section under `Accessibility` that mentions buttons with only an icon should pass an `aria-label` property to `htmlAttributes` in order to be described by screen readers
- Adds code examples for passing to `htmlAttributes` both on the Action Sheet and its buttons
- Adds the `htmlAttributes` type to the `ActionSheetButton` interface
- Adds the `id` type to the `ActionSheetButton` interface - I noticed this missing but it was actually [added in v6](https://github.com/ionic-team/ionic-framework/commit/9e24a0b49357a3a39ca89f026ff23271a365d935) so I can add this to `v6` also if desired or backport this fix off of `main`.

[Alert](https://ionic-docs-git-fw-4240-docs-ionic1.vercel.app/docs/api/alert#accessibility):
- Puts all of the `Accessibility` text under subheading `Screen Readers` to match `Toast`
- Adds a section under `Accessibility` that mentions the description of a button can be changed by passing the `aria-label` property to `htmlAttributes` for screen readers
  - Note: this does not mention icons because the `AlertButton` interface does not accept an `icon` property
- Adds code examples for passing to `htmlAttributes` both on the Alert and its buttons
- Adds the `htmlAttributes` type to the `AlertButton` interface
- Adds the `id` type to the `AlertButton` interface - I noticed this missing but it was actually [added in v6](https://github.com/ionic-team/ionic-framework/commit/9e24a0b49357a3a39ca89f026ff23271a365d935) so I can add this to `v6` also if desired or backport this fix off of `main`.
- Updates the `handler` type in the `AlertButton` interface - this was also [added in v6](https://github.com/ionic-team/ionic-framework/commit/8e4783c17298e8f7654590108430e80f22ed6a7a) so I can do the same as above if desired.

[Toast](https://ionic-docs-git-fw-4240-docs-ionic1.vercel.app/docs/api/toast#accessibility):
- Adds a section under `Accessibility` that mentions buttons with only an icon should pass an `aria-label` property to `htmlAttributes` in order to be described by screen readers
- Adds code examples for passing to `htmlAttributes` to the toast buttons
- Adds the `htmlAttributes` type to the `ActionSheetButton` interface
- Moves the `Accessibility` section above `Interfaces` to match the other API docs

----

This depends on the following PRs being merged:
- https://github.com/ionic-team/ionic-framework/pull/27863
- https://github.com/ionic-team/ionic-framework/pull/27862
- https://github.com/ionic-team/ionic-framework/pull/27855

### DIFF
--- a/docs/api/action-sheet.md
+++ b/docs/api/action-sheet.md
@@ -96,11 +96,41 @@ import CssCustomProperties from '@site/static/usage/v7/action-sheet/theming/css-
 
 ### Screen Readers
 
+Action Sheets set aria properties in order to be [accessible](../reference/glossary#a11y) to screen readers, but these properties can be overridden if they aren't descriptive enough or don't align with how the action sheet is being used in an app.
+
+#### Role
+
 Action Sheets are given a `role` of [`dialog`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role). In order to align with the ARIA spec, either the `aria-label` or `aria-labelledby` attribute must be set.
+
+#### Action Sheet Description
 
 It is strongly recommended that every Action Sheet have the `header` property defined, as Ionic will automatically set `aria-labelledby` to point to the header element. However, if you choose not to include a `header`, an alternative is to use the `htmlAttributes` property to provide a descriptive `aria-label` or set a custom `aria-labelledby` value.
 
-Buttons containing text will be read by a screen reader. If a button contains only an icon, a label should be assigned to the button by passing `aria-label` or `aria-labelledby` to the `htmlAttributes` property. The `aria-label` property accepts a text description and `aria-labelledby` accepts an element that contains the description of the button.
+```javascript
+actionSheetController.create({
+  htmlAttributes: {
+    'aria-label': 'action sheet dialog',
+  },
+});
+```
+
+#### Action Sheet Buttons Description
+
+Buttons containing text will be read by a screen reader. If a button contains only an icon, or a description other than the existing text is desired, a label should be assigned to the button by passing `aria-label` to the `htmlAttributes` property on the button.
+
+```javascript
+actionSheetController.create({
+  header: 'Header',
+  buttons: [
+    {
+      icon: 'close',
+      htmlAttributes: {
+        'aria-label': 'close',
+      },
+    },
+  ],
+});
+```
 
 ## Interfaces
 
--- a/docs/api/action-sheet.md
+++ b/docs/api/action-sheet.md
@@ -1,6 +1,9 @@
 ---
 title: "ion-action-sheet"
 ---
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
 import Props from '@ionic-internal/component-api/v7/action-sheet/props.md';
 import Events from '@ionic-internal/component-api/v7/action-sheet/events.md';
 import Methods from '@ionic-internal/component-api/v7/action-sheet/methods.md';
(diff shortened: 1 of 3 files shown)


docs/api/action-sheet.md
```diff
@@ -94,10 +94,14 @@ import CssCustomProperties from '@site/static/usage/v7/action-sheet/theming/css-
 
 ## Accessibility
 
+### Screen Readers
+
 Action Sheets are given a `role` of [`dialog`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role). In order to align with the ARIA spec, either the `aria-label` or `aria-labelledby` attribute must be set.
 
 It is strongly recommended that every Action Sheet have the `header` property defined, as Ionic will automatically set `aria-labelledby` to point to the header element. However, if you choose not to include a `header`, an alternative is to use the `htmlAttributes` property to provide a descriptive `aria-label` or set a custom `aria-labelledby` value.
 
+Buttons containing text will be read by a screen reader. If a button contains only an icon, a label should be assigned to the button by passing `aria-label` or `aria-labelledby` to the `htmlAttributes` property. The `aria-label` property accepts a text description and `aria-labelledby` accepts an element that contains the description of the button.
```
mapsandapps, line 103: I think we could perhaps make this slightly more clear that they need to set the `htmlAttributes` property on the button, not on the action sheet itself. The rest of this section refers to the action sheet as a whole, and the paragraph before this one even includes a reference to `htmlAttributes` on the action sheet. A code snippet would be nice, but more specific text would suffice.

mapsandapps (commented):
Just one thought. Everything else looks good. Good catch on adjusting the Alert text to be different from the others.

docs/api/toast.md
```diff
@@ -86,6 +86,50 @@ import ThemingPlayground from '@site/static/usage/v7/toast/theming/index.md';
...
+
+`ion-toast` has `role="status"` and `aria-live="polite"` set on the inner `.toast-content` element. This causes screen readers to only announce the toast message and header. Buttons and icons will not be announced.
+
+`aria-live` causes screen readers to announce the content of the toast when it is updated. However, since the attribute is set to `'polite'`, screen readers should not interrupt the current task.
+
+Since toasts are intended to be subtle notification, `aria-live` should never be set to `"assertive"`. If developers need to interrupt the user with an important message, we recommend using an [alert](./alert).
+
+#### Toast Buttons Description
+
+Buttons containing text will be read by a screen reader. If a button contains only an icon, or a description other than the existing text is desired, a label should be assigned to the button by passing `aria-label` to the `htmlAttributes` property on the button.
```
brandyscarney, line 109: > Buttons containing text will be read by a screen reader. 

Is this even accurate? The section above it says:

> `ion-toast` has `role="status"` and `aria-live="polite"` set on the inner `.toast-content` element. This causes screen readers to only announce the toast message and header. Buttons and icons will not be announced.

But the original issue this is for is for toast and states:

> Buttons in an ion-toast that are only an icon (eg close) cannot be made accessible as they do not have an aria-label property that can be set.

So I think I should update this to say that they won't be read unless `aria-label` is provided.

docs/api/action-sheet.md
brandyscarney: Separated them out into different subheadings and added code examples to each. Lmk if that's better!

docs/api/toast.md
mapsandapps: I think what happens is that it doesn't announce the text or aria-label when the toast appears on the screen, but it will announce it if the user navigates to the toast. I suppose the current text is a bit confusing, but I think it's all correct.

docs/api/action-sheet.md
mapsandapps: This is great! Thanks! 💯

mapsandapps (approved):

docs/api/toast.md
brandyscarney: Done: https://github.com/ionic-team/ionic-docs/pull/3064/commits/4229be3de4f1bb6719f8779024c31ec42754cd4f

Issue #2998, opened 2023-06-15
content: Dark Mode codepens should be replaced with full playgrounds


amandaejohnston:
### URL

https://ionicframework.com/docs/theming/dark-mode

### Issue Description

This page contains a couple of Codepen demos, which only showcase Ionic in vanilla JS. We should replace them (and perhaps some of the other code snippets on the page) with playgrounds/live demos like those found on the component docs ([example](https://ionicframework.com/docs/api/button#basic-usage)) so that we can easily include code for Angular, React, and Vue as well.

There is a PR at https://github.com/ionic-team/ionic-docs/pull/1648 that includes equivalent code for Angular and React.

PR #3067, opened 2023-08-02
docs(theming): migrate Dark Mode Codepens to playground examples

brandyscarney:
Resolves https://github.com/ionic-team/ionic-docs/issues/2998

Add Co-Authored-By

---

Preview link: https://ionic-docs-git-fw-4579-ionic1.vercel.app/docs/theming/dark-mode

---

Old Codepens for reference:
  - `Automatically Enable Dark Mode` - https://codepen.io/ionic/pen/jONzJpG
  - `Manually Toggle Dark Mode` - https://codepen.io/ionic/pen/zYOpQLj

---

This PR:

1. Migrates the Dark Mode Codepens to Playgrounds
    - While migrating I ran into an issue using the modern syntax with range [missing stacked label placement](https://github.com/ionic-team/ionic-framework/issues/27229)
    -  Due to this, I decided to update the demo to match the iOS settings app. As a result I updated the default dark theme in the demos to make the inset lists stand out more.
2. Updates the content in the Dark Mode guide to remove outdated sections & clarifies some of the existing sections
3. Updates the recommended `Ionic Dark Theme` to use what the starters provide
    - The old recommendation used the `body.dark` class but the new recommendation is to use the `prefers-color-scheme` media query.
    - When these were written the [browser support for prefers-color-scheme](https://caniuse.com/prefers-color-scheme) wasn't great so we recommended the fallback.

---

The first demo under `Automatically Enable Dark Mode` showcases toggling the dark theme by updating your system preferences. I added a note that it will prioritize the site's preferences over the system preferences if these are not the same, but toggling your system preferences will get them in sync. 

The second demo under `Manually Toggle Dark Mode` is the same as the previous demo but it adds on an `ion-toggle` which can be used to change the theme manually, regardless of system settings.

---

The JavaScript example has to have the entire `index.html` passed in order to update the `variables.css` file. This requires us to include the `<head>` and `config` code in order for the mode switching to work. This is rather ugly to provide to the users. I have created [FW-4886](https://ionic-cloud.atlassian.net/browse/FW-4886) to come up with a better way to achieve this. 

[FW-4886]: https://ionic-cloud.atlassian.net/browse/FW-4886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sean-perkins (approved):
Looks great 🚀

PR #3081, opened 2023-08-10
fix(): alphabetize sidebar

mapsandapps:
Put accordion before action sheet and icons before infinite scroll.

Screenshot 2023-08-10 at 2 22 07 PM

See https://github.com/ionic-team/ionic-docs/pull/3082 for similar change for Japanese docs.

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -238,15 +238,15 @@ module.exports = {
     },
     {
       type: 'category',
-      label: 'Action Sheet',
-      collapsed: false,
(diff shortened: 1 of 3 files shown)


brandyscarney (changes requested):
We should probably also make these changes for the [v5 and v6 sidebars](https://github.com/ionic-team/ionic-docs/tree/main/versioned_sidebars).

brandyscarney (approved):
Looks good!

PR #3084, opened 2023-08-13
fix(vue): fix bug with key type

fc92:
Avoid :
Type '{ filepath: string; webviewPath?: string | undefined; }' is not assignable to type 'string | number | symbol | undefined'.ts(2322) runtime-core.d.ts(914, 5): The expected type comes from property 'key' which is declared here on type '{ readonly push?: string | undefined; readonly modelValue?: string | number | boolean | undefined; readonly offset?: string | undefined; readonly offsetLg?: string | undefined; ... 32 more ...; style?: unknown; } & Record<...>' (property) key?: string | number | symbol | undefined


mapsandapps (approved):
Thanks for your contribution! This is a useful fix and matches the docs for React.

PR #3086, opened 2023-08-14
docs(item): remove old demo

liamdebeasi:
This demo showed how to use the legacy form syntax to change the highlight on `ion-item`. The highlight was added to `ion-input` and `ion-textarea` directly, so this playground no longer showcases our best practices. See https://github.com/ionic-team/ionic-framework/issues/27985

This PR removes references to the playground. However, the source is not removed as that will cause the JP docs to fail. (see: https://github.com/ionic-team/ionic-docs/pull/3004).

### DIFF
--- a/docs/api/item.md
+++ b/docs/api/item.md
@@ -121,18 +121,6 @@ import CSSProps from '@site/static/usage/v7/item/theming/css-properties/index.md
 
 <CSSProps />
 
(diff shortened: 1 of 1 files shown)


liamdebeasi:
This content has since been removed, so the PR is not needed anymore.

PR #3088, opened 2023-08-15
fix(): alphabetize sidebar

mapsandapps:
Reimplements #3081 which was reverted by #3087


liamdebeasi:
Closing for now. If the team would still like to work on this please bring it up internally!

PR #3098, opened 2023-08-24
feat(generation): generate playgrounds for guides

mapsandapps:
See FW-4717 and [design doc update](https://github.com/ionic-team/ionic-framework-design-documents/pull/149) for extra context.

Previously, you couldn't create a playground for animations, gestures, etc., without using some workarounds. Now the prompts ask if you're creating a playground for a component or not and adjust the other prompt questions accordingly.

⚠️ I also did a little refactoring & renaming to make the code more clear. Walking through the commits one-by-one may help for review.

I included an `ion-card` in the templates, since the animation playgrounds tend to have those.

### DIFF
--- a/_templates/playground/new/index.js
+++ b/_templates/playground/new/index.js
@@ -17,6 +17,7 @@ module.exports = {
       .then((answers) => {
         return inquirer
           .prompt([
+            // ask a different question for components vs. other playgrounds
             answers.is_component
               ? {
                   type: 'input',
--- a/_templates/playground/new/index.js
+++ b/_templates/playground/new/index.js
@@ -33,7 +33,7 @@ module.exports = {
               : {
                   type: 'input',
                   name: 'name',
-                  message: 'What is this playground for?',
+                  message: 'Which guide section is this playground for?',
                   initial: 'animations',
                 },
             {
--- a/_templates/playground/new/index.js
+++ b/_templates/playground/new/index.js
@@ -69,12 +69,12 @@ module.exports = {
             },
           ])
           .then((answers) => {
-            const componentName = changeCase.pascal(answers.path.split('/').pop());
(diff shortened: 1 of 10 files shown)


_templates/playground/new/index.js
```diff
@@ -5,60 +5,80 @@ const changeCase = require('change-case');
...
+                    return value.match(/^ion-[a-z/-]*[a-z]+$/)
+                      ? true
+                      : "Component name must be kebab-case and begin with 'ion-'";
+                  },
+                }
+              : {
+                  type: 'input',
+                  name: 'name',
+                  message: 'What is this playground for?',
+                  initial: 'animations',
```
brandyscarney, line 37: Should we add some validation here like the others to make sure only letters are entered?

_templates/playground/new/index.js
```diff
@@ -5,60 +5,80 @@ const changeCase = require('change-case');
...
+                  validate(value) {
+                    return value.match(/^ion-[a-z/-]*[a-z]+$/)
+                      ? true
+                      : "Component name must be kebab-case and begin with 'ion-'";
+                  },
+                }
+              : {
+                  type: 'input',
+                  name: 'name',
+                  message: 'What is this playground for?',
```
brandyscarney, line 36:
```suggestion
                  message: 'Which guide section is this playground for?',
```

I think it's probably safe to say this has to go in a guide if it isn't a component, right?

brandyscarney (changes requested):

_templates/playground/new/index.js
mapsandapps: Good idea! Done in 4f4abb3.

brandyscarney (approved):
Looks great!

Issue #2892, opened 2023-04-07
bug: icons don't render with Angular and JS in StackBlitz


thetaPC:
### Describe the Bug

- v7 & v8

The `<ion-icon>` does not render when using Angular and JS. It displays nothing at all. The inspector shows that it's in the DOM but the `<svg>` isn't in the DOM.

Screenshot of the missing `<svg>`:
![Screenshot 2023-04-07 at 1 15 20 PM](https://user-images.githubusercontent.com/13530427/230673251-90d84236-d684-4573-aeb0-33f9805452c0.png)

### Expected Behavior

I expect to visually see the icon on the screen.

Screenshot of what the DOM should be:
![Screenshot 2023-04-07 at 1 18 08 PM](https://user-images.githubusercontent.com/13530427/230673512-a62cd2bd-c87a-452d-a62f-0e1f139413c0.png)

### Steps to Reproduce

1. Navigate to the [icon page](https://ionicframework.com/docs/api/icon) on the docs site.
2. Verify that you are viewing version 7 or 8.
3. Navigate to the [basic usage](https://ionicframework.com/docs/api/icon#basic-usage).
4. Click on either Angular or JS
5. Click on either iOS or MD (bug is shown in both)
6. Click on the StackBlitz logo
7. Wait for StackBlitz to load
8. You will see only a white screen

PR #3105, opened 2023-08-31
docs(fab): add playground for safe area workaround

mapsandapps:
This playground and docs section explains how to adjust the fab positioning to account for the safe area when not using ion-header or ion-footer.

Vercel deploy: https://ionic-docs-git-fw-4713-ionic1.vercel.app/docs/api/fab#safe-area


static/usage/v7/fab/safe-area/angular/example_component_html.md
```diff
@@ -0,0 +1,7 @@
+```html
+<ion-fab slot="fixed" vertical="top" horizontal="center">
+  <ion-fab-button>
+    <ion-icon name="add"></ion-icon>
```
mapsandapps, line 4: The icons in the angular and javascript examples don't load in stackblitz because of #2892

docs/api/fab.md
```diff
@@ -41,6 +41,26 @@ import Positioning from '@site/static/usage/v7/fab/positioning/index.md';
 
 <Positioning />
 
+### Safe Area
+
+If there is no `ion-header` or `ion-footer` component, the fab may be covered by a device's notch, status bar, or other device UI. In these cases, the [safe area](/docs/layout/safe-area) on the top and bottom is not taken into account. This can be adjusted by using the [`--ion-safe-area-(dir)` variables](#application-variables).
```
brandyscarney, line 46: The `safe area` link brings me to a `Page Not Found` and `#application-variables` doesn't seem to exist?

brandyscarney (commented):
The demos look good but `--ion-safe-area-top` is `0` so it doesn't really show anything. Do we want to set this to something for demo purposes?

mapsandapps:
Hmm, yeah it looks good in the device preview but doesn't really show something in the stackblitzes. I'm not sure if it makes sense to adjust them or not. What do you think?

If we set it, the code will show up in the snippets, which seems more confusing?

docs/api/fab.md
mapsandapps: Thanks for catching those! Fixed in 5b13f717e873e9d98ac5f0ebbffe31ae29b1012d

docs/api/fab.md
```diff
@@ -41,6 +41,26 @@ import Positioning from '@site/static/usage/v7/fab/positioning/index.md';
...
+
+And when using a fab with `vertical` set to `"bottom"` without an `ion-footer`, the bottom margin needs to be set:
+
+```css
+ion-fab {
+  margin-bottom: var(--ion-safe-area-bottom, 0);
+}
+```
+
+If there is an `ion-header` (for a fab with `vertical` set to `"top"`) or `ion-footer` (for a fab with `vertical` set to `"bottom"`), no CSS adjustment is needed because the fab gets positioned relative to the header or footer.
```
brandyscarney, line 62: Can we just combine these into the same Playground? This info kind of gets lost underneath the existing one.

static/usage/v7/fab/safe-area/angular/global_css.md
```diff
@@ -0,0 +1,16 @@
...
+   */
+  --ion-safe-area-top: 20px;
+  --ion-safe-area-bottom: 20px;
+  --ion-safe-area-left: 20px;
+  --ion-safe-area-right: 20px;
+}
+
+ion-fab {
+  margin-top: var(--ion-safe-area-top, 0);
+}
```
brandyscarney, line 15: Should we separate this CSS out into the example's component file, that way it doesn't apply to _all_ `ion-fab` elements if they just had one page where a header was missing, for example.

docs/api/fab.md
```diff
@@ -41,6 +41,32 @@ import Positioning from '@site/static/usage/v7/fab/positioning/index.md';
...
+ion-fab {
+  margin-bottom: var(--ion-safe-area-bottom, 0);
+}
+```
+
+import SafeArea from '@site/static/usage/v7/fab/safe-area/index.md';
+
+<SafeArea />
+
+If there is an `ion-header` (for a fab with `vertical` set to `"top"`) or `ion-footer` (for a fab with `vertical` set to `"bottom"`), no CSS adjustment is needed because the fab gets positioned relative to the header or footer.
```
brandyscarney, line 68: I think we should keep this above the playground. It's easy to miss it with the long code examples:

Screenshot 2023-09-05 at 10 25 20 AM

brandyscarney (approved):

PR #3107, opened 2023-09-06
feat(many): update playgrounds to use showConsole prop instead of showing data within demo

amandaejohnston:
See https://github.com/ionic-team/ionic-docs/pull/3060 for details on this prop. The commits are split up by playground, so you may find it easier to review them individually.

### DIFF
--- a/static/usage/v7/modal/controller/demo.html
+++ b/static/usage/v7/modal/controller/demo.html
@@ -23,7 +23,6 @@
       </ion-header>
       <ion-content class="ion-padding">
         <ion-button expand="block" onclick="openModal()">Open</ion-button>
-        <p id="message">This modal example uses the modalController to present and dismiss modals.</p>
       </ion-content>
     </ion-app>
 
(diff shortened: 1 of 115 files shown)


static/usage/v7/modal/controller/demo.html
```diff
@@ -58,7 +57,7 @@
         const { data, role } = await modal.onWillDismiss();
 
         if (role === 'confirm') {
-          document.querySelector('#message').innerHTML = `Hello, ${data}!`;
```
mapsandapps, line 61: I wonder if maybe these modal ones felt more natural before the change? It's fine either way, but I kind of actually expected this one to be displayed in the DOM instead of the console. I feel like it's a common pattern for this sort of example.

mapsandapps (commented):
Everything looks great! 🤩 Just one question

static/usage/v7/modal/controller/demo.html
amandaejohnston: I was sort of wondering the same, to be honest 🤔 I only changed them for consistency's sake. I'll go ahead and revert them.

mapsandapps (approved):

PR #3108, opened 2023-09-07
Update 2-taking-photos.md

StevenInc:
corrected composables to components


mapsandapps:
The original code snippet is correct. `usePhotoGallery` is not a component, and it is intended to be located in the `composables` folder.

PR #3111, opened 2023-09-08
docs: add docs for dynamic font scaling

liamdebeasi:
~Do not merge until Dynamic Font Scaling has been released.~

We're merging this into `feature-7.5` which won't go live until Dynamic Font Scaling is live.


docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,205 @@
...
+
+:::tip
+Be sure to try this on an Android, iOS, or iPadOS device.
+:::
+
+Follow the [Changing the Font Size on a Device](#changing-the-font-size-on-a-device) guide to set your preferred font size, and watch the text in the demo below grow or shrink according to your preferences.
+
+import DynamicFontScaling from '@site/static/usage/v7/layout/dynamic-font-scaling/index.md';
+
+<DynamicFontScaling />
```
amandaejohnston, line 17: This doesn't work for me, unfortunately 🤔 I wonder if the iframe is causing problems, since the font size adjustment doesn't make it all the way through. I'm using Chrome on a Galaxy S10e.

|Before font scaling|After font scaling |
|--------------|---------------|
|before|after|

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,205 @@
...
+
+### iOS
+
+Font scaling on iOS can be configured in Settings > Accessibility > Display & Text Size > Larger Text. The slider on that page allows users to make the text larger or smaller across all apps that support font scaling.
+
+Users can also tap the "Large Accessibility Sizes" switch to access more font scale options.
+
+### Android
+
+Font scaling on Android can be configured in Settings > Accessibility > Display size and text. The "Font size" slider on that page allows users to make the text larger or smaller across all apps that support font scaling.
```
amandaejohnston, line 172: This varies widely across devices; for example, on my phone it's Settings > Accessibility > Visibility enhancements > Font size and style. I also worry that we'll need to maintain these steps as OS updates come out. Maybe we should just say it can be updated in the accessibility settings and leave it at that?

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Troubleshooting
+
+### Dynamic Font Scaling is not working
+
+There are a number of reasons why Dynamic Font Scaling may not have any effect on an app . The following list, while not exhaustive, provides some things to check to debug why Dynamic Font Scaling is not working.
```
amandaejohnston, line 178:
```suggestion
There are a number of reasons why Dynamic Font Scaling may not have any effect on an app. The following list, while not exhaustive, provides some things to check to debug why Dynamic Font Scaling is not working.
```

docs/layout/dynamic-font-scaling.md
liamdebeasi: Yep that's fair. Fixed in https://github.com/ionic-team/ionic-docs/pull/3111/commits/80e48a6aa0c1d7731e399af1b2d993343ca068d1

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Using Dynamic Font Scaling
+
+### Integrating Custom Components
+
+Developers can configure their custom components to take advantage of Dynamic Font Scaling by converting any pixel font sizes to use [rem units](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units#lengths). An easy way to convert from `px` to `rem` is to divide the pixel font size by the default browser font size, which is typically 16px.
+
+For example, if a component has a font size of `14px`, then this could be converted to `rem` by doing `14px / 16px = 0.875rem`.
+
+Developers should also audit their applications and make any other changes necessary to support larger font sizes. One thing to keep in mind is that the dimensions of your components may need to change to accommodate the larger font sizes. As a result, it may be necessary to change properties such as `width` and `height` to `min-width` and `min-height`. We also recommend having long text wrap to the next line instead of truncating to keep large text readable.
```
brandyscarney, line 51: Should we also mention `line-height` here? That seems to be one that gets set to `px` values sometimes.

docs/layout/dynamic-font-scaling.md
amandaejohnston: Should we also update the iOS description, so we don't need to worry about maintaining that either? I don't know how stable that sort of thing usually is on iPhones 😆

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,205 @@
...
+```css
+html {
+  --ion-dynamic-font: var(--ion-default-dynamic-font);
+  --ion-font-family: 'Comic Sans';
+}
+```
+
+### `em` units versus `rem` units
+
+Developers have two options for relative font sizes: `em` and `rem`.
```
brandyscarney, line 68: Maybe link to this section? https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units#ems_and_rems

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,205 @@
...
+
+`em` units set the font size of an element relative to the font size of its parent.
+
+In the following example, the computed font size of `.child` is `40px` because it is a child of `.parent` (2em \* 20px = 40px).
```
brandyscarney, line 72:
```suggestion
In the following example, the computed font size of `.child` is `40px` because it is a child of `.parent` (`2em \* 20px = 40px`).
```

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,205 @@
...
+<div class="parent">
+  Parent element with 20px
+  <div class="child">
+    Child element with 40px
+    <div class="child">Child element with 80px</div>
+  </div>
+</div>
+```
+
+<div style={{ fontSize: '20px' }}>
```
brandyscarney, line 96: Should we put this in a playground demo with just a javascript example? I know you don't need ionic for this, so we could override the index file and keep it simple.

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,205 @@
...
+}
+
+.child {
+  font-size: 2rem;
+}
+```
+
+## How Dynamic Font Scaling works in Ionic
+
+All Ionic components that define font sizes use [rem units](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units#lengths). This sizes the text in each component relative to the font size of the root element, which is usually the `html` element. This means that as the root element's font size changes, the text in all Ionic components scale in a consistent manner. This avoids the need to manually override each component's font size.
```
brandyscarney, line 120: > All Ionic components that define font sizes use rem units

This is inaccurate, no? We have some components that don't scale (tabs) and others that use `em` units (icons).

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,205 @@
...
+
+### iOS
+
+Dynamic Font Scaling in Ionic builds on top of an iOS feature called [Dynamic Type](https://developer.apple.com/documentation/uikit/uifont/scaling_fonts_automatically#overview). To do this, Ionic sets the [font](https://developer.mozilla.org/en-US/docs/Web/CSS/font) of the root element to an Apple-defined text style. For consistency, Ionic uses the [body](https://developer.apple.com/documentation/uikit/uifont/textstyle/1616682-body) text style.
+
+Using the Apple-defined text style enables Dynamic Type, allowing all text in Ionic components to scale according to the system-level preference. Note that these Apple-defined fonts only work on Apple devices. As a result, these fonts will not work on Android devices even if your app is using `'ios'` mode.
```
brandyscarney, line 126:
```suggestion
Using the Apple-defined text style enables Dynamic Type, allowing all text in Ionic components to scale according to the system-level preference. Note that these Apple-defined fonts only work on Apple devices. As a result, these fonts will not work on Android devices even if your app is using `ios` mode.
```

We don't usually put this in quotes

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,205 @@
...
+
+Ionic follows [Apple's Human Interface Guidelines for Typography](https://developer.apple.com/design/human-interface-guidelines/typography) when an app is in `'ios'` mode. As a result, important content is prioritized when the text size changes. This means a few things:
```
brandyscarney, line 128:
```suggestion
Ionic follows [Apple's Human Interface Guidelines for Typography](https://developer.apple.com/design/human-interface-guidelines/typography) when an app is in `ios` mode. As a result, important content is prioritized when the text size changes. This means a few things:
```

docs/layout/dynamic-font-scaling.md
liamdebeasi: Good idea: https://github.com/ionic-team/ionic-docs/pull/3111/commits/c105df8a27fceb2ec8b0d3c65754010761018f50

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,205 @@
...
+
+1. Content in an `ion-header` or an `ion-footer` will have maximum font sizes to prioritize text in `ion-content` which is deemed more important than text in `ion-header` and `ion-footer`.
```
brandyscarney, line 130:
```suggestion
1. Content in an `ion-header` or an `ion-footer` will have maximum font sizes to prioritize content in `ion-content` which is deemed more important than content in the `ion-header` and `ion-footer`.
```

Content other than text can be in all of these elements.

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,205 @@
...
+
+In the following example we are using the [min()](https://developer.mozilla.org/en-US/docs/Web/CSS/min) function to indicate that the font size of `.foo` should be no larger than `14px`.
+
+```css
+.foo {
+  font-size: min(1rem, 14px);
+}
+```
+
+If the root element's default font size is `16px`, and the system-level font scale is 1.5 (i.e text sizes should be increased by 50%), then `1rem` will evaluate to `24px` because `16 * 1.5 = 24`.
```
brandyscarney, line 146:
```suggestion
If the root element's default font size is `16px`, and the system-level font scale is `1.5` (i.e text sizes should be increased by 50%), then `1rem` will evaluate to `24px` because `16 * 1.5 = 24`.
```

docs/layout/dynamic-font-scaling.md
liamdebeasi: I updated the demo so it opens in a new tab/avoids iframes: https://github.com/ionic-team/ionic-docs/pull/3111/commits/d16510a95eb09db4c2ad4e0feb2c5fa25bfaf7e3

I also added a troubleshooting step because of this in case someone tries to load Ionic in an iframe: https://github.com/ionic-team/ionic-docs/pull/3111/commits/b37af9d598dbf1072e53aee75aa00662e4e34461

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,205 @@
...
+
+Using the Apple-defined text style enables Dynamic Type, allowing all text in Ionic components to scale according to the system-level preference. Note that these Apple-defined fonts only work on Apple devices. As a result, these fonts will not work on Android devices even if your app is using `'ios'` mode.
+
+Ionic follows [Apple's Human Interface Guidelines for Typography](https://developer.apple.com/design/human-interface-guidelines/typography) when an app is in `'ios'` mode. As a result, important content is prioritized when the text size changes. This means a few things:
+
+1. Content in an `ion-header` or an `ion-footer` will have maximum font sizes to prioritize text in `ion-content` which is deemed more important than text in `ion-header` and `ion-footer`.
+2. Components such as `ion-badge` and `ion-back-button` will have minimum font sizes so they remain readable.
+3. Text in components such as `ion-tab-bar` and `ion-picker` do not participate in Dynamic Font Scaling according to Apple's Human Interface Guidelines.
+
+### Android
```
brandyscarney, line 134: I'm confused by this entire section. We aren't using `min`, `max` or `clamp` on Android values in Ionic and this section is under `How Dynamic Font Scaling works in Ionic`. Would this be better suited as info under `Using Dynamic Font Scaling` or a section with info on how the different devices / operating systems handle dynamic scaling?

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,204 @@
...
+
+If the root element's default font size is `16px`, and the system-level font scale is 1.5 (i.e text sizes should be increased by 50%), then `1rem` will evaluate to `24px` because `16 * 1.5 = 24`.
+
+This is larger than our defined maximum of `14px`, so one might assume that the evaluated font size of `.foo` is `14px`. However, since the Android Web View scales any font sizes defined using the `px` unit, this means the `14px` used in our `min()` function will also be scaled by 1.5.
+
+As a result, this means that the maximum computed font size is actually `21px` since `14 * 1.5 = 21` and therefore the overall computed font size of `.foo` is `21px`.
+
+### Using Modes on Different Platforms
+
+Each platform has slightly different font scaling behaviors, and the `'ios'` and `'md'` modes have been implemented to take advantage of the scaling behaviors on their respective platforms.
```
brandyscarney, line 154:
```suggestion
Each platform has slightly different font scaling behaviors, and the `ios` and `md` modes have been implemented to take advantage of the scaling behaviors on their respective platforms.
```

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,204 @@
...
+
+For example, `'ios'` mode makes use of maximum and minimum font sizes to follow [Apple's Human Interface Guidelines for Typography](https://developer.apple.com/design/human-interface-guidelines/typography). `'md'` mode does not implement this same behavior because Material Design does not have that same guidance. This means that using `'md'` mode on an iOS device may allow for very large font sizes in headers and footers.
```
brandyscarney, line 156:
```suggestion
For example, `ios` mode makes use of maximum and minimum font sizes to follow [Apple's Human Interface Guidelines for Typography](https://developer.apple.com/design/human-interface-guidelines/typography). `md` mode does not implement this same behavior because Material Design does not have that same guidance. This means that using `md` mode on an iOS device may allow for very large font sizes in headers and footers.
```

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,204 @@
...
+
+As a result, we strongly recommend using `'ios'` mode on iOS devices and `'md'` mode on Android devices when using Dynamic Font Scaling.
```
brandyscarney, line 158:
```suggestion
As a result, we strongly recommend using `ios` mode on iOS devices and `md` mode on Android devices when using Dynamic Font Scaling.
```

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,206 @@
...
+}
+```
+
+Developers should ensure that the [typography.css](./global-stylesheets#typographycss) file is imported.
+
+## Using Dynamic Font Scaling
+
+### Integrating Custom Components
+
+Developers can configure their custom components to take advantage of Dynamic Font Scaling by converting any pixel font sizes to use [rem units](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units#lengths). An easy way to convert from `px` to `rem` is to divide the pixel font size by the default browser font size, which is typically 16px.
```
brandyscarney, line 47:
```suggestion
Developers can configure their custom components to take advantage of Dynamic Font Scaling by converting any pixel font sizes to use [rem units](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units#lengths). An easy way to convert from `px` to `rem` is to divide the pixel font size by the default browser font size, which is typically `16px`.
```

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Scaled Ionic iOS component font sizes do not exactly match native iOS equivalents
+
+Certain native iOS components such as the Action Sheet make use of private font scales that Ionic does not have access to. While we try to stay as close as possible to the native behavior, text in some components may render slightly larger or smaller than their native counterparts.
+
+### The text size in my Ionic app on iOS changed when enabling Dynamic Font Scaling
+
+The root element's default font size is typically 16px. However, Dynamic Font Scaling on iOS devices make use of the ["Body" text style](https://developer.apple.com/design/human-interface-guidelines/typography#Specifications) which has a default font size of 17px. Since the text in Ionic components is scaled relative to the root element's font size, some text may get larger or smaller when Dynamic Font Scaling is enabled, even if the system-level text scale did not change.
```
brandyscarney, line 202:
```suggestion
The root element's default font size is typically `16px`. However, Dynamic Font Scaling on iOS devices make use of the ["Body" text style](https://developer.apple.com/design/human-interface-guidelines/typography#Specifications) which has a default font size of `17px`. Since the text in Ionic components is scaled relative to the root element's font size, some text may get larger or smaller when Dynamic Font Scaling is enabled, even if the system-level text scale did not change.
```

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,206 @@
...
+
+:::info
+iOS provides a "Callout" text style which has a default font size of 16px. However, this font style is currently not exposed to web content. See [the supported text styles in WebKit](https://webkit.org/blog/3709/using-the-system-font-in-web-content/) for more information.
```
brandyscarney, line 205:
```suggestion
iOS provides a "Callout" text style which has a default font size of `16px`. However, this font style is currently not exposed to web content. See [the supported text styles in WebKit](https://webkit.org/blog/3709/using-the-system-font-in-web-content/) for more information.
```

brandyscarney (commented):
Some of my comments might be outdated since changes were made after I started reviewing.

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,205 @@
...
+
+Dynamic Font Scaling is already enabled by default on Android. Developers can enable it on iOS by setting the following CSS globally:
+
+```css
+html {
+  --ion-dynamic-font: var(--ion-default-dynamic-font);
+}
+```
+
+Under the hood, Ionic sets the following CSS on iOS devices to enable Dynamic Font Scaling:
```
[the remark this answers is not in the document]
liamdebeasi: I had it there to give more insight into how this feature works, but I can remove it if you think it will be confusing for devs.

docs/layout/dynamic-font-scaling.md
liamdebeasi: I mentioned `width` and `height` as an example -  I wasn't focused on calling out the most common properties. I can add it if you think it would help though.

docs/layout/dynamic-font-scaling.md
liamdebeasi: I think that may be overkill since that's going to show you Angular, React, and Vue code snippets which really aren't need. The purpose of this is to demonstrate how part of CSS works.

docs/layout/dynamic-font-scaling.md
liamdebeasi: That's true. Clarified in https://github.com/ionic-team/ionic-docs/pull/3111/commits/338b4730b0c039518c63e0d303af9ff17c4444b2

docs/layout/dynamic-font-scaling.md
liamdebeasi: Added in https://github.com/ionic-team/ionic-docs/pull/3111/commits/d94841fd5e6c2cb03b8c7819451db5536a168dd8

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,205 @@
...
+
+```css
+html {
+  font: var(--ion-dynamic-font);
+}
+```
+
+Developers should ensure that the [typography.css](./global-stylesheets#typographycss) file is imported.
```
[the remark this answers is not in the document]
liamdebeasi: Done in https://github.com/ionic-team/ionic-docs/pull/3111/commits/7eac2a32692f22cea02a70209b19e40b9af00a23.

docs/layout/dynamic-font-scaling.md
liamdebeasi: Our APIs are built on top of the browser behavior, so how Ionic behaves is influenced by the browser behavior which is why I had this section here. I mentioned min/max/clamp because developers potentially will need to use additional CSS to ensure their custom components support Dynamic Font Scaling. I wanted to call it out because developers may think the min/max/clamp behavior is an Ionic bug.

docs/layout/dynamic-font-scaling.md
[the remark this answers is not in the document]
brandyscarney: Maybe we could put it inside a "Note" or "Info" block so it is more of an informative note.

docs/layout/dynamic-font-scaling.md
brandyscarney: What if we linked to the documentation on [length](https://developer.mozilla.org/en-US/docs/Web/CSS/length) and just specified they should look over any CSS properties that accept length and change them from `px` to `rem` units?

docs/layout/dynamic-font-scaling.md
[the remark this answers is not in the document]
liamdebeasi: Updated in https://github.com/ionic-team/ionic-docs/pull/3111/commits/03febb1e212a48c915e603fb12ab721cd6df289f

docs/layout/dynamic-font-scaling.md
liamdebeasi: How does https://github.com/ionic-team/ionic-docs/pull/3111/commits/a9e12380a2df5fc6ec70a70f5d3ee0cefc99cae8 sound?

docs/layout/dynamic-font-scaling.md
amandaejohnston: We talked through this on Slack; turns out that there's a Chrome flag that's disabled by default on some devices (including mine) but is needed to make dynamic font scaling work. I confirmed that with the flag on, the playground now scales appropriately 👍 

Could we remove the troubleshooting item about not loading Ionic in an iframe, and replace it with one about ensuring the Chrome flag is enabled?

liamdebeasi: Ok we demystified this a bit more, so I documented our findings here: https://github.com/ionic-team/ionic-docs/pull/3111/commits/85f54d04f17550bd917e6b63cdc6b823493f2f81

Can you give this a try again with the Accessibility Page Zoom feature enabled?

docs/layout/dynamic-font-scaling.md
amandaejohnston: This paragraph was originally less about updating px values to rem and more dealing with edge cases like the element needing to be allowed to grow in size. At the moment, the info is pretty much duplicated between this paragraph and the two above it.

liamdebeasi: So my original goal with this section is to note that following:

1. Devs need to convert font size px units to rem
3. Devs may need to change other CSS values to account for the new font sizes

The change I made is meant to highlight point 3 whereas the first paragraph is meant to highlight point 1. Not every non-font-size CSS property that uses px will necessarily need to be converted to rem.

amandaejohnston: I suppose I felt like specifically mentioning that `width` and `height` may need updating to `min-width` and `min-height` was more useful, since that's not a px/rem change and seems common. It's okay how it is now though.

amandaejohnston (approved):

docs/layout/dynamic-font-scaling.md
brandyscarney: I like the addition of the link to the length docs. I think mentioning the `width`/`height` needing to change to `min-width`/`min-height` might be good for certain cases, like if they have a button with the following and want to support text wrapping:

```diff
button {
-  height: 48px;
+  min-height: 48px;
}
```

But in other cases they might have a rounded button like this that needs to scale proportionately:

```diff
button {
-  width: 48px;
-  height: 48px;
+  width: 3rem;
+  height: 3rem;
}
```

docs/layout/dynamic-font-scaling.md
```diff
@@ -0,0 +1,224 @@
...
+
+See https://bugs.chromium.org/p/chromium/issues/detail?id=645717 for more information.
+
+### Using Modes on Different Platforms
+
+Each platform has slightly different font scaling behaviors, and the `ios` and `md` modes have been implemented to take advantage of the scaling behaviors on their respective platforms.
+
+For example, `ios` mode makes use of maximum and minimum font sizes to follow [Apple's Human Interface Guidelines for Typography](https://developer.apple.com/design/human-interface-guidelines/typography). `md` mode does not implement this same behavior because Material Design does not have that same guidance. This means that using `md` mode on an iOS device may allow for very large font sizes in headers and footers.
+
+As a result, we strongly recommend using `ios` mode on iOS devices and `md` mode on Android devices when using Dynamic Font Scaling.
```
brandyscarney, line 172: > when using Dynamic Font Scaling.

Is there going to be a way for them to disable it when we make this the default?

brandyscarney (approved):

docs/layout/dynamic-font-scaling.md
liamdebeasi: OK I cleaned this up a bit. Thoughts on https://github.com/ionic-team/ionic-docs/pull/3111/commits/b0360dfb76ecd8bf971023f6b1d81d045fd61867?

amandaejohnston: Much better 👍

liamdebeasi: Thanks for the help both of you!

docs/layout/dynamic-font-scaling.md
brandyscarney: Maybe we can add this in once it becomes the default in Ionic 8

PR #3119, opened 2023-09-15
fix(playground): better handle multiple files for Javascript

mapsandapps:
When passing a `files` object to a Javascript playground, e.g.

```jsx
<Playground
  version="7"
  code={{
    javascript: {
      files: {
        'index.html': javascript_index_html,
        'theme/variables.css': variables_css,
      },
    },
    ...
  }}
  ...
>
```

you currently need to include the full HTML (e.g. the `head`) in the playground's markdown file that becomes index.html. This means that the code sample displayed also has all the extra stuff in it instead of being streamlined like it is when there's only the single file for a Javascript playground.

Screenshot 2023-09-18 at 9 59 57 AM

This PR updates the playground code so that you can simplify the file and what is displayed.

Explanation of change in Playground/index.tsx:
If we pass `codeBlock` through when there are multiple files for a framework just like we've been doing when there's a single file, we'll be able to use it in stackblitz.utils.ts:

Explanation of change in stackblitz.utils.ts:
Previously, we were replacing things in the index.html that was passed in (if one was). Now, we are replacing things in the default index.html. This preserves the head, etc., from the default index.html while still allowing the code from e.g. javascript.md to be passed in.

I don't think a design doc change is needed. This seems like more of a bug with the existing functionality than a new requirement. Let me know if you disagree.

Here's a [link](https://ionic-docs-git-fw-4886-ionic1.vercel.app/docs/theming/dark-mode#automatically-enable-dark-mode) to one of the updated playgrounds in the Vercel build.

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -366,13 +366,10 @@ export default function Playground({
       version,
     };
 
-    let codeBlock;
-    if (!hasUsageTargetOptions) {
-      // codeSnippets are React components, so we need to get their rendered text
-      // using outerText will preserve line breaks for formatting in Stackblitz editor
-      codeBlock = codeRef.current.querySelector('code').outerText;
-    } else {
-      codeBlock = codeRef.current.querySelector('code').outerText;
+    // using outerText will preserve line breaks for formatting in Stackblitz editor
+    let codeBlock = codeRef.current.querySelector('code').outerText;
+
+    if (hasUsageTargetOptions) {
       editorOptions.files = Object.keys(codeSnippets[usageTarget])
         .map((fileName) => ({
           [fileName]: hostRef.current!.querySelector<HTMLElement>(`#${getCodeSnippetId(usageTarget, fileName)} code`)
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -372,6 +372,7 @@ export default function Playground({
       // using outerText will preserve line breaks for formatting in Stackblitz editor
       codeBlock = codeRef.current.querySelector('code').outerText;
     } else {
+      codeBlock = codeRef.current.querySelector('code').outerText;
       editorOptions.files = Object.keys(codeSnippets[usageTarget])
         .map((fileName) => ({
           [fileName]: hostRef.current!.querySelector<HTMLElement>(`#${getCodeSnippetId(usageTarget, fileName)} code`)
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -367,7 +367,7 @@ export default function Playground({
     };
 
     // using outerText will preserve line breaks for formatting in Stackblitz editor
-    let codeBlock = codeRef.current.querySelector('code').outerText;
+    const codeBlock = codeRef.current.querySelector('code').outerText;
 
     if (hasUsageTargetOptions) {
       editorOptions.files = Object.keys(codeSnippets[usageTarget])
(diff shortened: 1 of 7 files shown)


static/usage/v7/theming/automatic-dark-mode/javascript.md
```diff
@@ -1,105 +1,86 @@
...
-        }
-      </script>
-
-      <style>
-        /*
+<style>
+  /*
       * Optional CSS
       * -----------------------------------
       */
```
brandyscarney, line 65: Can this be outdented? It doesn't align right in the code block:

Screenshot 2023-09-18 at 11 11 40 AM

static/usage/v7/theming/manual-dark-mode/javascript.md
```diff
@@ -1,132 +1,113 @@
...
+  function toggleDarkTheme(shouldAdd) {
+    document.body.classList.toggle('dark', shouldAdd);
+  }
+</script>
+
+<style>
+  /*
       * Optional CSS
       * -----------------------------------
       */
```
brandyscarney, line 92: Same as my other comment, this isn't aligned

brandyscarney (commented):
Nice work! This is looking great. I am just going to add @sean-perkins for review to make sure I didn't miss anything.

brandyscarney (approved):

src/components/global/Playground/index.tsx
```diff
@@ -372,6 +372,7 @@ export default function Playground({
       // using outerText will preserve line breaks for formatting in Stackblitz editor
       codeBlock = codeRef.current.querySelector('code').outerText;
     } else {
+      codeBlock = codeRef.current.querySelector('code').outerText;
```
sean-perkins, line 375: Can we refactor the code now that we are doing the same operation in both cases?

e.g.:
```ts
// using outerText will preserve line breaks for formatting in Stackblitz editor
const codeBlock = codeRef.current.querySelector('code').outerText;
if (hasUsageTargetOptions) {
  codeBlock = codeRef.current.querySelector('code').outerText;
  editorOptions.files = Object.keys(codeSnippets[usageTarget])
    .map((fileName) => ({
      [fileName]: hostRef.current!.querySelector<HTMLElement>(`#${getCodeSnippetId(usageTarget, fileName)} code`)
        .outerText,
    }))
    .reduce((acc, curr) => ({ ...acc, ...curr }), {});
  editorOptions.dependencies = (code[usageTarget] as UsageTargetOptions).dependencies;
}
```

sean-perkins (changes requested):

src/components/global/Playground/index.tsx
mapsandapps: oh yeah, good catch!

src/components/global/Playground/index.tsx
```diff
@@ -366,13 +366,10 @@ export default function Playground({
...
 
-    let codeBlock;
-    if (!hasUsageTargetOptions) {
-      // codeSnippets are React components, so we need to get their rendered text
-      // using outerText will preserve line breaks for formatting in Stackblitz editor
-      codeBlock = codeRef.current.querySelector('code').outerText;
-    } else {
-      codeBlock = codeRef.current.querySelector('code').outerText;
+    // using outerText will preserve line breaks for formatting in Stackblitz editor
+    let codeBlock = codeRef.current.querySelector('code').outerText;
```
sean-perkins, line 370: Do we ever re-assign `codeBlock`, if not this can be a `const`.

sean-perkins (approved):

PR #3122, opened 2023-09-18
chore: update v7 playground dependencies for React and Vue

liamdebeasi:
React/Vue playgrounds are pulling in Ionic 7.0.10. We are up to 7.4.0 now, so these should be updated.


sean-perkins (approved):
Hmm.. clearly the renovate configuration isn't working as expected.

PR #3139, opened 2023-09-21
chore(): update node to v19

mapsandapps:


sean-perkins (commented):
This looks good! Should we combine the changes from https://github.com/ionic-team/ionic-docs/pull/3137/files (using v19 instead of v18 though), so we can cut down on extra CI build time and just merge a single PR. 

Edit: Actually one sec

mapsandapps:
sure, i can do that!

sean-perkins (approved):

PR #3150, opened 2023-09-25
docs(angular): add links to Angular's standalone migration guide

amandaejohnston:
When going through the standalone component docs, I got confused because some of the code is specific to standalone architecture apps, but mine was still using `NgModule`. To help remedy this, I've added a couple links to Angular's migration guide, and noted that migrating the app and migrating the Ionic components used in the app are two different processes.

Open to suggestions on the copy. It may also be better to reorganize the sections somehow; I wasn't sure 🤔


amandaejohnston:
Closing in favor of https://github.com/ionic-team/ionic-docs/pull/3151.

PR #3153, opened 2023-09-26
docs(select): reorder examples, add alert playground

sean-perkins:
- Reorders the "Multiple selection" examples to be located after "Single selection"
- Adds "Alert" interface section that re-uses the existing playground example for "Single selection"


mapsandapps (approved):
nice 😎

liamdebeasi (approved):

PR #3158, opened 2023-09-28
docs(toast): add playground for positionAnchor

amandaejohnston:
Feature PR: https://github.com/ionic-team/ionic-framework/pull/28248

Dev build for testing: `7.4.2-dev.11695927273.1323de22`

### DIFF
--- a/docs/api/toast.md
+++ b/docs/api/toast.md
@@ -64,6 +64,14 @@ import ButtonsPlayground from '@site/static/usage/v7/toast/buttons/index.md';
 
 Toasts can be positioned at the top, bottom or middle of the viewport. The position can be passed upon creation. The possible values are `top`, `bottom` and `middle`. If the position is not specified, the toast will be displayed at the bottom of the viewport.
 
+### Handling Navigation Elements
+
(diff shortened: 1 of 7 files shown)


static/usage/v7/toast/position-anchor/react.md
```diff
@@ -0,0 +1,41 @@
+```tsx
+import React from 'react';
+import { IonButton, IonContent, IonFooter, IonHeader, IonTitle, IonToast, IonToolbar } from '@ionic/react';
+
+function Example() {
+  return (
+    <IonHeader id="header">
```
mapsandapps, line 7: I think all the stuff in this `return` needs to be wrapped by something, even if it's just a `<></>`

docs/api/toast.md
```diff
@@ -64,6 +64,14 @@ import ButtonsPlayground from '@site/static/usage/v7/toast/buttons/index.md';
 
 Toasts can be positioned at the top, bottom or middle of the viewport. The position can be passed upon creation. The possible values are `top`, `bottom` and `middle`. If the position is not specified, the toast will be displayed at the bottom of the viewport.
 
+### Handling Navigation Elements
```
mapsandapps, line 67: Perhaps this should say something like "Relative to navigation elements"? I'm not sure.

mapsandapps (changes requested):
Looks great! I like the demos. I just have one syntax issue and one question.

static/usage/v7/toast/position-anchor/react.md
amandaejohnston: Whoops, good catch 👀

docs/api/toast.md
amandaejohnston: Changed to "Relative Positioning" to be a little snappier, open to other ideas though 👀

mapsandapps (approved):

PR #3164, opened 2023-10-03
chore(support): update support table

mapsandapps:
I went and updated the support tables for the versioned docs too. If you have a better approach than just updating the tables, I'm open to ideas.


brandyscarney (approved):
LGTM!

PR #3165, opened 2023-10-04
chore: resolve cspell errors

sean-perkins:
- Fixes a missing word for the `feature-7.5` branch (Callout)
  - Added through VSCode, unsure why it sorted the list
- Updates the ignore paths to not parse any locally installed `node_modules` directories in the stackblitz template projects


cspell.json
```diff
@@ -17,7 +17,8 @@
     "docs/native",
     "versioned_docs/**/api",
     "versioned_docs/**/cli",
-    "versioned_docs/**/native"
+    "versioned_docs/**/native",
+    "static/code/stackblitz/**/node_modules/**"
```
brandyscarney, line 21: Should we just include all `node_modules`?

cspell-wordlist.txt
```diff
@@ -1,43 +1,31 @@
-Appflow
```
brandyscarney, line 1: Hmm I was trying to keep the sections separate because they are:

- Product names
- Component names
- All other words

sean-perkins: Hmm 🤔 so the question is if that is valuable over being able to `Cmd + .` to quick fix and add to the list. I'd argue I'm never going to open up this `txt` file to look at something other than adding a new word.

cspell.json
sean-perkins: Let me see. I know it excludes the root `node_modules` directory by default, but doesn't seem to account for nested folder structures with it.

sean-perkins: Silly that manually specifying it works 🤷 addressed here: https://github.com/ionic-team/ionic-docs/pull/3165/commits/874299f7e89049c0770417c08def4cbb042cae39

cspell-wordlist.txt
brandyscarney: I didn't even know it was possible to quickly add here. 😂 That's fine I am not set on it or anything.

brandyscarney (approved):

PR #3167, opened 2023-10-05
add titles and document router behavior

liamdebeasi:


mapsandapps (approved):
It took me a while to get these all working in sample apps (mostly user error). The docs are solid, though, and I did get everything working. But part of my confusion was that all of the code snippets in this PR are for using standalone, even though some of them are in apps that are also NgModules. I'm not sure if maybe these docs are getting too long to be on one page, or what the problem is. There are basically docs for three different situations all on this one page, and even with the table of contents, it's hard to remember which you're in: fully standalone, combo, or fully modules. If you accidentally scrolled into a section that doesn't apply to you, I think it could be really confusing.

Anyway, that's not a problem with this PR (although this does exacerbate it slightly), so let's go ahead and merge this.

PR #3169, opened 2023-10-05
feat(playground): persist framework and mode selection

amandaejohnston:
When the framework and/or mode are changed on any playground, the new selection is stored and persisted. This applies both to all other playgrounds currently on the page, and any future page loads.

In the design doc, it was mentioned [here](https://github.com/ionic-team/ionic-framework-design-documents/blob/main/projects/ionic-framework-docs/playground/0004-framework-preference.md#synchronization) that we should not immediately update the mode for any playgrounds that haven't been scrolled to yet, for performance reasons. I elected not to do this because it doesn't seem to confer any benefit -- the iframes are not rendered at all until the playground is scrolled to, even if the mode is changed ahead of time.

There also shouldn't be any performance issues with switching modes on playgrounds that _have_ been scrolled to, since both iframes are loaded at the same time anyway, so no extra loading is incurred. (Switching modes only toggles which iframe is hidden via CSS.)

Also note that we decided ([here](https://github.com/ionic-team/ionic-framework-design-documents/blob/main/projects/ionic-framework-docs/playground/0004-framework-preference.md#out-of-scope)) not to sync playground framework selection with Docusaurus `<Tabs>` components, at least for now.

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -432,7 +432,7 @@ export default function Playground({
    * causing isBrowser to be set to false even if running the app
    * in a browser (vs. SSR). isBrowser is then updated on the next
    * render cycle.
-   * 
+   *
    * This useEffect contains code that can only run in the browser,
    * and also needs to run on that first go-around. Note that
    * isBrowser will never be set from true back to false, so the
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -187,7 +187,7 @@ export default function Playground({
      * Otherwise, if there is a saved mode from previously clicking
      * the mode button, use that.
      */
-    const storedMode = localStorage.getItem(MODE_STORAGE_KEY);
+    const storedMode = window !== undefined ? localStorage.getItem(MODE_STORAGE_KEY) : null;
     if (storedMode) return storedMode;
 
     /**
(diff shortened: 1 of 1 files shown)


src/components/global/Playground/index.tsx
```diff
@@ -158,21 +170,66 @@ export default function Playground({
...
+  const isBrowser = useIsBrowser();
+
   const hostRef = useRef<HTMLDivElement | null>(null);
   const codeRef = useRef(null);
   const frameiOS = useRef<HTMLIFrameElement | null>(null);
   const frameMD = useRef<HTMLIFrameElement | null>(null);
   const consoleBodyRef = useRef<HTMLDivElement | null>(null);
 
-  const defaultMode = typeof mode !== 'undefined' ? mode : Mode.iOS;
+  const { blockElementScrollPositionUntilNextRender } = useScrollPositionBlocker();
```
thetaPC, line 186: `PlaygroundTabs` already uses this function. Does it need to be included in this file as well?

amandaejohnston: Yes, you have the call the function for every component that needs to block scroll on click.

amandaejohnston: Ah, I see where the confusion might lie -- PlaygroundTabs is for the tabs that switch which file is being shown within the code. The new addition is for the framework switcher for the whole playground.

thetaPC: Is it possible to not use `useScrollPositionBlocker`? I'm asking in preparation of the Docusaurus update. It was changed for [internal use only](https://github.com/ionic-team/ionic-docs/pull/3014#discussion_r1281143661). If it's needed, then I'll make a note of it for the update.

amandaejohnston: Hmm 🤔 We definitely need it or something like it; otherwise the scroll can jump around when toggling the frameworks which leads to an unpleasant UX. If we got rid of the import, we'd either need to copy the code directly into our repo, or roll our own solution.

thetaPC: Just adding the decision: `useScrollPositionBlocker` will be used as normal even when upgrading Docusaurus.

mapsandapps (approved):
Looks great; works great! 🤩 I'm glad it ended up being simpler than we anticipated.

PR #3175, opened 2023-10-06
docs(native): properly type camera example usage

danielcregg:
Updating to prevent error message "Type 'string | undefined' is not assignable to type 'string'.  Type 'undefined' is not assignable to type 'string.'" means that you are trying to assign a value that could be either a string or undefined to a variable that expects a string. In this case, you are trying to assign the value of imageUrl to the imageSrc variable, but imageUrl could be undefined if the user does not take a picture. My solution is to change the type of the imageSrc variable to string | undefined.


brandyscarney (approved):
Thank you for the PR and sorry for the delayed review! 🙂

PR #3176, opened 2023-10-06
docs(angular): add more info on migrating to standalone

brandyscarney:
While going through migrating an NgModules app to use Ionic standalone components I got confused because the previous section mentioned adding the imports in the `@Component` and not the `@NgModule` so I have added an example to help clarify.

### DIFF
--- a/docs/angular/build-options.md
+++ b/docs/angular/build-options.md
@@ -415,6 +415,33 @@ if (environment.production) {
 export class AppModule {}
 ```
 
+For example, all modules that are using Ionic components need to have the Ionic components imported in their component module.
+
+```diff title="home.module.ts"
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { FormsModule } from '@angular/forms';
+import { HomePage } from './home.page';
+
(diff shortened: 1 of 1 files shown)


liamdebeasi:
I'm a bit confused by the motivation for this change. The previous example in the migration guide mentions adding the imports to `@NgModule`, not `@Component`.

brandyscarney:
Sorry that was confusing. I was going through the follow sections, migrating from:

- [Standalone-based Applications​](https://ionicframework.com/docs/angular/build-options#standalone-based-applications)
- [NgModule-based Applications​](https://ionicframework.com/docs/angular/build-options#ngmodule-based-applications)

In the Standalone-based migration guide you add the imports to the `@Component` like `app.component.ts`, but in the NgModule-based migration you add the imports to the page's `@NgModule` files like `app.module.ts` and `home.module.ts`. I got confused because I thought the imports needed to be added to the page's `@Component` such as in `home.page.ts`, which caused a lot of errors during the migration. I had a call with @sean-perkins to clear up my confusion and he recommended adding a note for the individual page files to make it clearer.

liamdebeasi (approved):

PR #3185, opened 2023-10-10
fix(playground): wait until playground is in view before loading stored framework/mode

amandaejohnston:
We've found that switching the mode for all playgrounds on the page at once can cause performance issues for some devices. Namely, if many, many playgrounds are being toggled (such as on the [Datetime page](https://ionicframework.com/docs/api/datetime)), the browser may run out of memory and forcibly refresh the page.

Instead of updating all playgrounds at once, we now only update the ones currently on the screen, and wait for others to be scrolled on-screen before updating them as well.

Slack thread for context: https://ionicteam.slack.com/archives/C02U5GFFXED/p1696944131395219

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -837,5 +837,3 @@ const isFrameReady = (frame: HTMLIFrameElement) => {
 
 const USAGE_TARGET_STORAGE_KEY = 'playground_usage_target';
 const MODE_STORAGE_KEY = 'playground_mode';
-// const USAGE_TARGET_UPDATED_EVENT = 'playground-usage-target-updated';
-// const MODE_UPDATED_EVENT = 'playground-event-updated';
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -13,7 +13,6 @@ import TabItem from '@theme/TabItem';
 
 import { IconHtml, IconTs, IconVue, IconDefault, IconCss, IconDots } from './icons';
 
-import { useScrollPositionBlocker } from '@docusaurus/theme-common';
 import useIsBrowser from '@docusaurus/useIsBrowser';
 
 const ControlButton = forwardRef(
@@ -183,8 +182,6 @@ export default function Playground({
   const frameMD = useRef<HTMLIFrameElement | null>(null);
   const consoleBodyRef = useRef<HTMLDivElement | null>(null);
 
-  const { blockElementScrollPositionUntilNextRender } = useScrollPositionBlocker();
-
   const getDefaultMode = () => {
     /**
      * If a custom mode was specified, use that.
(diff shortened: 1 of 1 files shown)


src/components/global/Playground/index.tsx
```diff
@@ -401,73 +368,46 @@ export default function Playground({
...
-   * causing isBrowser to be set to false even if running the app
-   * in a browser (vs. SSR). isBrowser is then updated on the next
-   * render cycle.
-   *
-   * This useEffect contains code that can only run in the browser,
-   * and also needs to run on that first go-around. Note that
-   * isBrowser will never be set from true back to false, so the
-   * code within the if(isBrowser) check will only run once.
-   */
-  useEffect(() => {
```
amandaejohnston, line 441: Loading the mode/framework when the playground comes into view also covers playgrounds that were in view from the start (such as the [Button page](https://ionicframework.com/docs/api/button)), so this entire `useEffect` is no longer needed.

src/components/global/Playground/index.tsx
```diff
@@ -401,72 +407,92 @@ export default function Playground({
...
    *
-   * This useEffect contains code that can only run in the browser,
-   * and also needs to run on that first go-around. Note that
-   * isBrowser will never be set from true back to false, so the
-   * code within the if(isBrowser) check will only run once.
+   * We also re-run when isInView changes because the event callbacks
+   * would otherwise capture a stale state value. Since we need to
+   * listen for these events only when the playground is in view,
+   * we check the state before adding the listeners at all, rather
+   * than within the callbacks.
```
amandaejohnston, line 483: An alternative would be to manually calculate whether the playground is on screen within the event callbacks, rather than using a state variable at all. I chose the state route because we already have an `IntersectionObserver` doing the heavy lifting with calculation, but I can switch this over if preferred.

src/components/global/Playground/index.tsx
```diff
@@ -401,72 +407,92 @@ export default function Playground({
...
    *
-   * This useEffect contains code that can only run in the browser,
-   * and also needs to run on that first go-around. Note that
-   * isBrowser will never be set from true back to false, so the
-   * code within the if(isBrowser) check will only run once.
+   * We also re-run when isInView changes because otherwise, a stale
+   * state value would be captured. Since we need to listen for these
+   * events only when the playground is in view, we check the state
+   * before adding the listeners at all, rather than within the
+   * callbacks.
```
amandaejohnston, line 483: An alternative would be to manually calculate whether the playground is on screen within the event callbacks, rather than using a state variable at all. I chose the state route because we already have an IntersectionObserver doing the heavy lifting with calculation, but I can switch this over if preferred.

liamdebeasi (approved):
Works great on my Pixel 4a and iPhone 13

PR #3192, opened 2023-10-12
feat(keyboard): replace Codepens with playgrounds using modern syntax

amandaejohnston:
https://ionicframework.com/docs/developing/keyboard The two code demos on this page had several problems:
1. The demos were shown using Codepens instead of proper playgrounds.
2. The inputs were using the legacy form control syntax.
3. The code shown didn't match the Codepens.

This PR resolves all three.

Shortcut: https://ionic-docs-git-keyboard-playgrounds-ionic1.vercel.app/docs/developing/keyboard

### DIFF
--- a/docs/developing/keyboard.md
+++ b/docs/developing/keyboard.md
@@ -2,7 +2,6 @@
 title: Keyboard
 ---
 
-import Codepen from '@components/global/Codepen';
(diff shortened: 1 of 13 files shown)


mapsandapps (approved):
very nice 😎

PR #3220, opened 2023-10-31
fix(locales): correct links

thetaPC:
Issue URL: N/A

The links within the locale dropdown are not correct when navigating to the main page (without a trailing slash).

`/docs`
English -> `/docs/docs`
JP -> `/docs/ja/docs`

`/docs/ja`
English -> `/docs/docs/ja`
JP -> `/docs/ja/docs/ja`

## What is the new behavior?

The bad links now redirect to the correct page. We opted out from enabling `trailingSlash` on the docusaurus config since it would be a site-wide change that can lead to more issues.

- [ ] Yes
- [x] No

## Other information

I also noticed that the swizzled file `LocaleDropdownNavbarItem` had been updated at Docusaurus. I've updated it here as well.


thetaPC:
It was automatically fixed. It might have been done by Vercel.

PR #3225, opened 2023-11-02
docs(react): link to vitest docs

sean-perkins:
Adds a mention and link to the Vitest docs for unit testing with React.


liamdebeasi (changes requested):
Can we also add "Vitest" to the introduction section?

liamdebeasi (approved):

PR #3226, opened 2023-11-02
Fixed typo in dark-mode.md

oliveryasuna:
`#1c1c1d` follows iOS guidelines.


liamdebeasi:
Thanks for the PR! I spoke with the team, and Ionic's dark mode is working correctly here. Lists on iOS have different background colors based on the type of list. Regular/full width lists match the background color of the content. Inset lists tend to be a bit darker/lighter than the background color depending on if the app is in light/dark mode. We only expose a single list background color for simplicity, so we opted to stick with having the list match the content background color.

PR #3231, opened 2023-11-06
chore(renovate): update config to ignore typescript updates in react v6

brandyscarney:
I couldn't find a way to test this but I looked through the docs and issues and this seems like the right config. Let me know if I am missing something!


sean-perkins (approved):
The easiest way that I have found to test renovate is to fork the repository, update the config and then run renovate immediately to see which PR suggestions it creates. 

However these changes look accurate and are very low risk if we need to make an additional PR if they don't do what we think.

PR #3242, opened 2023-11-10
fix(playground): support angular 17 template

sean-perkins:
Fixes the generated playground examples to work with the new Angular 17 templates from Stackblitz.


amandaejohnston (approved):
Tested with the basic datetime playground, and then the controller modal one for something a little more complex; looks good 👍

PR #3246, opened 2023-11-13
docs(react): v6 playgrounds use vite

liamdebeasi:
The team would like to update the v6 React playgrounds to use Vite to a) allow for TypeScript 5 usage and b) align more closely with our starter apps. Create React App does not support TypeScript 5.

I made the following changes:

1. Remove create react app in favor of vite
2. Moved eslint rules to `.eslintrc.js` (alignment with react-vite starter)
3. Moved browserlist rules to `.browserslistrc` (alignment with react-vite starter)
4. Updated to TypeScript 5
5. Updated index.html to more closely align with the react-vite starters

Preview: https://ionic-docs-git-fw-5437-ionic1.vercel.app/docs

Note: There should not be any behavior changes in the playgrounds as a result of this update.


sean-perkins (approved):
Looks good against a few v6 playgrounds 🎉

PR #3248, opened 2023-11-14
chore(renovate): revert typescript ignore rule

liamdebeasi:
Reverts ionic-team/ionic-docs#3231

As of https://github.com/ionic-team/ionic-docs/pull/3246, the Ionic React v6 playgrounds now use Vite, so we can safely use TypeScript 5.


brandyscarney (approved):
All my hard work gone

PR #3251, opened 2023-11-16
Replace src/main.tsx by src/index.tsx

hbea:
It looks like `src/main.tsx` has been replaced by `src/index.tsx` in current version of ionic starter template (as of today at least).

### DIFF
--- a/docs/react/your-first-app.md
+++ b/docs/react/your-first-app.md
@@ -99,7 +99,7 @@ npm install @ionic/pwa-elements
 
 After installation, open up the project in your code editor of choice.
 
-Next, import `@ionic/pwa-elements` by editing `src/main.tsx`.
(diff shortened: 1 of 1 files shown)


brandyscarney:
Thank you for the PR! The latest major version of the Ionic CLI (v7) installs the `react-vite` starter when running the following command (from the guide):

```
ionic start photo-gallery tabs --type=react --capacitor
```

The `react-vite` starter does have a `main.tsx` file: https://github.com/ionic-team/starters/blob/main/react-vite/base/src/main.tsx

If you are running this command with a CLI version of v6 or earlier it will use the old react starter which contains an `index.tsx` file: https://github.com/ionic-team/starters/blob/main/react/base/src/index.tsx

Make sure you are using the latest version of the Ionic CLI when following the "Your First Ionic App: React" guide and if you are still seeing issues after that please create a new issue on this repo, thank you!

PR #3258, opened 2023-11-22
docs(autofocus): playground examples for setFocus

sean-perkins:
Preview: https://ionic-docs-git-sp-fw-1423-ionic1.vercel.app/docs/developing/managing-focus

This PR adds a new top-level page to "Developing" for educating developers on how and when to use the `setFocus` API available on certain components. 

Since the public API docs are automatically generated based on the `core.json`, links to these docs will be submitted in a PR in Ionic Framework once we agree on the content here.


docs/developing/autofocus.md
```diff
@@ -0,0 +1,29 @@
...
+    name="description"
+    content="Learn how to manage focus in Ionic applications using the setFocus API instead of the autofocus attribute."
+  />
+</head>
+
+The `autofocus` attribute is a standard HTML attribute that allows developers to set focus to an element when a page loads. This attribute is commonly used to set focus to the first input element on a page. However, the `autofocus` attribute can cause issues in routing applications when navigating between pages. This is because the `autofocus` attribute will set focus to the element when the page loads, but will not set focus to the element when the page is revisited.
+
+# Managing Focus
+
+Ionic provides a `setFocus` API that allows developers to manually set focus to an element. This API should be used in place of the `autofocus` attribute and called within the `ionViewDidEnter` lifecycle event for routing applications, within the `didPresent` lifecycle event for overlays, and the `appload` event for vanilla JavaScript applications.
```
liamdebeasi, line 17: We need to also note some of the platform-level restrictions since `setFocus` won't always work either. We don't need to create a detailed list of every little detail, but I think we could stick to the following high level points:

1. Interactive elements can only be focused as a result of a user gesture in Mobile Safari. (In other words, this playground won't work in mobile Safari). So for example, you can call `setFocus` in the `click` callback on a button as long as the user is the one tapping the button. However, you can't programmatically dispatch a click on the button and have `setFocus` work.
2. Android (both mobile Chrome and WebView) require user interaction on the app before `setFocus` can work. This can be as simple as tapping the screen.

docs/developing/autofocus.md
```diff
@@ -0,0 +1,29 @@
+---
+title: Autofocus
```
liamdebeasi, line 2: I think we should call this "Managing Focus" or something similar. Much of the content here has to do with `setFocus`, though `autofocus` is one instance where `setFocus` comes into play.

docs/developing/autofocus.md
```diff
@@ -0,0 +1,29 @@
...
+
+<head>
+  <title>Autofocus</title>
+  <meta
+    name="description"
+    content="Learn how to manage focus in Ionic applications using the setFocus API instead of the autofocus attribute."
+  />
+</head>
+
+The `autofocus` attribute is a standard HTML attribute that allows developers to set focus to an element when a page loads. This attribute is commonly used to set focus to the first input element on a page. However, the `autofocus` attribute can cause issues in routing applications when navigating between pages. This is because the `autofocus` attribute will set focus to the element when the page loads, but will not set focus to the element when the page is revisited.
```
liamdebeasi, line 13: This could probably be under its own heading since this documentation should be about managing focus in general, not just autofocus.

docs/developing/autofocus.md
```diff
@@ -0,0 +1,29 @@
...
+
+# Managing Focus
+
+Ionic provides a `setFocus` API that allows developers to manually set focus to an element. This API should be used in place of the `autofocus` attribute and called within the `ionViewDidEnter` lifecycle event for routing applications, within the `didPresent` lifecycle event for overlays, and the `appload` event for vanilla JavaScript applications.
+
+The example below highlights the difference in behavior between the `autofocus` attribute and the `setFocus` API.
+
+import Basic from '@site/static/usage/v7/autofocus/basic/index.md';
+
+<Basic />
```
liamdebeasi, line 23: The demo doesn't work in Safari for some reason.

sean-perkins:
After discussion with Liam, we decided to have the playground example showcase a manual interaction that requests focus, so that the experience is consistent across browsers. 

StackBlitz will steal the focus to the code editor, which is problematic for consistency in Safari with the lifecycle approach. As a result, I've added code snippets for each framework to demonstrate how to use the lifecycle APIs and how to handle it for overlays as well.

docs/developing/managing-focus.md
```diff
@@ -0,0 +1,244 @@
...
+  />
+</head>
+
+## Autofocus Attribute
+
+The `autofocus` attribute is a standard HTML attribute that allows developers to set focus to an element when a page loads. This attribute is commonly used to set focus to the first input element on a page. However, the `autofocus` attribute can cause issues in routing applications when navigating between pages. This is because the `autofocus` attribute will set focus to the element when the page loads, but will not set focus to the element when the page is revisited.
+
+## Managing Focus
+
+Ionic provides a `setFocus` API that allows developers to manually set focus to an element. This API should be used in place of the `autofocus` attribute and called within:
```
liamdebeasi, line 22:
```suggestion
Ionic provides a `setFocus` API on components such as [Input](../api/input), [Searchbar](../api/searchbar), and [Textarea](../api/textarea) that allows developers to manually set focus to an element. This API should be used in place of the `autofocus` attribute and called within:
```

This API is only exposed on specific elements.

static/usage/v7/input/set-focus/react.md
```diff
@@ -0,0 +1,23 @@
...
+        <IonItem>
+          <IonInput ref={input} label="Email" labelPlacement="floating"></IonInput>
+        </IonItem>
+      </IonList>
+    </IonPage>
+  );
+};
+
+export default Home;
+```
```
liamdebeasi, line 23: IonPage is no longer needed now that we removed routing from this demo:

```suggestion
```tsx
import React, { useRef } from 'react';
import { IonInput, IonItem, IonList, IonButton } from '@ionic/react';

const Home = () => {
  const input = useRef<HTMLIonInputElement>(null);

  return (
    <IonList>
      <IonItem>
        <IonButton onClick={() => input.current?.setFocus()}>Click to set focus</IonButton>
      </IonItem>
      <IonItem>
        <IonInput ref={input} label="Email" labelPlacement="floating"></IonInput>
      </IonItem>
    </IonList>
  );
};

export default Home;
```
```

File: static/usage/v7/input/set-focus/vue.md [hunk 7]
```diff
@@ -0,0 +1,25 @@
...
+  import { IonInput, IonItem, IonList, IonButton, IonPage } from '@ionic/vue';
+  import { ref } from 'vue';
+
+  const input = ref();
+
+  function setFocus() {
+    input.value.$el.setFocus();
+  }
+</script>
+```
```
liamdebeasi, line 25: IonPage is no longer needed now that we removed routing from this demo:

```suggestion
```html
<template>
  <ion-list>
    <ion-item>
      <ion-button @click="setFocus">Click to set focus</ion-button>
    </ion-item>
    <ion-item>
      <ion-input ref="input" label="Email" labelPlacement="floating"></ion-input>
    </ion-item>
  </ion-list>
</template>

<script setup lang="ts">
  import { IonInput, IonItem, IonList, IonButton } from '@ionic/vue';
  import { ref } from 'vue';

  const input = ref();

  function setFocus() {
    input.value.$el.setFocus();
  }
</script>
```
```

**Review by liamdebeasi — APPROVED** [review]
Good to go once the below comments are addressed. Nice job!

File: docs/developing/managing-focus.md [hunk 8]
```diff
@@ -0,0 +1,244 @@
...
+    name="description"
+    content="Learn how to manage focus in Ionic applications using the setFocus API instead of the autofocus attribute."
+  />
+</head>
+
+## Autofocus Attribute
+
+The `autofocus` attribute is a standard HTML attribute that allows developers to set focus to an element when a page loads. This attribute is commonly used to set focus to the first input element on a page. However, the `autofocus` attribute can cause issues in routing applications when navigating between pages. This is because the `autofocus` attribute will set focus to the element when the page loads, but will not set focus to the element when the page is revisited.
+
+## Managing Focus
```
amandaejohnston, line 20: Having this header be the same as the page title looks awkward. It also feels strange to offer `autofocus` and `setFocus` as two top-level sections; it looks like we're presenting the two as equal options with their own benefits and tradeoffs, when really we're explaining why `autofocus` doesn't work for most Ionic apps and showing how to use `setFocus` instead.

If it were me, I would do the following:
1. Remove the "Autofocus Attribute" header and go straight into the introductory content instead.
2. Trim the `autofocus` description paragraph to focus more on why it shouldn't be used. Link to the MDN docs for the attribute instead so developers can use that to learn more if they don't already know what it is.
3. Remove the "Managing Focus" header as well.
4. Raise all remaining headers to `##` instead of `###`.

An alternative would be to rename the `##` headers to something like "Downsides of `autofocus`" and "The `setFocus` API" respectively. Names could use workshopping, of course, but something that better shows the intent of the page. Personally though, I would rather draw the focus (hehe) of the page's outline entirely onto `setFocus` so things feel less cluttered.

docs/developing/managing-focus.md
```diff
@@ -0,0 +1,244 @@
...
+export default Home;
+```
+
+</TabItem>
+<TabItem value="vue">
+
+```html
+<template>
+  <ion-page>
+    <ion-input ref="input" label="setFocus" labelPlacement="floating"></ion-input>
```
amandaejohnston, line 108:
```suggestion
    <ion-input ref="input" label="setFocus" label-placement="floating"></ion-input>
```

static/usage/v7/input/set-focus/vue.md
```diff
@@ -0,0 +1,25 @@
+```html
+<template>
+  <ion-page>
+    <ion-list>
+      <ion-item>
+        <ion-button @click="setFocus">Click to set focus</ion-button>
+      </ion-item>
+      <ion-item>
+        <ion-input ref="input" label="Email" labelPlacement="floating"></ion-input>
```
amandaejohnston, line 9:
```suggestion
        <ion-input ref="input" label="Email" label-placement="floating"></ion-input>
```

amandaejohnston (changes requested):

amandaejohnston (approved):
Much better, thank you!

PR #3265, opened 2023-11-25
docs(react): update live reload steps to fix order in return

sylvainvdm:
Set return order of usePhotoGallery the same as expected in Tab2


brandyscarney (approved):

brandyscarney:
Thank you for the PR!

PR #3272, opened 2023-11-29
feat(translation/jp): 20231129 update

rdlabo:

PR #3273, opened 2023-11-29
chore(renovate): update react and react-dom

liamdebeasi:
The installed version of React on the docs is outdated. This can cause issues with other dependencies when the `package-lock.json` file is removed.

Example: https://github.com/ionic-team/ionic-docs/pull/3272#issuecomment-1831988786

A dependency we used dropped support for React 17 in a minor release. The package-lock ensured that an older version of the dep was installed (which did support React 17), but re-generating the package-lock results in the newer dep (which does not support React 18) getting installed.

Note: This won't actually do anything for now because we are using the latest version of React 17. Once we update to React 18 our installed version of react and react-dom will be kept up to date (up until v19 of react) by renovatebot.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -18,6 +18,10 @@
       "allowedVersions": "^5.0.0"
     },
     {
+      "matchPackageNames": ["react", "react-dom"],
+      "groupName": "react",
+    },
+    {
       "matchPackagePatterns": ["@ionic/"],
       "allowedVersions": "^6.0.0",
       "groupName": "ionic",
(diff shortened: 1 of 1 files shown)


thetaPC:
@liamdebeasi how can I test this?

liamdebeasi:
@thetaPC There's nothing to test at the moment. Our `package.json` wants `react@17` and `react-dom@17`, but we have the latest 17.x versions of those packages installed (so renovate won't run). Once we update to React 18 we will be able to keep these versions up to date (until v19) with renovate bot.

thetaPC (approved):
LGTM

PR #3274, opened 2023-11-30
fix(color-generator): contrast color calculation

sean-perkins:
This PR fixes a miscalculation in the color generator resulting in an incorrect contrast color. 

For example, with a color of `#EF3442` the expected contrast is `#000000`. However the color generator is incorrectly recommending `#ffffff`. 

This PR updates the logic for how the contrast color is determined by calculating the contrast ratio for the current color against a white and black background. The variant with the highest contrast ratio will be used as the contrast. 

The calculation for the contrast ratio was sourced from: https://stackoverflow.com/a/9733420. The code was refactored to match the existing code style for representing the `rgb` color and using the available types. 

Quick preview: https://ionic-docs-git-sp-color-generator-ionic1.vercel.app/docs/theming/color-generator

### DIFF
--- a/src/components/page/theming/_utils/color.ts
+++ b/src/components/page/theming/_utils/color.ts
@@ -133,6 +133,30 @@ const rgbToYIQ = ({ r, g, b }: RGB): number => {
   return (r * 299 + g * 587 + b * 114) / 1000;
 };
 
+const RED = 0.2126;
+const GREEN = 0.7152;
+const BLUE = 0.0722;
+const GAMMA = 2.4;
+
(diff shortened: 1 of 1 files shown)


sean-perkins:
@amandaejohnston the contrast color for `primary` being `#ffffff` is incorrect. If you enter those values into: https://webaim.org/resources/contrastchecker/ you will observe it fails appropriate color contrast. 

This corrects the calculation, which also corrects the values for `primary` to have proper color contrast (or at least take the highest contrast ratio).

amandaejohnston (approved):

PR #3278, opened 2023-12-01
translation(jp): merge contrast color calculation

rdlabo:
https://github.com/ionic-team/ionic-docs/pull/3274 is important changes.

So I would like to incorporate it into jp docs, albeit at a small granularity.


liamdebeasi (approved):

liamdebeasi:
Thank you!

PR #3287, opened 2023-12-04
fix(playground): tabs stackblitz example loads in Safari

### LINKED ISSUE #3260: bug: ion-tabs "Usage with Router" example doesn't send app.module.ts file contents to Stackblitz in Safari
### Describe the Bug

Enter in "https://stackblitz.com/run?file=src%2Fapp%2Fapp.component.ts" e see the error.
```typescript
Error in src/main.ts (3:27)
File '/~/src/app/app.module.ts' is not a module.
``` 

### Expected Behavior

Open the ion-tabs example

### Steps to Reproduce

Just enter in Ion-tabs official documentation "https://ionicframework.com/docs/api/tabs", nav to first example and click in "Open in Stackblitz"

### Screenshots

Screenshot 2023-11-22 at 17 55 13
Screenshot 2023-11-22 at 17 55 54

### Operating System

iOS

### Browser

Safari

### Version

17.1

### Additional Information

_No response_

sean-perkins:
Resolves #3260

Playground examples with many files would result in an issue with Safari where the `outerText` of the parsed code block would only include whitespace, not the actual contents of the code. 

This PR updates that logic to clone the DOM node to a detached node that results in the `outerText` consistently being parsed with the actual text content & whitespace/line breaks.

While working on the tabs example I noticed that the mode replacement logic would not apply because the regular expression matches on `IonicModule.forRoot({})` not `IonicModule.forRoot()`. I've updated the v7 & v6 examples to use the correct content. 

### Testing

To test this change, open the [Tabs docs](https://ionic-docs-git-sp-fw-5649-ionic1.vercel.app/docs/api/tabs) on Safari and attempt to open it in Stackblitz. You can confirm the expected behavior by observing the Stackblitz loads correctly & the `app.module.ts` has content.

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -547,18 +547,29 @@ export default function Playground({
     if (hasUsageTargetOptions) {
       editorOptions.files = Object.keys(codeSnippets[usageTarget])
         .map((fileName) => {
-          /**
-           * Safari has an issue where accessing the `outerText` on a non-visible
-           * DOM element results in a string with only whitespace. To work around this,
-           * we create a clone of the element, not attached to the DOM, and parse
-           * the outerText from that.
-           */
-          const el = document.createElement('div');
-          el.innerHTML = hostRef.current!.querySelector<HTMLElement>(
+          const codeBlock = hostRef.current!.querySelector<HTMLElement>(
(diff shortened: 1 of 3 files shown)


thetaPC (approved):
LGTM, thank you for including the comment. Verified that it works.

PR #3295, opened 2023-12-06
chore(select): remove unneeded patch from start/end slots demo

amandaejohnston:
This code was added to get around a bug in `feature-7.6` which has since been fixed, so it isn't needed anymore.


amandaejohnston:
The best way to test this would be to create a dev build of `feature-7.6`, but that's being actively released as I type this, so it may be easier to wait until that's done 😆

brandyscarney (approved):

PR #3298, opened 2023-12-07
docs(angular): note jest transformIgnorePatterns for es modules

liamdebeasi:
Developers have noted that Ionic's standalone components do not work with Jest by default: https://github.com/ionic-team/ionic-framework/issues/28667

This is happening because Ionic's components are ES Modules which Jest does not support. Developers need to use `transformIgnorePatterns` to ensure the ES Modules are transpired to a format Jest can use.

### DIFF
--- a/docs/angular/build-options.md
+++ b/docs/angular/build-options.md
@@ -1,3 +1,6 @@
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
 # Build Options
 
 Developers have two options for using Ionic components: Standalone or Modules. This guide covers both options as well as the benefits and downsides of each approach.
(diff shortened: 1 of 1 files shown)


liamdebeasi:
You'll need something slightly different when using pnpm:

```
transformIgnorePatterns: ['<rootDir>/node_modules/.pnpm/(?!(@ionic/angular|@ionic/core|ionicons|@stencil/core|@angular/*)@)']
```

This is covered in the Jest documentation: https://jestjs.io/docs/configuration#transformignorepatterns-arraystring (See the "Tip" section)

Squixx:
Man, thank you so much. I thought i was going crazy. That also explains why (for me) the direct import works, but the indirect doesnt... (pnpm puts a direct import into node_modules, but the indirect is hidden behind a `@` 

Would be nice to also add the pnpm part to the docs, as pnpm seems to be growing steadily in use. And I can imagine there are more people not familiar with it's inner workings

I dont think you need `@angular/*` for most cases, atleast our (large) repo hasn't ever needed it. So you might not want to include that as its quite a lot to transpile.

brandyscarney (approved):

PR #3303, opened 2023-12-08
docs(datetime): add new css variable for styling picker

brandyscarney:
Adds new CSS variable to datetime that was added with https://github.com/ionic-team/ionic-framework/pull/28654

---

⚠️ TODO - remove dev build `7.5.8-dev.11702320396.1294c304`

Preview link - https://ionic-docs-git-fw-5582-ionic1.vercel.app/docs/v8/api/datetime#wheel-pickers

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -46,6 +46,10 @@
     "static/code/stackblitz/v7/angular/package.json",
     "static/code/stackblitz/v7/html/package.json",
     "static/code/stackblitz/v7/react/package.json",
-    "static/code/stackblitz/v7/vue/package.json"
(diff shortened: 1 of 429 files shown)


liamdebeasi (approved):
Works well on my end.

> For some reason the preview isn't using the dev build but it works locally

Make sure you are on the v8 docs not the v7 docs (we default to v7). So the actual preview link would be https://ionic-docs-git-fw-5582-ionic1.vercel.app/docs/v8/api/datetime#wheel-pickers

brandyscarney:
@liamdebeasi That's exactly what I was doing wrong. 🤦‍♀️

PR #3320, opened 2023-12-13
Icon example not working - Update your-first-app.md

quickseries:
It would seem that none of the examples using a fresh new project following the getting started examples work with icons. It would seem that we need to import icons to our component classes?


thetaPC:
Thank you for submitting the PR!

However, I'm going to close this PR since the page is intended for an Angular NgModules app. Your proposed code would only work for a Standalone app. I'll be updating the page to highlight which option to choose when running `ionic start` to clear up any confusion. Thank you again for pointing this out!

PR #3345, opened 2023-12-21
docs(refresher): add rubber band effect for virtual scroll

liamdebeasi:
There was some confusion over how to get the native iOS refresher to work in a virtual scroll container. Ionic adds special CSS to allow for the rubber band effect even the container does not need to scroll. Developers need to add this to their apps too.


thetaPC (approved):
LGTM, pretty straight forward and easy to grasp.

PR #3346, opened 2023-12-21
docs(refresher): add rubber band effect for virtual scroll

liamdebeasi:
This is the same as https://github.com/ionic-team/ionic-docs/pull/3345 but targets Ionic 8.


thetaPC (approved):
LGTM, pretty straight forward and easy to grasp.

PR #3364, opened 2024-01-05
docs(hbb): document close watcher support

liamdebeasi:
Documentation for https://github.com/ionic-team/ionic-framework/pull/28705

I made the following changes:

1. Documented `experimentalCloseWatcher` in the Ionic Config.
2. Moved the Cordova/Capacitor text to a general "Overview" section. This text describes how HBB support in Ionic works. Now that we have experimental support for browser/PWA I think this text makes more sense in its own Overview section.
3. Added a support table so we can clearly define the status of HBB support in Ionic.
4. Updated the browser/PWA section to describe how to get HBB support in a browser as well as the limitations of this support.

Preview: https://ionic-docs-git-5804-docs-ionic1.vercel.app/docs


sean-perkins (approved):
Documentation looks great, nice work 👍

PR #3365, opened 2024-01-05
docs(desktop): remove desktop app deployment docs

mapsandapps:
Ionic Framework is focused on building mobile applications. This page no longer aligns with our strategy and therefore should be removed.

I'm going to take a look for other changes we may want to make regarding desktop, but I'll put those in a separate PR.


amandaejohnston (approved):

brandyscarney:
We should probably remove these here too: https://github.com/ionic-team/ionic-nginx/blob/d0c44a40a7f9a3bf5d6a747aade7dab8b7cd0500/config/redirects.conf#L741-L742

mapsandapps:
@brandyscarney oh yeah, good catch. maybe both sets of URLs should redirect somewhere else? maybe https://ionicframework.com/docs/deployment/progressive-web-app is the most similar page?

brandyscarney:
If you want to, I'll leave it up to you!

PR #3367, opened 2024-01-05
fix(playground): use correct usage target or mode, when limited

mapsandapps:
This PR fixes a bug: When the options for a playground's usage target or mode are limited, one of the unavailable options could show as selected. For mode, the playground would even show in the mode that wasn't supposed to be available.

**Reproduction (mode):**

1. set any playground to "md" mode
2. load this page https://ionicframework.com/docs/api/buttons#collapsible-buttons
3. see that it has md mode selected and is also displaying in md mode even though md mode is not available in the switcher

Screenshot 2024-01-05 at 2 13 46 PM

**Reproduction (framework):**

1. set any playground to a framework other than JavaScript
2. load https://ionicframework.com/docs/api/router
3. see that it has the other framework selected, when it should have JavaScript selected

Screenshot 2024-01-05 at 2 13 15 PM

### DIFF
--- a/src/components/global/Playground/index.tsx
+++ b/src/components/global/Playground/index.tsx
@@ -433,12 +433,8 @@ export default function Playground({
          * Load the stored mode and/or usage target, if present
          * from previously being toggled.
          */
-        if (isBrowser) {
-          const storedMode = localStorage.getItem(MODE_STORAGE_KEY);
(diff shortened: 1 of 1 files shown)


liamdebeasi (approved):
Thank you!

PR #3369, opened 2024-01-06
fix: correct debounced event name for ion-searchbar

SmallhillCZ:
see https://stackoverflow.com/a/56391887/1812179

### DIFF
--- a/docs/api/searchbar.md
+++ b/docs/api/searchbar.md
@@ -66,7 +66,7 @@ import Toolbar from '@site/static/usage/v7/toolbar/searchbars/index.md';
 
 ## Debounce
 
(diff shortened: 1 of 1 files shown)


mapsandapps:
Hi, thank you for the PR. In Ionic v7, Searchbar's `debounce` changed from delaying `ionChange` to delaying `ionInput`. You can read about this in the [changelog](https://github.com/ionic-team/ionic-framework/blob/main/BREAKING.md#version-7x-searchbar).

If you're using an earlier version of Ionic Framework, you can view the docs for that version by using the version switcher at the top of the docs. For example, [here are the Searchbar `debounce` docs for v6](https://ionicframework.com/docs/v6/api/searchbar#debounce).

I'm going to close this PR. Please let me know if you have any questions.

PR #3378, opened 2024-01-10
refactor(Stackblitz): use imported dark theme instead of hard-coded styles

amandaejohnston:
Updates the playground Stackblitz apps in all frameworks to use the new imported stylesheets for dark mode. It also updates the lockfiles for the React and Vue apps with the dev build that includes the latest AA contrast work; this was added in a previous PR, but updating the lockfiles was missed.

Because the light theme tokens have already been removed as well, the `variables.css` file for each framework is now empty. I decided to keep it around because the `open[Framework]Editor` utils in `stackblitz.utils.ts` have the file hard-coded, so we would need to update those functions to load different files depending on framework version, which would be hard to maintain. This way also allows us to add `variables.css` files to playgrounds without further infrastructure updates, similar to `global.css`.

To test, open any playground in all frameworks from the v8 version of the docs. Ensure that when you change your device to dark mode, the app updates accordingly.


static/code/stackblitz/v8/react/package.json
```diff
@@ -3,8 +3,8 @@
   "version": "0.1.0",
```
sean-perkins, line 3: Need to run `npm install` to update the lock file.

static/code/stackblitz/v8/html/index.ts
```diff
@@ -21,6 +21,17 @@ import '@ionic/core/css/display.css';
 /* Theme variables */
 import './theme/variables.css';
```
brandyscarney, line 22: I think we want this file to come after the import for the dark theme so it will properly override that file if desired.

amandaejohnston: Fixed in https://github.com/ionic-team/ionic-docs/pull/3378/commits/5b4cfed4bb75a63a29d86ca7c389264c3e9e75c7

brandyscarney (approved):

PR #3384, opened 2024-01-12
feat(theming): update dark mode guide with new dark theme files and playgrounds

brandyscarney:
Overhauls the Ionic Dark Mode documentation [here](https://ionicframework.com/docs/theming/dark-mode) to document the new dark theme files in v8. The section changes are:

| Before | After |
| ---| ---| 
| Before | After |

> [!NOTE]
> I did not modify the "Adjusting System UI Components" section.

In the v7 documentation, the sections are broken up to show how to add the dark theme by copying and pasting the variables and then using the media query or a class. In the v8 documentation, the sections go over how to add each of the 3 distributed theme files which includes all of the variables for the developer. 

The new playgrounds include ways to override the Ionic dark theme that we provide. I thought this would be beneficial so developers know that they can expand on our dark theme, and helpful to show which selectors they should use and where the styles should be placed.  

Please be critical when reviewing this. If anything seems unclear, needs refined headings or organization, or if the code examples could be improved, please let me know.

Preview link: https://ionic-docs-git-fw-5788-ionic1.vercel.app/docs/v8/theming/dark-mode
 
 > [!IMPORTANT]
> This uses a dev build off of https://github.com/ionic-team/ionic-framework/pull/28833 and requires that PR to be merged for this documentation to be accurate.


docs/theming/dark-mode.md
```diff
@@ -117,96 +210,246 @@ For developers looking to customize the theme color under the status bar in Safa
...
+2. Setting variables for the dark theme on `ios` devices using the `:root.ios` selector.
+3. Setting variables for the dark theme on `md` devices using the `:root.md` selector.
+
+:::caution
+It is important to pay attention to the specificity if you want to override any of the Ionic dark theme variables. For example, because the `--ion-item-background` variable is set for each mode, it cannot be overridden in the `:root` selector. A higher specificity selector, such as `:root.ios`, is required.
+:::
+
+```css
+:root {
+  --ion-color-primary: #4d8dff;
```
brandyscarney, line 233: I copied and pasted the dark theme variables from the framework here: https://github.com/ionic-team/ionic-framework/blob/feature-8.0/core/src/css/themes/dark.scss

I believe it would be better to import the file contents [from the CDN](https://cdn.jsdelivr.net/npm/@ionic/core@7.6.2-dev.11705355381.14b22962/css/themes/dark.class.css) to keep this section up to date with the latest version. However, I chose not to do so as it would expand the scope of this work.

static/usage/v8/theming/class-dark-mode/theme/variables_css.md
```diff
@@ -0,0 +1,33 @@
+```css
+/* Ionic Variables and Theming
+ * ---------------------------------------------------------------
+ * Any overrides to theme variables should be placed in this file.
+ * For more information, please see:
+ * http://ionicframework.com/docs/theming/
+ */
+
+/* This sets a different item border color for the default theme on ios and md */
```
brandyscarney, line 9: The reason I chose to specify that it updates the `default theme` and not the `light theme` or `light mode` in all of these examples is because the variables assigned to the `:root` selector will apply to **all** themes unless they are overridden specifically for that theme. This is also why I say `when dark mode is enabled` below, to make it clear that those variables will only apply in that case. If I can make this any more clear let me know.

docs/theming/dark-mode.md
```diff
@@ -14,76 +17,166 @@ inlineHtmlPreviews: true
...
+
+:::info Demo
+This demo combines site settings, system settings, and the toggle to decide when to show dark mode. The site's theme takes precedence over system settings. If your system settings differ from the site's theme when the demo loads, it will use the site's theme.
+
+Not sure how to change the system settings? Here's how to enable dark mode on [Windows 11](https://support.microsoft.com/en-us/windows/change-colors-in-windows-d26ef4d6-819a-581c-1581-493cfcc005fe) and on [macOS](https://support.apple.com/en-us/HT208976).
+:::
+
+import ClassDarkMode from '@site/static/usage/v8/theming/class-dark-mode/index.md';
+
+<ClassDarkMode />
```
thetaPC, line 179: I'm a bit confused as why the demo is using  the system settings. As a user, I would expect that the demo would only have the toggle to display how adding/removing the class causing the app's theme to change.

docs/theming/dark-mode.md
thetaPC: Could we provide a smaller example so we don't need to be maintaining this code snippet? If we keep it open ended then we don't have to worry about having to update the hex colors.

```css
:root {
  --ion-color-primary: // color provided by Ionic
}

:root.ios {
  --ion-color-primary: // your preferred color
}

:root.md {
  --ion-color-primary: // your preferred color
}
```

static/usage/v8/theming/class-dark-mode/javascript/index_ts.md
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Ionic Dark Theme
+ * -----------------------------------------------------
+ * For more information, please see:
+ * https://ionicframework.com/docs/theming/dark-mode
+ */
+
+// import '@ionic/core/css/themes/dark.always.css';
+import '@ionic/core/css/themes/dark.class.css';
+// import '@ionic/core/css/themes/dark.system.css';
```
thetaPC, line 29: Suggestion: Remove the commented out imports since this file targets the class dark mode. The same goes for the other class dark mode examples.

static/usage/v8/theming/system-dark-mode/angular/styles_css.md
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Ionic Dark Theme
+ * -----------------------------------------------------
+ * For more information, please see:
+ * https://ionicframework.com/docs/theming/dark-mode
+ */
+
+/* @import "~@ionic/angular/css/themes/dark.always.css"; */
+/* @import "~@ionic/angular/css/themes/dark.class.css"; */
```
thetaPC, line 37: Suggestion: Remove the commented out imports since this file targets the system dark mode. The same goes for the other system dark mode examples.

static/usage/v8/theming/always-dark-mode/angular/styles_css.md
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Ionic Dark Theme
+ * -----------------------------------------------------
+ * For more information, please see:
+ * https://ionicframework.com/docs/theming/dark-mode
+ */
+
+@import '~@ionic/angular/css/themes/dark.always.css';
+/* @import "~@ionic/angular/css/themes/dark.class.css"; */
+/* @import "~@ionic/angular/css/themes/dark.system.css"; */
```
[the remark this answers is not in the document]
brandyscarney: I brought this up [in Slack](https://outsystems.slack.com/archives/C06C5HAHURY/p1705078436774389) last week and everyone voted to keep the files commented out. I think the reason we want to keep all of them is because it shows that you need to comment out the file for the dark theme you don't want when changing from the default in the [starters](https://github.com/ionic-team/starters/blob/51dd4f8dca6ae28a1d9db378d168d4b13de8bea2/angular/base/src/global.scss#L28-L37).

docs/theming/dark-mode.md
```diff
@@ -14,76 +17,166 @@ inlineHtmlPreviews: true
...
+
+<TabItem value="angular">
 
 ```css
-@media (prefers-color-scheme: dark) {
-  :root {
-    /* Dark mode variables go here */
-  }
-}
+@import '~@ionic/angular/css/themes/dark.always.css';
```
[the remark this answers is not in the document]
brandyscarney: I must've copied it from the [Angular conference app](https://github.com/ionic-team/ionic-conference-app/blob/c99444b8270aa6c112fca13e7612cff5a377a7fb/src/global.scss#L47)

But it doesn't appear in the [Angular starter](https://github.com/ionic-team/starters/blob/51dd4f8dca6ae28a1d9db378d168d4b13de8bea2/angular/base/src/global.scss#L37) so I have removed it: https://github.com/ionic-team/ionic-docs/pull/3384/commits/1b3749a1deaeaf2a885fecabaa783e307921cecb

docs/theming/dark-mode.md
brandyscarney: It isn't supposed to be an example of setting a dark theme; instead, it is meant to serve as a reference for the entire Ionic dark theme we provide. I tried to override an application variable (`--ion-toolbar-background`) in `:root`, but it didn't work because the `md` theme was overriding it. As a result, I had to look up the defaults to see which variables are overridden by the theme. Providing developers with the entire theme file gives them the ability to see where we are setting the variables and how to override them.

docs/theming/dark-mode.md
brandyscarney: This was something [outlined in the design document](https://github.com/ionic-team/ionic-framework-design-documents/blob/main/projects/ionic-framework/theming/0002-distributing-theme-updates.md#documentation). I believe it was specifically requested to demonstrate to developers how they can expand upon reading from the system. Previously, we had these demos split up so that the first demo showed how to add the class based on the system settings, and then the second demo added the class based on the system settings along with a toggle. Since we offer the theme file for setting it based on the system, I used both the system and toggle for the class.

docs/theming/dark-mode.md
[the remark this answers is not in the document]
thetaPC: Understood, there's still some instances within `dark-mode.md` that also need to have the tilde removed.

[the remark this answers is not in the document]
brandyscarney: Oops https://github.com/ionic-team/ionic-docs/pull/3384/commits/b97806d5633c87065d6962aaf4cb440fa6c48342

docs/theming/dark-mode.md
```diff
@@ -14,76 +17,166 @@ inlineHtmlPreviews: true
...
+
+This sets the [application colors](/docs/theming/themes#application-colors) and [stepped colors](/docs/theming/themes#stepped-colors) on the [`:root`](https://developer.mozilla.org/en-US/docs/Web/CSS/:root) selector.
+
+:::caution Important
+Avoid targeting the `.ios` or `.md` selectors to override the Ionic dark theme, as these classes are added to each component and will take priority over globally defined variables. Instead, we can target the mode-specific classes on the `:root` element.
+:::
+
+:::info Demo
+This demo will always display the dark theme, regardless of the system settings for dark mode.
+:::
```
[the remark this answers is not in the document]
brandyscarney: I can remove the info, I mostly just kept it there to bring attention to how to change the system settings to view the dark mode, and then also added it to the "always" demo to be consistent:

Screenshot 2024-01-18 at 6 22 41 PM

This part:

> Not sure how to change the system settings? Here's how to enable dark mode on [Windows 11](https://support.microsoft.com/en-us/windows/change-colors-in-windows-d26ef4d6-819a-581c-1581-493cfcc005fe) and on [macOS](https://support.apple.com/en-us/HT208976).

I feel like it's easy to miss otherwise and people might not immediately know how to toggle the dark theme on. 

Here is what v7 looked like:

Screenshot 2024-01-18 at 6 24 43 PM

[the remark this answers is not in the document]
brandyscarney: Removed here: https://github.com/ionic-team/ionic-docs/pull/3384/commits/a00ba1120602efb26c2eba229611218b0c4f210c

static/usage/v8/theming/class-dark-mode/theme/variables_css.md
brandyscarney: So, I believe we should include examples demonstrating how to override the dark theme for a few reasons:

1. This guides developers on the appropriate selectors to use for overriding the existing dark theme, based on the specific dark theme file being used. For example, `dark.always.css` requires overrides in `:root` while `dark.system.css` requires overrides in `@media (prefers-color-scheme: dark) { :root {} }`. This is not covered by the [Colors documentation](https://ionicframework.com/docs/theming/colors#modifying-colors) or the [CSS Variables documentation](https://ionicframework.com/docs/theming/css-variables).
2. It provides an example of overriding theme variables based on the mode. For instance, in `dark.always.css`, `--ion-toolbar-background` is set on `:root.md`, not `:root`. Attempting to override it on `:root` will not work for `md` mode. This aligns with what I mentioned earlier about [including entire theme files for reference](https://github.com/ionic-team/ionic-docs/pull/3384#discussion_r1456506043).
3. It informs developers which file to use for overriding variables to ensure the variables are applied globally. Setting them in a `<style>` tag in Vue, for example, would not work, saving developers from trial and error. Variables should be set in the `variables.css` file.

If we wanted to emphasize more strongly that these overrides are optional, we can do that. Alternatively, we could consider splitting the demos, but this would require adding three more demos because each file has a different way to override variables.

docs/theming/dark-mode.md
thetaPC: Why not point to the repo file?

static/usage/v8/theming/class-dark-mode/theme/variables_css.md
thetaPC: Very good points. I'm good with leaving the files. There's also no need to emphasize since the file starts with a comment stating how they're used for overriding.

docs/theming/dark-mode.md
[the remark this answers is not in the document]
thetaPC: Hmm, then let's only keep the "Not sure how to..." part in the info block. I also feel that this can be easily missed.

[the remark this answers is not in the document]
brandyscarney: Done https://github.com/ionic-team/ionic-docs/pull/3384/commits/65fd8cb8fca306e0eec7c96b57bd6c2f9b60a0cf

docs/theming/dark-mode.md
brandyscarney: I'm not sure how to do that in a way that it would stand out and not be missed. I feel like links to the files would be easy to overlook. Ultimately I think we should have some kind of a preview widget where we could just include the CDN URL and then it would display the contents, but we would need to add that feature.

thetaPC: Ah, that makes sense. I can't think of a different alternative so this works.

thetaPC (approved):
LGTM

static/usage/v8/theming/always-dark-mode/angular/styles_css.md
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Ionic Dark Theme
+ * -----------------------------------------------------
+ * For more information, please see:
+ * https://ionicframework.com/docs/theming/dark-mode
+ */
+
+@import '@ionic/angular/css/themes/dark.always.css';
+/* @import "~@ionic/angular/css/themes/dark.class.css"; */
```
sean-perkins, line 37: We should remove the leading `~` from the commented out import.

static/usage/v8/theming/system-dark-mode/angular/styles_css.md
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Ionic Dark Theme
+ * -----------------------------------------------------
+ * For more information, please see:
+ * https://ionicframework.com/docs/theming/dark-mode
+ */
+
+/* @import '@ionic/angular/css/themes/dark.always.css'; */
+/* @import "~@ionic/angular/css/themes/dark.class.css"; */
```
sean-perkins, line 37: Remove the leading `~` here too.

sean-perkins (changes requested):
Docs look fantastic, great work! Two really minor changes for the commented out code to help avoid confusion for the Angular developers.

sean-perkins (approved):

static/usage/v8/theming/always-dark-mode/angular/styles_css.md
brandyscarney: I swear I did a find and replace to fix these and I don't know how these got missed lol. Fixed now: https://github.com/ionic-team/ionic-docs/blob/FW-5788/static/usage/v8/theming/always-dark-mode/angular/styles_css.md

static/usage/v8/theming/system-dark-mode/angular/styles_css.md
brandyscarney: Fixed https://github.com/ionic-team/ionic-docs/blob/FW-5788/static/usage/v8/theming/system-dark-mode/angular/styles_css.md

PR #3388, opened 2024-01-16
chore(playgrounds): update demos to use latest major version 

brandyscarney:
Some of our demos are using dev builds still, this updates them to the latest major version.


mapsandapps (approved):
good catch 👍🏻

PR #3404, opened 2024-01-26
chore(node): upgrade to v20

thetaPC:
Upgrading to be the same as the Vercel setting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "engines": {
-    "node": ">=18.0.0"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "engines": {
-    "node": ">=18.0.0"
+    "node": ">=20.0.0"
```
thetaPC, line 6: We don't need to have this key value set since it would follow whatever is set in Vercel. I left it in so enforce the node version locally.

sean-perkins (approved):

PR #3412, opened 2024-01-29
fix(docs): version the common css and js files 

brandyscarney:
I separated this out into several commits to make it easier to review:

- https://github.com/ionic-team/ionic-docs/pull/3412/commits/a0883961b68468beb6c38dda469ccf81cccb4aa2: versions the `common.css` and `common.js` and reverts them for `v6` and `v7` to match what is on `main`. The `v8` version of `common.css` in this commit has the `light` theme colors added back that I removed in [this commit](https://github.com/ionic-team/ionic-docs/commit/50546f5cb5bec29a6cce25aebaf8ce9dba2863ac#diff-8d3c5aa578302182fa1dd48fa734e5d9a484e7c8245088a13417d4d8dc590f41)
- https://github.com/ionic-team/ionic-docs/pull/3412/commits/ba594ac515a50187a8b288dab7f405553603c965: Removed the light colors that were added in the previous commit. I thought I broke them but it turns out our demos are just pointing to `v7` and not a dev build. If you change a demo, such as [button colors](https://github.com/ionic-team/ionic-docs/blob/50546f5cb5bec29a6cce25aebaf8ce9dba2863ac/static/usage/v8/button/theming/colors/demo.html#L9-L10) to point to `7.6.7-dev.11706551842.118ce6e0` and then view the colors in `v8` you will see that these do not need to be defined in `common.css`, so my original commit on `feature-8.0` was correct
- https://github.com/ionic-team/ionic-docs/pull/3412/commits/1c0600a6d0839976481e1d770cfb93ead2878c84: Updates the paths to point to the new versioned path for `common.css` and `common.js`. Due to the presence of nested demos, I had to perform the following find-and-replace operations in this specific order to ensure that I was only changing every path up one level:

    | Find | Replace |
    | ---| ---|
    | `href="../../../common.css` | `href="../../common.css` |
    | `href="../../../../common.css` | `href="../../../common.css` |
    | `src="../../../common.js"` | `src="../../common.js"` |
    | `src="../../../../common.js"` | `src="../../../common.js"` |  

:warning: I chose to only version these on `feature-8.0` because `v6` and `v7` were using the same files so this change is only needed to support the updates to `v8` theme colors.


mapsandapps (approved):
LGTM

PR #3415, opened 2024-01-30
docs(react, vue): update Quickstart page to use modern checkbox syntax

### LINKED ISSUE #3327: content: React/Vue Quickstart guides use legacy form control syntax
### URL

https://ionicframework.com/docs/react/quickstart, https://ionicframework.com/docs/vue/quickstart

### Issue Description

Under "A component with style," both pages have code snippets that use the legacy syntax for `ion-checkbox` (see [checkbox docs page](https://ionicframework.com/docs/api/checkbox#migrating-from-legacy-checkbox-syntax) for more info).

The surrounding content in the section also references the syntax and will need updating as well; examples:
- The checkbox is currently slotted in an `ion-item`, which is used to explain the `slot` property. The explanation may need to be changed to refer to the `ion-badge` instead.
- On the React page, the checkbox is referred to as not having any children, but this will change if the label text is moved into the default slot.

There may be other nearby things that need tweaking. The entire section should be reviewed to ensure everything still makes sense.

amandaejohnston:
Resolves #3327

Shortcuts:
- https://ionic-docs-git-fw-5781-ionic1.vercel.app/docs/react/quickstart#a-component-with-style
- https://ionic-docs-git-fw-5781-ionic1.vercel.app/docs/vue/quickstart#a-component-with-style

These pages are currently using legacy form control syntax for the checkbox examples. Some of the surrounding content also needed updating to keep things cohesive.

### DIFF
--- a/docs/react/quickstart.md
+++ b/docs/react/quickstart.md
@@ -171,7 +171,7 @@ When creating your own pages, don't forget to have `IonPage` be the root compone
 Our current content is relatively simple but does not contain anything that could be used in a real app, so let's change that.
 
 :::note
-For brevity, we're excluding repeating part of our component, like the function declaration or import statements for other components.
+For brevity, we're excluding repeating parts of our component, like the function declaration or import statements for other components.
 :::
 
 ```tsx
@@ -180,11 +180,10 @@ For brevity, we're excluding repeating part of our component, like the function 
   <IonContent>
     <IonList>
       <IonItem>
-        <IonCheckbox slot="start" />
-        <IonLabel>
+        <IonCheckbox labelPlacement="end" justify="start">
           <h1>Create Idea</h1>
           <IonNote>Run Idea by Brandy</IonNote>
-        </IonLabel>
+        </IonCheckbox>
         <IonBadge color="success" slot="end">
           5 Days
         </IonBadge>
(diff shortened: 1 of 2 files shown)


docs/react/quickstart.md
```diff
@@ -171,7 +171,7 @@ When creating your own pages, don't forget to have `IonPage` be the root compone
 Our current content is relatively simple but does not contain anything that could be used in a real app, so let's change that.
 
 :::note
-For brevity, we're excluding repeating part of our component, like the function declaration or import statements for other components.
+For brevity, we're excluding repeating parts of our component, like the function declaration or import statements for other components.
```
amandaejohnston, line 174: Long-standing typo.

docs/react/quickstart.md
```diff
@@ -198,20 +197,17 @@ Here in our `IonContent`, we're adding an `IonList` and a much more involved `Io
...
   <IonBadge color="success" slot="end">
     5 Days
   </IonBadge>
 </IonItem>
 ```
 
-Item is important as it clearly shows the mix of React concepts and Web Component concepts. The first clear example of a React concept is self-closing tags for React Components in `IonCheckbox`. This is just a simpler way of writing components that do not contain any child content.
-
-From the Web Components side, we have a special attribute called `slot`. This is key for letting the `IonItem` know where to place the `IonCheckbox` when it renders. This is not a React API, but a web standards API.
+Item is important as it clearly shows the mix of React concepts and Web Component concepts. From the Web Components side, we have a special attribute called `slot`. This is key for letting the `IonItem` know where to place the `IonBadge` when it renders. This is not a React API, but a web standards API.
```
amandaejohnston, line 210: I couldn't figure out a way to keep the idea of demonstrating a React-specific concept without it sounding forced. I'm open to also removing the "clearly shows a mix of React concepts and Web Component concepts" sentence; the Vue page doesn't have it, but I wasn't sure whether the idea was important to mention.

sean-perkins: I don't think there is a lot of value to stating that `Item` shows React concepts. We can somewhat assume a developer using React with Ionic is at least familiar with React at a basic level. 

The concept of slots is unfamiliar to a React developer. They either 'slot' content as children or use a property with a JSX binding to render additional mark-up in a targeted area. 

I'd be in favor of cutting this information down in the quick start to focus more on the `slot` specifics. Something similar to:

> Item uses a feature similar to `children` or JSX props in React, called [`slots`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot). This feature allows developers to let `IonItem` know where to render the `IonBadge` in the template. This is a web standards API that is used across many Ionic Framework components. 

What do you think?

amandaejohnston: Sounds good to me! I tweaked things in https://github.com/ionic-team/ionic-docs/pull/3415/commits/eeed7748d37d2bc48465a33318a399fdd5aac600.

PR #3435, opened 2024-02-09
docs(angular): update swiper events

santiagommejia:
The documentation is outdated, now the methods will only be triggered if they are prefixed with the word 'swiper'.

### DIFF
--- a/docs/angular/slides.md
+++ b/docs/angular/slides.md
@@ -256,27 +256,27 @@ To migrate, we would change the name of the event to `slidechange`:
 
 Below is a full list of event name changes when going from `ion-slides` to Swiper Angular:
 
-| ion-slides Event          | Swiper Event                 |
-| ------------------------- | ---------------------------- |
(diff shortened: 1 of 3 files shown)


santiagommejia:
> Hey there! Apologies for the delay. I appreciate you making this PR. Would you be able to make these changes on https://ionicframework.com/docs/react/slides and https://ionicframework.com/docs/vue/slides too?

Hey there! no worries. I tested it out on vue and react and it doesn't seem to be the case, the swiper prefix only applies to Angular.

liamdebeasi:
Thanks for checking! It looks like you are correct. According to the [documentation](https://swiperjs.com/swiper-api#param-eventsPrefix), the prefixed events only apply to Swiper Element (the Web Component). Our Vue/React migration guides still use the React/Vue components which is why the events do not apply there.

liamdebeasi:
Merged. Thank you!

Issue #3096, opened 2023-08-24
feat(ja): reflect translation of core.json


rdlabo:
### Describe Problem

The `translation/jp` branch also translates core.json, and the preview screen shows the translated version. But https://ionicframework.com/docs/ja/ shows untranslated ones.

![スクリーンショット 2023-08-24 午前11 33 28](https://github.com/ionic-team/ionic-docs/assets/9690024/c353b3d0-40c7-435d-b245-3b0c62eb6860)

![スクリーンショット 2023-08-24 午前11 33 43](https://github.com/ionic-team/ionic-docs/assets/9690024/ceeda318-da43-4293-b4e9-31cc73a551c8)

### Describe Preferred Solution

In `plugins/docusaurus-plugin-ionic-component-api/index.js` , changes in translation/jp should be reflected when displayed at https://ionicframework.com/docs/ja/.

main branch:
https://github.com/ionic-team/ionic-docs/blob/main/plugins/docusaurus-plugin-ionic-component-api/index.js#L22

translation/jp branch
https://github.com/ionic-team/ionic-docs/blob/translation/jp/plugins/docusaurus-plugin-ionic-component-api/index.js#L23

PR #3437, opened 2024-02-11
feat(translation/jp): 20240211 update

rdlabo:


scripts/api-ja.js
```diff
@@ -0,0 +1,236 @@
+// const fetch = require('node-fetch');
```
liamdebeasi, line 1: Can we remove this file if we don't need it?

liamdebeasi (changes requested):

scripts/api-ja.js
rdlabo: Yes, it can now be deleted. It can also be removed from the current build script.

liamdebeasi: Sorry, I should have asked this originally, but can you clarify why we don't need this script anymore? In `main` it looks like we still use it: https://github.com/ionic-team/ionic-docs/blob/e9c3e62bb83a13392145d5f2d836b4925e6a8540/scripts/i18n.sh#L8

rdlabo: Yes, remove from main branch will not effect. Files no longer in use now.

However, this issue needs to be discussed: https://github.com/ionic-team/ionic-docs/issues/3096

liamdebeasi: Thanks for the clarification!

liamdebeasi (approved):
Thank you!

PR #3481, opened 2024-02-28
chore(jp): migrate script to mjs

sean-perkins:
When running the build process in CI after merging in main, an exception will be thrown:
```
Error: Cannot find module './utils'
```

This was due to the `api-jp.js` script not being migrated to mjs when we upgraded to Docusaurus 3. This PR migrates that script and usage around accessing the `.readme` which no longer exists in the metadata since we moved component documentation into this repository.


brandyscarney:
+1

brandyscarney (approved):

PR #3482, opened 2024-02-28
docs: archive v5 docs

liamdebeasi:
We are getting out of memory errors on deployment because all sites are being rebuilt and loaded into memory on deployment. As a result, we are going to try archiving the v5 docs as [recommended on the Docusuarus website](https://docusaurus.io/docs/versioning#keep-the-number-of-versions-small). This will keep the v5 docs available but will avoid Docusaurus having to rebuild the v5 docs on every deployment.


docusaurus.config.js
```diff
@@ -4,6 +4,22 @@ const fetch = require('node-fetch');
...
 const VERSIONS_JSON = require('./versions.json');
 
+/**
+ * Old versions of the Ionic Docs are archived so
+ * that we do not need to re-build it every time we deploy.
+ * Building a large number of docs sites at once can cause
+ * out of memory issues, so archiving old docs sites
+ * allow us to keep memory usage and build times low.
+ */
+const ARCHIVED_VERSIONS_JSON = require('./versionsArchived.json');
```
thetaPC, line 14: Can we also add a note that this JSON file is only for versions that live in the site? Else we might be questioning why v4 and v3 are not included.

thetaPC (approved):
LGTM, with one suggestion.

I did notice that if I refresh the page on v5 then I get a Page Not Found for a split second. The same if I navigate to v5 for the first time. But this can be looked into at a later date.

liamdebeasi:
Yeah that's interesting.. I'm getting the same thing on the production website: https://ionicframework.com/docs/v5/ so at least we know it's not related to these changes.

docusaurus.config.js
liamdebeasi: Good idea! Added in https://github.com/ionic-team/ionic-docs/pull/3482/commits/914bc7ef132de0201831d39f04619d10ef98030e

PR #3483, opened 2024-02-28
docs: add correct breaking changes url

liamdebeasi:
The v8 breaking changes are available on the `feature-8.0` branch, so we need to link there until v8 is merged into `main`.


thetaPC (approved):
LGTM

PR #3484, opened 2024-02-28
chore: remove trailing slash to avoid flash of 404 page

liamdebeasi:
Jared told me that to avoid the flash of the 404 page when going to the v5 docs you need to link to it without the trailing slash.


thetaPC (approved):
LGTM, works great!

Suggestion: Add a note to the Docusaurus config saying that if the Page not found flashes then it's b/c of the trailing slash. I'm sure we'll forget this down the line.

liamdebeasi:
Good idea! Added in https://github.com/ionic-team/ionic-docs/pull/3484/commits/c19efff89c26d4ebc25d25480eb5316eb81c2b19

PR #3514, opened 2024-03-08
docs(vue): update quickstart to reference the correct file

nvispute:
Update the Non existing file name to latest existing file.

- Old code used to be shipped with  `shims-vue.d.ts`, Newer Code ships with `vite-env.d.ts`  pointer to non-existent file might Create confusion with beginner levels


brandyscarney (approved):
Thank you for the PR and sorry for the delayed review! 🙂

PR #3521, opened 2024-03-13
docs: rename themes to palettes

sean-perkins:
Renames the dark theme and high contrast themes to palettes. 

The demos will not load the correct stylesheet until https://github.com/ionic-team/ionic-framework/pull/29149 is merged and released.


static/usage/v8/theming/always-high-contrast-mode/react/main_tsx.md
```diff
@@ -22,7 +22,7 @@ import { personCircle, personCircleOutline, sunny, sunnyOutline } from 'ionicons
 function Example() {
   return (
     <>
-      <IonHeader class="ion-no-border">
+      <IonHeader>
```
liamdebeasi, line 25: These demos likely always had the class, but I decided to remove them while I was in here. The iframe demo doesn't have the class, and the Stackblitz demos should match the iframe demos in terms of visual result

liamdebeasi (approved):

PR #3522, opened 2024-03-13
docs: note event bubbling

liamdebeasi:
I've [seen](https://github.com/ionic-team/ionic-framework/issues/27724) [a](https://github.com/ionic-team/ionic-framework/issues/27986) [few](https://github.com/ionic-team/ionic-framework/issues/28794) instances of developers getting confused about the `ionChange` handler on accordion groups firing when a descendant component such as an input/checkbox has their value changed. This is happening because `ionChange` events bubble, so when Input emits `ionChange` that event bubbles up to the accordion group and causes the `ionChange` callback there to fire.

I think it's worth calling this behavior out on the docs. I made the following changes:

1. Updated the API generation script to note which events bubble.
2. Updated the accordion docs for v7 and v8 to specifically call out this behavior. Most components that emit `ionChange` do not typically have descendants that also fire `ionChange`, so accordion is a bit of an outlier here.

### DIFF
--- a/plugins/docusaurus-plugin-ionic-component-api/index.js
+++ b/plugins/docusaurus-plugin-ionic-component-api/index.js
@@ -168,11 +168,9 @@ function renderEvents({ events }) {
   }
 
   return `
-| Name | Description |
-| --- | --- |
-${events.map((event) => `| \`${event.event}\` | ${formatMultiline(event.docs)} |`).join('\n')}
-
-`;
+| Name | Description | Bubbles |
+| --- | --- | --- |
+${events.map((event) => `| \`${event.event}\` | ${formatMultiline(event.docs)} | \`${event.bubbles}\` |`).join('\n')}`;
 }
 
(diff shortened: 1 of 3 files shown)


plugins/docusaurus-plugin-ionic-component-api/index.js
```diff
@@ -168,11 +168,9 @@ function renderEvents({ events }) {
...
 
   return `
-| Name | Description |
-| --- | --- |
-${events.map((event) => `| \`${event.event}\` | ${formatMultiline(event.docs)} |`).join('\n')}
-
-`;
+| Name | Description | Bubbles |
+| --- | --- | --- |
+${events.map((event) => `| \`${event.event}\` | ${formatMultiline(event.docs)} | \`${event.bubbles}\` |`).join('\n')}`;
```
mapsandapps, line 173: neat!

mapsandapps (approved):

PR #3533, opened 2024-03-19
docs(your-first-app): use correct type on "Adding Mobile" page

Nozarno:
Editing the Typing to Fix a TypeScript Error

PR #3534, opened 2024-03-19
Update 5-adding-mobile.md

Nozarno:
Editing the Typing to Fix a TypeScript Error


brandyscarney:
Thank you for the PR! It looks like this fix was added in with https://github.com/ionic-team/ionic-docs/pull/3533 so I am going to close this. 🙂

PR #3543, opened 2024-03-21
fix: rename themes to palettes in playground imports

mapsandapps:
The stackblitz v8 playgrounds are broken right now: https://ionicframework.com/docs/v8/api/button

I also updated the guide for migrating to v8 to use the new paths.


liamdebeasi (approved):
thank you!

amandaejohnston (approved):
Thank you!

PR #3578, opened 2024-04-04
Add another step If the project is using Jest with jest-preset-angular

orl99:
We have to add this configuration because there are now several projects using Jest instead of Karma and Jasmine, I had to spend hours investigating until I found the solution, this could have been avoided If on the documentation we had explained this before.


thetaPC:
@orl99 Thank you for the pull request!

Could you share a minimal reproduction? Ionic 6 should not require CJS transpiling as it includes a CJS version of Ionic Angular. It would help to understand why this step was necessary.

thetaPC:
Closing this due to the lack of a response. Feel free to open another PR with an answer to my [question](https://github.com/ionic-team/ionic-docs/pull/3578#issuecomment-2050129457). Thank you!

PR #3585, opened 2024-04-09
docs(title): add information on setting headings

liamdebeasi:
As part of this ticket the team wanted to provide some information on how to indicate the heading level of an IonTitle to assistive technology. This is part of https://github.com/ionic-team/ionic-framework/issues/23746. We originally tried to have IonTitle automatically determine this, but doing so lead to brittle code with a lot of edge cases to consider. As a result, we felt it would be best to let developers handle this until we can collect more data/feedback on how IonTitle is being used.

Preview: https://ionic-docs-git-fw-119-ionic1.vercel.app/docs/api/title#headings


thetaPC (approved):
LGTM

PR #3586, opened 2024-04-09
docs(migration): add info about removing old light palette colors

amandaejohnston:
While going through the migration guide, I noticed that we don't call out having to remove the old light palette colors in order for the updated colors to appear.

### DIFF
--- a/docs/updating/8-0.md
+++ b/docs/updating/8-0.md
@@ -86,7 +86,7 @@ Previous versions shipped a set of default color variables for the light theme:
   --ion-color-primary-rgb: 56, 128, 255;
   --ion-color-primary-contrast: #ffffff;
   --ion-color-primary-contrast-rgb: 255, 255, 255;
-  ...
+  /* ... */
 }
 ```
 
--- a/docs/updating/8-0.md
+++ b/docs/updating/8-0.md
@@ -90,7 +90,9 @@ Previous versions shipped a set of default color variables for the light theme:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/updating/8-0.md
```diff
@@ -74,6 +74,26 @@ npm install @ionic/core@next
...
+  /** primary **/
+  --ion-color-primary: #3880ff;
+  --ion-color-primary-rgb: 56, 128, 255;
+  --ion-color-primary-contrast: #ffffff;
+  --ion-color-primary-contrast-rgb: 255, 255, 255;
+  ...
+}
+```
+
+If you are not using a custom color scheme, these color variables should all be removed. In Ionic Framework version 8, the default colors have been updated to improve accessibility. The new palette is also automatically included with all Ionic apps. Removing the old color variables ensures that the latest palette is not overwritten.
```
liamdebeasi, line 93:
```suggestion
In Ionic Framework version 8, the default color palette is included as long as `core.css` is imported. Removing the old color variables ensures that the latest palette is not overwritten.

Developers who are customizing this color palette can continue to keep the custom variables values, but any of the variables that use the default values should be removed.
```

Optional: I changed a bit of the wording. Mentioning the new a11y improvements isn't required to make this change, so I opted to get rid of it. I also wanted to call out the relationship between the light theme and the `core.css` file.

liamdebeasi (approved):
One optional change, but otherwise looks good.

docs/updating/8-0.md
amandaejohnston: Works for me! I was a bit unsure of the wording; agree that this is much cleaner.

PR #3587, opened 2024-04-09
docs(high-contrast): fix typo in Angular dark palette import

amandaejohnston:
1. Fixes the dark palette import to use the right file name (previously the light and dark palette imports were the same)
2. Changes the code block to `scss` so the comment renders properly


thetaPC (approved):
LGTM

PR #3589, opened 2024-04-10
docs(modal): clarify changing height of sheet modal

liamdebeasi:
We've seen some confusion in https://github.com/ionic-team/ionic-framework/issues/24583 where developers are trying to adjust the height of the sheet modal by changing the max breakpoint. This causes issues because it means some of the modal (while visually shorter) is inaccessible outside of the viewport.

I noticed we had a playground that showed how to do this properly, but it did not show up in the sidemenu because of it being a level 4 heading. As a result, I made the following changes:

1. Removed the "Types of Modals" heading and made Card/Sheet Modal headings go from level 3 to level 2 headings
2. Moved the subheadings in Sheet Modal from level 4 to level 3 headings so they show up in the sidebar
3. Modified the "Auto Sheet Height" title to make it more related to changing the height of the modal rather than a special kind of height modification. (The playground remains unchanged).
4. Added content to the section noted in item 3 to explicitly call out that developers should use `--height` to change the height of the sheet modal. I also provided context on why changing the max breakpoint should be avoided.

### DIFF
--- a/docs/api/modal.md
+++ b/docs/api/modal.md
@@ -98,10 +98,8 @@ Here's a simplified example illustrating how a child component can interact with
 import CanDismissChildStateExample from '@site/static/usage/v8/modal/can-dismiss/child-state/index.md';
 
 <CanDismissChildStateExample />
-
-## Types of modals
  
-### Card Modal
+## Card Modal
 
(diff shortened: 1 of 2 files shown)


thetaPC (approved):
LGTM

PR #3594, opened 2024-04-12
docs(button): update shape and icons playgrounds to showcase circular icon buttons

brandyscarney:
Preview: https://ionic-docs-git-fw-5965-ionic1.vercel.app/docs/v8/api/button

I didn't think it made sense to add an entirely new playground for this since the circular button can now be achieved by setting `shape="round"` on a button with only an icon. Instead, I updated the `Shape` playground to include buttons containing icons, and the `Icons` playground to include buttons containing icons at different sizes. I wasn't sure how much overlap we wanted in the playgrounds since a `Size` playground also exists. Happy to follow a different approach if anyone has other ideas.


amandaejohnston (approved):
Arranging things this way works for me 👍

PR #3597, opened 2024-04-15
add examples for setting an initial value when using maskito (via maskitoTransform)

marlon-ionic:
Includes changes to the input docs (input-masking section) to show an example of how to set an initial value with Maskito via maskitoTransform (based on their docs https://maskito.dev/core-concepts/transformer)

### DIFF
--- a/static/usage/v7/input/mask/demo.html
+++ b/static/usage/v7/input/mask/demo.html
@@ -39,10 +39,13 @@
       async function initPhoneMask() {
         const ionInput = document.querySelector('#phone');
         const nativeEl = await ionInput.getInputElement();
-
-        new window.Maskito(nativeEl, {
+        const phoneMaskOptions = {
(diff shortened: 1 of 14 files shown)


static/usage/v7/input/mask/demo.html
sean-perkins, line 1: Can you double check your changes for the demo preview for the v7 and v8 playgrounds?

They do not appear to set the default value as they do when opened in Stackblitz.

https://ionic-docs-git-feat-maskito-initial-value-ionic1.vercel.app/docs/api/input#input-masking

In the console you will observe an error:
```
TypeError: window.maskitoTransform is not a function
    at initPhoneMask
```

sean-perkins (changes requested):

static/usage/v7/input/mask/demo.html
marlon-ionic: Sorry for that - I didn't realize demo.html had its own Maskito reference

sean-perkins (approved):
Great work 👍

PR #3598, opened 2024-04-16
Mention other popular input masking libraries.

oliveryasuna:


sean-perkins:
Hello @oliveryasuna thank you for this suggested change.

The Ionic Framework team did extensive research into input masking solutions and found inconsistency and a lack of support for web components with many of them. 

Maskito is the only library we have full confidence in recommending as a preferred solution for input masking with Ionic Framework components. We have worked closely with their team and they have been committed to providing necessary features to improve the compatibility. 

Developers can use alternative solutions and may find success with them, but our team does not feel confident recommending alternatives that we have not vetted or believe to be a better solution.

PR #3663, opened 2024-05-21
docs(vue): remove the coming soon page

thetaPC:
There are no plans to update this page. It's been decided to remove it entirely.

### DIFF
--- a/docs/vue/testing.md
+++ b/docs/vue/testing.md
@@ -1,3 +1,0 @@
-# Testing
-
-Coming Soon!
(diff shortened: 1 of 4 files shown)


brandyscarney:
Should we remove this for v5 and v6 too? https://ionic-docs-o31kiyk8l-ionic1.vercel.app/docs/v6/vue/testing

thetaPC:
@brandyscarney I thought about that but I'm not sure if it would cause any issues since those are archived. Thoughts?

brandyscarney:
Ah true. Okay let's ignore those!

brandyscarney (approved):

PR #3666, opened 2024-05-22
fix(redirects): add redirects for v8 links

jaredcbaum:
This PR will redirect pages formerly hosted under v8 to the main site (since v8 is now the latest)

**It will need to be reverted once v8 is archived**


thetaPC (approved):
LGTM

PR #3674, opened 2024-05-28
feat: segment css properties by mode

### LINKED ISSUE #3668: bug: CSS Custom Properties present 2x in a tables
### Describe the Bug

Pages have all CSS Custom Properties present 2x in a tables for v8 docs

### Expected Behavior

Each CSS Custom property should be present only once in a table

### Steps to Reproduce

Open [Split Pane > CSS Custom Properties](https://ionicframework.com/docs/api/split-pane#css-custom-properties-1), see

| Name | Description |
| -- | -- |
| --border | Border between panes |
| --border | Border between panes |
| --side-max-width | Maximum width of the side pane. Does not apply when split pane is collapsed. |
| --side-max-width | Maximum width of the side pane. Does not apply when split pane is collapsed. |
| --side-min-width | Minimum width of the side pane. Does not apply when split pane is collapsed. |
| --side-min-width | Minimum width of the side pane. Does not apply when split pane is collapsed. |
| --side-width | Width of the side pane. Does not apply when split pane is collapsed. |
| --side-width | Width of the side pane. Does not apply when split pane is collapsed. |

### Screenshots

_No response_

### Operating System

N/A

### Browser

Chrome

### Version

_No response_

### Additional Information

_No response_

sean-perkins:
Resolves #3668

--------------------

Adds an iOS or MD tab if the component has css properties available for a specific mode, otherwise falls back to the existing UI.

This uses MDX features which required updating the generated imports for all components.

### DIFF
--- a/docs/api/accordion-group.md
+++ b/docs/api/accordion-group.md
@@ -5,7 +5,7 @@ import Props from '@ionic-internal/component-api/v8/accordion-group/props.md';
 import Events from '@ionic-internal/component-api/v8/accordion-group/events.md';
 import Methods from '@ionic-internal/component-api/v8/accordion-group/methods.md';
 import Parts from '@ionic-internal/component-api/v8/accordion-group/parts.md';
-import CustomProps from '@ionic-internal/component-api/v8/accordion-group/custom-props.md';
+import CustomProps from '@ionic-internal/component-api/v8/accordion-group/custom-props.mdx';
 import Slots from '@ionic-internal/component-api/v8/accordion-group/slots.md';
 
 import EncapsulationPill from '@components/page/api/EncapsulationPill';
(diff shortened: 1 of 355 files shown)


brandyscarney:
Do we want them to be combined if they are the same? https://ionic-docs-git-sp-duplicate-css-props-ionic1.vercel.app/docs/api/alert#css-custom-properties

sean-perkins:
I don't think so, since v9 is going to make that behavior obsolete with components potentially having their own styling APIs. The goal is to show developers which properties are available for a mode. It is ok if they are the same list of variables in both, that just means the component does not have any mode-specific properties. 

If we wanted to have a "shared" tab for common properties and then tabs for only the specific mod props we can, but that would need to be put in your team's backlog 😂

brandyscarney (approved):

PR #3678, opened 2024-06-02
docs(theming): replace charcoal variable with background since the property name is background

kgsatish:
The `getPropertyValue ` method incorrectly takes `--charcoal` as the input, it should be `--background`


brandyscarney (approved):

brandyscarney:
Thank you for the PR!

PR #3722, opened 2024-06-28
docs(demos): update v8 playgrounds to use correct version

brandyscarney:
This does two things:
1) Update the v8 playgrounds that were still using v7 to use the proper version
2) Update the syntax to use `""` instead of `{}` on the versions because these were skipped in a find & replace for that reason


sean-perkins (approved):
Thank you, you are the best.

thetaPC (approved):
LGTM

PR #3734, opened 2024-07-11
docs(updating): set versions to force major number

lil5:
With the advent of react 18 this documentation ages badly, forcing a major version number ensures this continues to work as specifies by the text above it


brandyscarney (approved):

brandyscarney:
Thank you for the PR!

PR #3737, opened 2024-07-17
docs(button): update icon playgrounds to use addIcons usage

brandyscarney:

[Preview](https://ionicframework.com/docs/api/button)  

## What is the new behavior?

We no longer recommend using `<ion-icon>` with a string in JavaScript or Angular, unless you add it using `addIcons`. This updates the playgrounds for Button to show that.

[Preview](https://ionic-docs-git-docs-button-icon-usage-fixes-ionic1.vercel.app/docs/api/button)


thetaPC (approved):
LGTM

PR #3743, opened 2024-07-18
docs(item): update icon playgrounds to use addIcons usage

brandyscarney:

[Preview](https://ionicframework.com/docs/api/item)  

## What is the new behavior?

We no longer recommend using `<ion-icon>` with a string in JavaScript or Angular, unless you add it using `addIcons`. This updates the playgrounds for Item to show that.

Additionally, I removed the `item/icons/` playground examples for both v7 and v8 as these were not being used anymore. This demo does still exist in the v6 docs. 

[Preview](https://ionic-docs-git-docs-item-icon-usage-fixes-ionic1.vercel.app/docs/api/item)


thetaPC (approved):
LGTM

PR #3753, opened 2024-07-18
docs(textarea): update icon playgrounds to use addIcons usage

brandyscarney:

[Preview](https://ionicframework.com/docs/api/textarea)  

## What is the new behavior?

We no longer recommend using `<ion-icon>` with a string in JavaScript or Angular, unless you add it using `addIcons`. This updates the playgrounds for Textarea to show that.

Additionally, the Textarea docs for Start and End Slots in v8 were using the v7 playground, so I added a v8 one. The v7 docs were not including this section or the v7 demo and the demo was also referencing the wrong path to the common files, so I fixed that.

[Preview](https://ionic-docs-git-docs-textarea-icon-usage-fixes-ionic1.vercel.app/docs/api/textarea)


thetaPC (approved):
LGTM

PR #3766, opened 2024-07-22
docs(api): pass ionicons dependency in the javascript and angular package file

brandyscarney:
This PR removes the `ionicons` dependency in Angular and Javascript from the individual demos and places it in their `package.json` files as a dependency instead. 

I realized after completing most of the `addIcon` usage demos that I should have done this in the first place but it was easier to wait until all of them were merged and change them all at once.


BenOsodrac (approved):

thetaPC (approved):
LGTM, works without issues

thetaPC:
@brandyscarney there are conflicting files so the auto-merge is being blocked.

PR #3771, opened 2024-07-24
docs(native): properly type camera example usage

brandyscarney:


brandyscarney:
I created this because there were issues with the Vercel deployment on https://github.com/ionic-team/ionic-docs/pull/3175 but it ended up working so merging that PR instead.

PR #3797, opened 2024-08-18
Remove incorrect information about ion-tab-bar being optional

SaebAmini:
Attempting to use an ion-tabs without an ion-tab-bar results in an uncaught exception:

image


thetaPC:
Thank you for the PR!

I'm going to close it because we realized that only React had the requirement to have tab bar as a child of tabs. We decided to remove this requirement in this [PR](https://github.com/ionic-team/ionic-framework/pull/29868). Thank you for bring this to our attention.

Issue #3796, opened 2024-08-14
content: Inaccurate property for ion-input (Vue)


DubNubz:
### URL

https://ionicframework.com/docs/api/input#input-counter

### Issue Description

The `maxlength` property of `ion-input` expects a number, not a string.

The current Vue example provides maxlength as a string.
```vue
<!--current example-->
<ion-input ... maxlength="20"></ion-input>

<!--correct implementation-->
<ion-input ... :maxlength="20"></ion-input>

PR #3807, opened 2024-08-29
fix(input): vue counter example uses maxlength as a number

sean-perkins:
Resolves: #3796

Updates the Vue examples for the `ion-input` counter example to bind the `maxlength property as a number instead of a string.


brandyscarney (approved):

thetaPC (approved):
LGTM

PR #3961, opened 2024-12-27
docs(segment): update angular to standalone

brandyscarney:
Migrates v7 & v8 Angular playgrounds to standalone

[Preview](https://ionic-docs-git-rou-11416-segment-ionic1.vercel.app/docs/api/segment)

Note: The Segments in Toolbars demo will error because that playground is in the `toolbar` directory


gnbm (approved):
LGTM 🚀

PR #3989, opened 2025-01-06
docs(build-options): remove enableProdMode() reference

rtpHarry:
It's discouraged now: https://angular.dev/api/core/enableProdMode

"Using this method is discouraged as the Angular CLI will set production mode when using the optimization option."


brandyscarney (approved):

brandyscarney:
Thank you 🙂

PR #4064, opened 2025-04-03
Specified current major version

elcreator:
Docs v8 says it will install v8 but it installs v7 as v8 cli was never released.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -15,7 +15,7 @@ The Ionic command-line interface ([CLI](/docs/reference/glossary#cli)) is the go
 
 ## Installation
 
-The Ionic CLI can be installed globally with npm:
+The Ionic CLI 7 can be installed globally with npm:
 
 ```shell
 npm install -g @ionic/cli


ShaneK:
I don't see a reason for specifically calling out the Ionic CLI is v7 right now. The command provided installs the latest version of the Ionic CLI, which has no version association with Ionic or Capacitor. It was coincidentally the same major version as Ionic Framework for a bit, but that was more of a coincidence than anything done intentionally.

I believe this change would incur unnecessary maintenance in the future if the Ionic CLI releases another major version and I don't see any benefit in mentioning it here.

elcreator:
Including at least statement about different versions of CLI and Ionic will be helpful.
image
As many Angular developers prefer Ionic they may assume the CLI version is bigger (as in Angular) or equal (as in many other npm packets) but not lower than version of thing it is intended to install.

ShaneK (approved):

brandyscarney (approved):

ShaneK:
Thank you for your contribution, congratulations on your first pull request to Ionic Docs! Your PR has been [merged and deployed](https://ionicframework.com/docs/cli#troubleshooting).

PR #4079, opened 2025-04-15
docs(navigation): correct routerDirection values

brandyscarney:
RouterDirection does not contain `none`: https://github.com/ionic-team/ionic-framework/blob/8dd566b5c1241022e26cc91c0f415de20c0d0f34/core/src/components/router/utils/interface.ts#L66

When using with Vue it gives this error:

```
Type '"none"' is not assignable to type 'RouterDirection | undefined'
```

While it might work for React, I think this is incorrect: https://github.com/ionic-team/ionic-framework/blob/8dd566b5c1241022e26cc91c0f415de20c0d0f34/packages/react/src/models/RouterDirection.ts#L1


thetaPC (approved):
LGTM

PR #4140, opened 2025-06-07
translation(jp): 20250607 update

rdlabo:
merge main branch, and translate input-otp


ShaneK (approved):
Tested to make sure it builds and runs correctly, looks like it's working great. Thanks for the update, we appreciate it!

PR #4221, opened 2025-07-30
feat(angular): add injection token docs

ShaneK:
In 8.7.0, we're adding [injection token](https://angular.dev/api/core/InjectionToken) support for ion-modals in Angular. This documentation helps explain what they are and how to use them. 

I created a whole page for injection tokens in case we add further injection token support in the future.

[Preview page](https://ionic-docs-git-docs-angular-injection-token-ionic1.vercel.app/docs/angular/injection-tokens)


brandyscarney (approved):
LGTM!

PR #4231, opened 2025-08-03
translation(jp): 20250803 update

rdlabo:
Since there were a bunch of important changes, I went ahead and merged it. I'll do a proper translation when I have time.


brandyscarney (approved):
Thanks so much! 🙂

PR #4263, opened 2025-09-04
docs(playgrounds): upgrade to typescript 5.9.0

brandyscarney:
Relates to ionic-team/ionic-framework#30650

This PR does the following:
- Removes the `package-lock.json` files from the playgrounds, StackBlitz will generate this file from the `package.json`
- Updates Typescript to `~5.9.0` to avoid updating to minor versions automatically
- Adds `"skipLibCheck": true` to the Angular `tsconfig` to silence the `Interface 'HTMLIonInputElement' cannot simultaneously extend types 'IonInput' and 'HTMLStencilElement'` errors so they will work with `5.9`


gnbm (approved):
LGTM 🚀

PR #4281, opened 2025-10-13
docs(tabs): add playground for programmatically selecting tabs

brandyscarney:
In addition to the [Vue Quickstart](https://ionicframework.com/docs/vue/quickstart#call-component-methods) which outlines how to call component methods, this adds a playground specifically for tabs methods, resolving #827.

I also updated the Basic tests for Tabs to include `ion-page` (either as a class or component). It's hard to tell in the basic examples but when you add more content you can tell that without it the content does not take up the full height:

https://github.com/user-attachments/assets/9583141d-d8ab-4361-ab89-27c000b6eb57

[Preview](https://ionic-docs-git-fw-6803-ionic1.vercel.app/docs/api/tabs#programmatically-selecting-tabs)

### DIFF
--- a/static/usage/v8/tabs/basic/vue.md
+++ b/static/usage/v8/tabs/basic/vue.md
@@ -26,8 +26,7 @@
       </ion-page>
     </ion-tab>
     <ion-tab tab="library">
-      <ion-page id="library-page"
-        >>
+      <ion-page id="library-page">
         <ion-header>
           <ion-toolbar>
             <ion-title>Library</ion-title>
@@ -39,8 +38,7 @@
       </ion-page>
     </ion-tab>
     <ion-tab tab="search">
-      <ion-page id="search-page"
-        >>
+      <ion-page id="search-page">
         <ion-header>
           <ion-toolbar>
             <ion-title>Search</ion-title>


static/usage/v8/tabs/basic/vue.md
```diff
@@ -33,10 +23,11 @@
         <ion-content>
           <div class="example-content">Radio content</div>
         </ion-content>
-      </div>
+      </ion-page>
     </ion-tab>
     <ion-tab tab="library">
-      <div id="library-page">
+      <ion-page id="library-page"
+        >>
```
thetaPC:
```suggestion
      <ion-page id="library-page">
```

static/usage/v8/tabs/select-method/angular/example_component_css.md
thetaPC, line 1: Might be beneficial if we can have a way to hide these types of files from rendering on the pages. It would be a nice quality of life feature.

brandyscarney: What do you mean? I don't think we should hide the styles that give them the same look as the demo example.

static/usage/v8/tabs/basic/vue.md
brandyscarney: https://github.com/ionic-team/ionic-docs/pull/4281/commits/1fbf0e913d144091323f786d791c741f13b2f57e

static/usage/v8/tabs/select-method/angular/example_component_css.md
thetaPC: You don't think that it can lead to confusion when the styles are only needed for demonstration purposes?

brandyscarney: Well the styles are needed to make the text centered in the view: 

CleanShot 2025-11-18 at 11 13 30@2x

If we remove the styles then users will be confused why their apps don't match the demo.

thetaPC: Good point! Then disregard my initial comment.

PR #4323, opened 2025-11-14
feat(styles): add background to code block title

brandyscarney:
Updated the background for the code block title to match the code background. Inspired by the [Docusaurus examples](https://docusaurus.io/docs/markdown-features/code-blocks).

You can see an example of `title` being used on the React: Add to Existing guide ([Before](https://ionicframework.com/docs/react/adding-ionic-react-to-an-existing-react-project#using-individual-ionic-components) | [After](https://ionic-docs-git-feat-code-block-title-ionic1.vercel.app/docs/react/adding-ionic-react-to-an-existing-react-project#using-individual-ionic-components)).

| Before | After |
| --- | --- |
| Before: Light Mode | After: Light Mode |
| Before: Dark Mode | After: Dark Mode |

### DIFF
--- a/src/styles/components/_code.scss
+++ b/src/styles/components/_code.scss
@@ -51,12 +51,9 @@ code {
 // Overrides
 #__docusaurus {
   [class*='codeBlockContainer_'] {
-    box-shadow: none;
-  }
+    --prism-background-color: var(--code-block-bg-c);
 
-  // Code block title
-  [class*='codeBlockTitle_'] {
-    background: var(--code-block-bg-c);
+    box-shadow: none;
   }
 }
 
@@ -68,7 +65,6 @@ code {
 
 code[class*='language-'],
 pre[class*='language-'] {
-  background: var(--code-block-bg-c);
   white-space: pre;
   word-spacing: normal;
   word-break: normal;


src/styles/components/_code.scss
thetaPC: We can update the background based on the recommendation from [Docusaurus](https://docusaurus.io/docs/markdown-features/code-blocks#theming) using `--prism-background-color`.

I highly recommend to not accept the suggestion since it's being weird because I'm making a suggestion on code that wasn't changed in this commit. What I changed was:
- updated `[class*='codeBlockContainer_']` to use `--prism-background-color`
- removed `[class*='codeBlockTitle_']`
- removed `background: var(--code-block-bg-c)` from `code[class*='language-'], pre[class*='language-']` 

Fun fact: We can use `--prism-color` to change the color of the title.

```suggestion
  [class*='codeBlockContainer_'] {
    --prism-background-color: var(--code-block-bg-c);
    box-shadow: none;
  }
}

/**
* prism.js Dark theme for JavaScript, CSS and HTML
* Based on the slides of the talk “/Reg(exp){2}lained/”
* @author Lea Verou
*/

code[class*='language-'],
pre[class*='language-'] {
```

brandyscarney: Done: https://github.com/ionic-team/ionic-docs/pull/4323/commits/81556b24cbf8b2540c596aef6e566abe6add8525

PR #4286, opened 2025-10-17
docs(intro): update packages & CDN to redirect to other guides

brandyscarney:
Resolves #446 in combination with #4322

## What is the current behavior?

The [Packages & CDN guide](https://ionic-docs-2gw8ev7f0-ionic1.vercel.app/docs/intro/cdn) has some information on how to get started with Angular, React and Vue but these sections are incomplete. For example, we cover how to add React to an existing app [here](https://ionic-docs-2gw8ev7f0-ionic1.vercel.app/docs/intro/cdn#ionic--react) but it is missing a lot of steps that are documented in the guide [here](https://ionic-docs-2gw8ev7f0-ionic1.vercel.app/docs/react/adding-ionic-react-to-an-existing-react-project). It doesn't make sense to have the same information in two places.

## What is the new behavior?

This PR updates the Packages & CDN guide to redirect out to the `Quickstart` & `Add to Existing` guides for the packages, but keeps the CDN content in the Packages & CDN guide. React was the only framework that already had an `Add to Existing` guide, so this also adds `Add to Existing` guides for Angular & Vue and moves the content from the Packages & CDN guide to those.

### Ionic Packages & CDN Guide Changes

Below is a before & after of the CDN page (right-click and open the image in a new tab to see the images larger):

<table>
  <tr>
    <td valign="top"><a href="https://ionic-docs-2gw8ev7f0-ionic1.vercel.app/docs/intro/cdn">Before</a></td>
    <td valign="top"><a href="https://ionic-docs-git-docs-cdn-updates-ionic1.vercel.app/docs/intro/cdn">After</a></td>
  </tr>
  <tr>
    <td valign="top">production view of packages and cdn page</td>
    <td valign="top">branch view of packages and cdn page</td>
  </tr>
</table>

### Add to Existing Guide Changes

**React**
- [Before](https://ionic-docs-2gw8ev7f0-ionic1.vercel.app/docs/react/adding-ionic-react-to-an-existing-react-project) | [After](https://ionic-docs-git-docs-cdn-updates-ionic1.vercel.app/docs/react/add-to-existing)

**Angular**
- No Before | [After](https://ionic-docs-git-docs-cdn-updates-ionic1.vercel.app/docs/angular/add-to-existing)

**Vue**
- No Before | [After](https://ionic-docs-git-docs-cdn-updates-ionic1.vercel.app/docs/vue/add-to-existing)

> [!IMPORTANT]
[...]

PR #4329, opened 2025-11-18
chore(react): remove old files that are no longer used

brandyscarney:
These React guides were renamed in https://github.com/ionic-team/ionic-docs/pull/4286 but not deleted because we need to update the JP docs before we can delete them. These files should not be reachable due to redirects so they are safe to delete.

⚠️ This PR should not be merged until https://github.com/ionic-team/ionic-docs/pull/4334 is merged. ⚠️


thetaPC (approved):
LGTM

PR #4482, opened 2026-04-30
fix(api): do not render a code block for unnamed slots

brandyscarney:
When the slots table renders a slot with no name, it shows up as backticks only. See [item](https://ionicframework.com/docs/api/item#slots):

before

This PR updates it to [not render anything](https://ionic-docs-git-fix-slots-render-ionic1.vercel.app/docs/api/item#slots):

after


thetaPC (approved):
LGTM

ShaneK (approved):

PR #4490, opened 2026-05-07
docs(select): document ionChange behavior change in v9 upgrade guide

ShaneK:
Adding the breaking changes for the select component from PR [#31125](https://github.com/ionic-team/ionic-framework/pull/31125)


thetaPC (approved):
LGTM

PR #4505, opened 2026-05-15
docs(vue): add troubleshooting entry for blank screen in Capacitor builds

### LINKED ISSUE #2106: Vue PublicPath Docs/Troubleshooting
**URL**
https://ionicframework.com/docs/vue/troubleshooting

https://cli.vuejs.org/config/#publicpath

Vue offers the ability to overrride the base URL for the application by specifying a "publicPath" in the vue.config.js file. This config is required if you deploy the application to GItHub pages, since the server behaves similar to this Nginx Config https://github.com/brentspector/reborn-pokepedia/blob/master/nginx/nginx.conf

The issue is that when converting the app to an Android app, the build uses this override. Unfortunately, android apps do not need this and it causes the app to deploy a blank white screen.

A couple of things to add

1. Troubleshooting if "publicPath" is correctly set.
2. Using "logcat" in Android Studio to gain insight into why an app is showing a blank white screen.

ShaneK:
resolves #2106

Adding documentation to help troubleshoot white screens with Vue builds in Capacitor

### DIFF
--- a/docs/vue/troubleshooting.md
+++ b/docs/vue/troubleshooting.md
@@ -132,7 +132,16 @@ export default defineConfig({
 
 In a Capacitor build the bundled assets are served from a local origin (`capacitor://localhost` on iOS and `https://localhost` on Android by default), so the prefixed paths never resolve and the app fails to bootstrap.
 
-To fix it, reset `base` to `/` (or remove the option) before running `npx cap copy`. If you need both targets, keep a separate config file for each and select it at build time with `vite build --config`.
+To fix it, reset `base` to `/` (or remove the option) before running `npx cap copy`.
+
+```js
+// vite.config.js
+export default defineConfig({
+  base: '/',
+});
+```
+
+If you need both targets, keep a separate config file for each and select it at build time with `vite build --config`.
 
 To confirm this is the cause, inspect the device log for 404s on the prefixed asset paths:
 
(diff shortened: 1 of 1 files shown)


docs/vue/troubleshooting.md
```diff
@@ -116,3 +116,25 @@ await modal.present();
...
+```js
+// vite.config.js
+export default defineConfig({
+  base: '/my-repo/',
+});
+```
+
+In a Capacitor build the bundled assets are served from a local origin (`capacitor://localhost` on iOS and `https://localhost` on Android by default), so the prefixed paths never resolve and the app fails to bootstrap.
+
+To fix it, reset `base` to `/` (or remove the option) before running `npx cap copy`. If you need both targets, keep a separate config file for each and select it at build time with `vite build --config`.
```
thetaPC: Let's also add the code block for this like you did above. It makes it easy for the dev to quickly copy it and paste it.


````suggestion
To fix it, reset `base` to `/` (or remove the option) before running `npx cap copy`.

```js
// vite.config.js
export default defineConfig({
  base: '/',
});
```

If you need both targets, keep a separate config file for each and select it at build time with `vite build --config`.
````

thetaPC (approved):
LGTM, with minor suggestion

PR #4512, opened 2026-05-20
docs(select): add new section on rich content options

brandyscarney:
Adds the following:

1. Select documentation for the new feature allowing [HTML to be passed to select options](https://github.com/ionic-team/ionic-framework/pull/31072).
2. Updates the StackBlitz utilities to combine any config object being passed with the mode of the Playground. 

[Preview](https://ionic-docs-git-fw-7289-ionic1.vercel.app/docs/api/select#rich-content-options)


src/components/global/Playground/index.tsx
```diff
@@ -150,6 +157,12 @@ export default function Playground({
   devicePreview?: boolean;
   showConsole?: boolean;
   includeIonContent: boolean;
+  /**
+   * Ionic config values to inject into generated StackBlitz examples. For
+   * example: `{ innerHTMLTemplatesEnabled: true }`. Merges with the active
+   * preview `mode` when opening the editor.
+   */
+  ionicConfig?: IonicConfig;
```
brandyscarney: I decided to add this as a config object instead of adding `innerHTMLTemplatesEnabled` as a `boolean` because this will allow us to enhance the [Config](https://ionicframework.com/docs/developing/config) documentation if desired by adding playgrounds passing different config options.

src/components/global/Playground/playground.types.ts
```diff
@@ -14,3 +14,9 @@ export interface ConsoleItem {
   type: 'log' | 'warning' | 'error';
   message: string;
 }
+
+/**
+ * Ionic app configuration. See [IonicConfig](/docs/developing/config#ionicconfig).
+ * Playground only injects serializable values (string, number, boolean) into StackBlitz.
+ */
+export type IonicConfig = Record<string, unknown>;
```
brandyscarney: I was going to import the type from Ionic but then we would have a dependency on `@ionic/core` and I don't know if we want that.

src/components/global/Playground/README.md
```diff
@@ -71,3 +71,11 @@ To opt-out of this behavior, set `includeIonContent={false}` to disable this wra
 ```tsx
 <Playground includeIonContent={false} />
 ```
+
+## Ionic Config
+
+Pass an [IonicConfig](/docs/developing/config#ionicconfig) object so generated StackBlitz projects bootstrap with the same settings. This is merged with the active preview `mode` when opening the editor. Only string, number, and boolean values are injected into generated code:
```
brandyscarney: I chose this rather than adding an entire file for each framework to pass the config like [Vue](https://github.com/ionic-team/ionic-docs/blob/main/static/usage/v8/tabs/router/vue/main_ts.md) but if we prefer that approach I can revert this.

static/usage/v9/select/rich-content-options/javascript/index_html.md
```diff
@@ -0,0 +1,112 @@
...
+  /* Helper function to format the interface label */
+  const formatInterfaceLabel = (selectInterface) => {
+    return selectInterface
+      .split('-')
+      .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
+      .join(' ');
+  };
+
+  /* Dynamically create the different Select components */
+  const selectInterfaces = ['alert', 'action-sheet', 'modal', 'popover'];
```
brandyscarney, line 78: Originally I defined the interface array like this:

```tsx
const selectInterfaces = [{
  label: 'Alert',
  value: 'alert',
}, {
  label: 'Action Sheet',
  value: 'action-sheet',
}, {
  label: 'Modal',
  value: 'modal',
}, {
  label: 'Popover',
  value: 'popover',
}];
```

but I removed it in favor of using the single string and converting it to the label. If we think this is too much for the demo, I can go back to the object array.

static/usage/v9/select/rich-content-options/javascript/index_html.md
```diff
@@ -0,0 +1,112 @@
...
+        <ion-badge slot="end" color="tertiary">$64</ion-badge>
+      </ion-select-option>
+    `;
+    select.innerHTML = optionsHTML;
+
+    /* The interfaceOptions property is for demonstration purposes only. */
+    /* It is not required for the rich options to work. */
+    select.interfaceOptions = {
+      header: 'Travel mode',
+    };
```
brandyscarney, line 120: This is just to make the demo look nicer but can be removed if desired.

docs/api/select.md
```diff
@@ -201,6 +201,20 @@ import StartEndSlots from '@site/static/usage/v9/select/start-end-slots/index.md
 
 <StartEndSlots />
 
+## Rich Content Options
```
thetaPC, line 204: It should also mention that a selected text will strip out the HTML to only provide plain text. Also start, end slots and descriptions are not included.

docs/api/select.md
```diff
@@ -201,6 +201,20 @@ import StartEndSlots from '@site/static/usage/v9/select/start-end-slots/index.md
...
+
+:::important
```
thetaPC, line 206: Not part of this PR but personal question: Do you think purple is appropriate for `important`?

docs/api/select.md
```diff
@@ -201,6 +201,20 @@ import StartEndSlots from '@site/static/usage/v9/select/start-end-slots/index.md
...
+Rich content in select options is disabled by default. Set [`innerHTMLTemplatesEnabled`](/docs/developing/config#ionicconfig) to `true` in your [global Ionic config](/docs/developing/config#global-config). Markup inside options is treated as plain text when it is disabled. Refer to [Security](/docs/techniques/security.md) for sanitization guidance when enabling custom HTML.
```
thetaPC: By adding `.md`, it will be easier to migrate the files later from `.md` to `.mdx`.


```suggestion
Rich content in select options is disabled by default. Set [`innerHTMLTemplatesEnabled`](/docs/developing/config.md#ionicconfig) to `true` in your [global Ionic config](/docs/developing/config.md#global-config). Markup inside options is treated as plain text when it is disabled. Refer to [Security](/docs/techniques/security.md) for sanitization guidance when enabling custom HTML.
```

docs/techniques/security.md
```diff
@@ -61,7 +61,7 @@ To learn more about the security recommendations for binding to directives such
 
 ## Enabling Custom HTML Parsing via `innerHTML`
 
-`ion-alert`, `ion-infinite-scroll-content`, `ion-loading`, `ion-refresher-content`, and `ion-toast` can accept custom HTML as strings for certain properties. These strings are added to the DOM using `innerHTML` and must be properly sanitized by the developer. This behavior is disabled by default which means values passed to the affected components will always be interpreted as plaintext. Developers can enable this custom HTML behavior by setting `innerHTMLTemplatesEnabled: true` in the [IonicConfig](../developing/config#ionicconfig).
+`ion-alert`, `ion-infinite-scroll-content`, `ion-loading`, `ion-refresher-content`, `ion-select-option`, and `ion-toast` can accept custom HTML as strings for certain properties. These strings are added to the DOM using `innerHTML` and must be properly sanitized by the developer. This behavior is disabled by default which means values passed to the affected components will always be interpreted as plaintext. Developers can enable this custom HTML behavior by setting `innerHTMLTemplatesEnabled: true` in the [IonicConfig](../developing/config#ionicconfig).
```
thetaPC: By adding `.md`, it will be easier to migrate the files later from `.md` to `.mdx`. Might as well add it since we're here.



```suggestion
`ion-alert`, `ion-infinite-scroll-content`, `ion-loading`, `ion-refresher-content`, `ion-select-option`, and `ion-toast` can accept custom HTML as strings for certain properties. These strings are added to the DOM using `innerHTML` and must be properly sanitized by the developer. This behavior is disabled by default which means values passed to the affected components will always be interpreted as plaintext. Developers can enable this custom HTML behavior by setting `innerHTMLTemplatesEnabled: true` in the [IonicConfig](../developing/config.md#ionicconfig).
```

src/components/global/Playground/index.tsx
thetaPC: I don't mind this, but my concern is that it might be confusing for devs who are reviewing the demos. Some might not realize the important admonition and only copy/paste code. They'll probably think that the select option has a bug. If we included `innerHTMLTemplatesEnabled` within the code then we can avoid this.

I don't feel strongly about this since:
1. you do have that admonition
2. the demos only provide the bare minimal (meaning we don't show files like `App.tsx` to showcase config changes)

src/components/global/Playground/README.md
thetaPC: By adding `.md`, it will be easier to migrate the files later from `.md` to `.mdx`.


```suggestion
Pass an [IonicConfig](/docs/developing/config.md#ionicconfig) object so generated StackBlitz projects bootstrap with the same settings. This is merged with the active preview `mode` when opening the editor. Only string, number, and boolean values are injected into generated code:
```

thetaPC (approved):
LGTM, minor suggestions!

Thank you for updating the Playground docs!

src/components/global/Playground/README.md
brandyscarney: This section has been removed.

docs/techniques/security.md
brandyscarney: https://github.com/ionic-team/ionic-docs/pull/4512/commits/14ba310fa777289943a9b6a880fc6874f90f26ef

src/components/global/Playground/README.md
brandyscarney: This has been reverted in favor of passing individual files: 

https://github.com/ionic-team/ionic-docs/pull/4512/changes/948fe5f6a631449e59ac208b49fcc089902a1596
https://github.com/ionic-team/ionic-docs/pull/4512/changes/eedbd2a26fd3b93641934ab74d620664468dff69
https://github.com/ionic-team/ionic-docs/pull/4512/changes/0b5529f769793c7913b90a61f7222ccf3eebd941

src/components/global/Playground/playground.types.ts
brandyscarney: This has been reverted in favor of passing individual files: 

https://github.com/ionic-team/ionic-docs/pull/4512/changes/948fe5f6a631449e59ac208b49fcc089902a1596
https://github.com/ionic-team/ionic-docs/pull/4512/changes/eedbd2a26fd3b93641934ab74d620664468dff69
https://github.com/ionic-team/ionic-docs/pull/4512/changes/0b5529f769793c7913b90a61f7222ccf3eebd941

src/components/global/Playground/index.tsx
brandyscarney: At first, I was going to argue that this would be easier for us to maintain, but I ultimately agree that the requirement to update a config value is somewhat hidden and could be confusing.

I reverted adding a new `ionicConfig` object: https://github.com/ionic-team/ionic-docs/pull/4512/changes/948fe5f6a631449e59ac208b49fcc089902a1596

I then updated the playgrounds to pass the config in their respective files: https://github.com/ionic-team/ionic-docs/pull/4512/changes/eedbd2a26fd3b93641934ab74d620664468dff69

I updated StackBlitz utils to merge the config object with the mode object that is taken from the docs website based on which mode is currently selected:
https://github.com/ionic-team/ionic-docs/pull/4512/changes/0b5529f769793c7913b90a61f7222ccf3eebd941

And then I added some spec tests to make sure this merging continues to work: https://github.com/ionic-team/ionic-docs/pull/4512/changes/8bab6aa05c8b67039a216fea332c6075da6baea8

Please review all of the new changes. Thank you for pointing this out!

docs/api/select.md
brandyscarney: I think we just copied from GitHub didn't we? https://github.com/orgs/community/discussions/16925

docs/api/select.md
brandyscarney: https://github.com/ionic-team/ionic-docs/pull/4512/commits/506b6ff18a6dd90d501c13cf27c4bbecfcaca6b8

docs/api/select.md
brandyscarney: https://github.com/ionic-team/ionic-docs/pull/4512/commits/506b6ff18a6dd90d501c13cf27c4bbecfcaca6b8

docs/api/select.md
thetaPC: I don't recall touching admonitions during the great upgrade so it seems that it's old styles.

thetaPC (approved):
LGTM

docs/api/select.md
brandyscarney: [I added it](https://github.com/ionic-team/ionic-docs/pull/4321)

thetaPC (approved):
LGTM

PR #4568, opened 2026-07-01
docs(support): update support tables for v9 and enterprise

brandyscarney:
Updates the support guide to add v9 and mark v8 as maintenance, plus removes the old enterprise links.


thetaPC (approved):
LGTM

PR #4569, opened 2026-07-01
docs(angular): migrate your-first-app tutorial to standalone components and signals

ShaneK:
## Description 

Rewrites the Angular "Build Your First App" tutorial to use standalone components and signals instead of NgModules, matching the v9 default. The `ionic start` step now tells readers to choose `Standalone`, `main.ts` uses `bootstrapApplication` with `provideIonicAngular`/`provideRouter`, and `PhotoService.photos` is now a signal.

## Other

Signals are used deliberately: on Angular 21+ this app runs zoneless, so mutating a plain array wouldn't trigger a re-render. The tutorial calls this out where the `photos` signal is introduced and links to the zoneless guide.

This guide assume you're using the Angular tutorial with this PR applied: https://github.com/ionic-team/tutorial-photo-gallery-angular/pull/81

### DIFF
--- a/docs/angular/your-first-app.md
+++ b/docs/angular/your-first-app.md
@@ -78,7 +78,7 @@ ionic start photo-gallery tabs --type=angular
 
 :::note
 
-When prompted to choose between `NgModules` and `Standalone`, opt for `NgModules` as this tutorial follows the `NgModules` approach.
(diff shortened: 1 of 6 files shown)


brandyscarney (approved):
It's hard to test this out since it seems to depend on a few in progress branches (mainly the starters) but visually it looks good!

PR #4585, opened 2026-07-14
docs(router): remove ion-nav from coordinated navigation outlets

ShaneK:
Currently, the `ion-router` API docs list `ion-nav` as one of the navigation outlets that `ion-router` coordinates. That's inaccurate. `ion-nav` manages its own imperative navigation stack and isn't URL-coordinated, which its own docs already state ("pushing others won't affect overall routers").

With this change, the outlet list on the `ion-router` page names only `ion-tabs` and `ion-router-outlet`. Applied to the `<head>` meta description and both overview sentences.

The same wording exists in the v8 and v7 snapshots since the inaccuracy predates all three, so I corrected `versioned_docs/version-v8/api/router.md` and `versioned_docs/version-v7/api/router.md` to match.

Also documented migration path for those using `ion-nav` as a router outlet.


thetaPC (approved):
LGTM

brandyscarney (approved):

docs/api/router.md
```diff
@@ -40,6 +40,16 @@ import BasicExample from '@site/static/usage/v9/router/basic/index.md';
 
 <BasicExample />
 
+## Using ion-nav within a Routed Page
+
+`ion-router` and [`ion-nav`](./nav.md) are separate systems. `ion-router` coordinates URL-based navigation through `ion-router-outlet`, while `ion-nav` manages a local stack that is independent of the URL. As of Ionic 9, `ion-nav` no longer integrates with `ion-router`: placing an `ion-nav` inside an `ion-router` does not turn it into a routed outlet, and pushing or popping views on an `ion-nav` never changes the URL. Refer to the [migration guide](../updating/9-0.md#nav) for details.
```
brandyscarney:
```suggestion
`ion-router` and [`ion-nav`](./nav.md) are separate systems. `ion-router` coordinates URL-based navigation through `ion-router-outlet`, while `ion-nav` manages a local stack that is independent of the URL. `ion-nav` does not integrate with `ion-router`: placing an `ion-nav` inside an `ion-router` does not turn it into a routed outlet, and pushing or popping views on an `ion-nav` never changes the URL. 
```

Since this is a new section I don't think we need to specify "as of Ionic 9" or link to the migration guide. If we want people to see it due to the change, we can link this from the migration guide.

docs/api/router.md
```diff
@@ -40,6 +40,16 @@ import BasicExample from '@site/static/usage/v9/router/basic/index.md';
...
+
+The two can still be composed. A routed page rendered by `ion-router-outlet` can host its own `ion-nav` for local, URL-less navigation within that page. In the example below, navigating to Details updates the URL, but stepping through the `ion-nav` inside that page does not.
```
brandyscarney:
```suggestion
The two can still be composed. A routed page rendered by `ion-router-outlet` can host its own `ion-nav` for local, URL-less navigation within that page. In the example below, navigating to `/details` updates the URL, but stepping through the `ion-nav` inside that page does not.
```

Can we make this more clear that we're talking about a page? Either this or `DetailsStepOne`?

static/usage/v9/router/nav-within-page/javascript.md
brandyscarney, line 1: This example does not work for me in StackBlitz:

https://github.com/user-attachments/assets/cead43f3-fd0b-41e1-89fc-384ef914a887

brandyscarney (commented):
Left a couple minor requests and StackBlitz wasn't working for me. Not sure if the versions are supposed to work together. 🤔

docs/api/router.md
ShaneK: Yeah, I guess. I mostly assumed people would be coming here from the migration, but I suppose it doesn't make sense to include this in the docs here

static/usage/v9/router/nav-within-page/javascript.md
ShaneK: I forget that the stackblitz and inline playground work differently sometimes 🤦‍♂️ 

Fixed: https://ionic-docs-git-fw-6976-ionic1.vercel.app/docs/api/router#using-ion-nav-within-a-routed-page

Commit: [fb6326b](https://github.com/ionic-team/ionic-docs/pull/4585/commits/fb6326b351e75e59e38eabdfd72535fd0daf900f)

This commit upgrades *all* vanilla JS playgrounds to v9, which should be fine - not much breaking in vanilla JS for v9, but did require a removal of dead code that would have broken the JS playgrounds across the board.

brandyscarney (approved):
Looks good now!

PR #4598, opened 2026-07-20
fix(release-notes): require token on CI, warn for local development

brandyscarney:
This PR fixes the release notes generation failing on Vercel:

- Updates the repository path to ionic-team/ionic-framework
- Improved build output with cleaner logging
- Adds contributing section for keeping PAT updated

Production Release Notes: https://ionicframework.com/docs/reference/release-notes
Branch Release Notes: https://ionic-docs-git-fw-7622-ionic1.vercel.app/docs/reference/release-notes

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "node": ">=20.0.0"
   },
   "scripts": {
-    "prebuild": "npm run generate-markdown; scripts/i18n.sh; if [ -n \"$CROWDIN_PERSONAL_TOKEN\" ]; then npm run crowdin:sync; fi",
+    "prebuild": "npm run generate-markdown && scripts/i18n.sh && if [ -n \"$CROWDIN_PERSONAL_TOKEN\" ]; then npm run crowdin:sync; fi",
     "build": "npm run build:${VERCEL_ENV:-preview}",
     "build:preview": "docusaurus build --locale en",
     "build:production": "docusaurus build",
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "crowdin:sync": "docusaurus write-translations && crowdin upload && crowdin download",
     "deploy": "docusaurus deploy",
     "docusaurus": "docusaurus",
-    "generate-markdown": "node scripts/native.mjs && concurrently \"node scripts/cli.mjs\" \"node scripts/release-notes.mjs\"",
+    "generate-markdown": "node scripts/native.mjs && node scripts/cli.mjs && node scripts/release-notes.mjs",
     "lint": "npm run prettier -- --write",
     "serve": "docusaurus serve",
     "playground:new": "hygen playground new",
--- a/package.json
+++ b/package.json
@@ -48,7 +48,6 @@
     "@stackblitz/sdk": "^1.6.0",
     "@tippyjs/react": "^4.2.6",
     "clsx": "^1.1.1",
-    "concurrently": "^6.2.0",
     "crowdin": "^3.5.0",
     "docusaurus-plugin-copy-page-button": "^0.8.2",
     "docusaurus-plugin-module-alias": "^0.0.2",
--- a/scripts/release-notes.mjs
+++ b/scripts/release-notes.mjs
@@ -15,22 +15,15 @@ const OUTPUT_PATH = resolve(__dirname, '../src/components/page/reference/Release
 //   task: async () => outputJson(OUTPUT_PATH, await getReleases(), { spaces: 2 })
 // };
 
-// Get the GitHub Releases from Ionic
+// Get the GitHub Releases from Ionic Framework
 // -------------------------------------------------------------------------------
-// This requires an environment GITHUB_TOKEN otherwise it may fail
-//
-// To add a GITHUB_TOKEN, follow the steps to create a personal access token:
-// https://docs.github.com/en/enterprise-cloud@latest/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token
-// and then authorize it to work with SSO:
-// https://docs.github.com/en/enterprise-cloud@latest/authentication/authenticating-with-saml-single-sign-on/authorizing-a-personal-access-token-for-use-with-saml-single-sign-on
+// Fetches from the public GitHub API. Unauthenticated requests have a 60 req/hour limit.
+// To increase the rate limit, set a GITHUB_TOKEN environment variable.
 const getReleases = async () => {
   try {
-    const request = await fetch(new URL('repos/ionic-team/ionic/releases', 'https://api.github.com'), {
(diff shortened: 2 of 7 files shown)


scripts/release-notes.mjs
```diff
@@ -98,7 +93,14 @@ function getVersionType(version) {
...
 async function run() {
   const { outputJson } = pkg;
-  outputJson(OUTPUT_PATH, await getReleases(), { spaces: 2 });
+  try {
+    const releases = await getReleases();
+    outputJson(OUTPUT_PATH, releases, { spaces: 2 });
+    console.log(`🚀 Release Notes Generated`);
+  } catch (error) {
+    console.error(`\n❌ Release Notes Failed\n  ⇢ ${error.message}\n`);
+    process.exit(1);
```
ShaneK: I don't love that `npm start` now hard-requires a `GITHUB_TOKEN`. `prestart` runs `generate-markdown`, so anyone editing an unrelated doc has to set up a PAT just to boot the dev server, where before it degraded gracefully. Could we keep the hard fail for CI/Vercel but let local dev fall back to empty release notes with a warning? Gating the `process.exit(1)` on `process.env.CI`/`VERCEL` would do it.

Separately: `prebuild` chains with `;` instead of `&&` (`npm run generate-markdown; scripts/i18n.sh; ...`), so this `process.exit(1)` gets swallowed. I verified a failing first command still leaves `prebuild` at exit 0. The Vercel build only actually fails because `release-notes.json` is gitignored and the missing import kills docusaurus later, so the loud exit is effectively redundant on the build path. If that JSON ever gets committed or cached, a failed regen would silently ship stale notes. Is the `;` intended?

scripts/release-notes.mjs
```diff
@@ -10,64 +10,59 @@ const __filename = fileURLToPath(import.meta.url);
...
+  // Check that the response is an array in case it was
+  // successful but returned an object
+  if (Array.isArray(releases)) {
+    return releases
+      .filter((release) => {
+        const releasePattern = /^v(\d+)\.(\d+)\.(\d+)$/;
+        return releasePattern.test(release.tag_name);
+      })
+      .map((release) => {
+        const body = renderMarkdown(release.body.replace(/^#.*/, '')).value;
```
ShaneK:
```suggestion
        const body = renderMarkdown((release.body ?? '').replace(/^#.*/, '')).value;
```

Nit: `release.body` can be null for a release with no description, and this isn't wrapped in the old catch-returns-`[]` anymore, so one such release would abort the whole build. I checked the API and 0 of the last 92 stable releases have an empty body, so it's basically theoretical for us. Up to you!

ShaneK (changes requested):
I like the idea behind these changes but I believe the current implementation of them limits public contribution, which isn't ideal

scripts/release-notes.mjs
ShaneK: I don't think this is super realistic, but it's really easy to add this fallback just in case

brandyscarney: https://github.com/ionic-team/ionic-docs/pull/4598/commits/e5ec9b8e2ee96efb4a9d431c674a6a05b4d00dae

scripts/release-notes.mjs
brandyscarney: Good catch! I wasn't even thinking about how annoying this would be for contributors. I have updated the code to check for CI or the VERCEL environment (I used only `production` and `preview` because the token is not set on Vercel for the `development` environment).

I don't know why it was like that in `prebuild` but I updated it and verified it fails properly now on `build` as well. Let me know if there are any other changes you think should be made: https://github.com/ionic-team/ionic-docs/pull/4598/commits/e5ec9b8e2ee96efb4a9d431c674a6a05b4d00dae

package.json
```diff
@@ -6,15 +6,15 @@
...
+    "prebuild": "npm run generate-markdown && scripts/i18n.sh && if [ -n \"$CROWDIN_PERSONAL_TOKEN\" ]; then npm run crowdin:sync; fi",
     "build": "npm run build:${VERCEL_ENV:-preview}",
     "build:preview": "docusaurus build --locale en",
     "build:production": "docusaurus build",
     "clear": "docusaurus clear",
     "crowdin:sync": "docusaurus write-translations && crowdin upload && crowdin download",
     "deploy": "docusaurus deploy",
     "docusaurus": "docusaurus",
-    "generate-markdown": "node scripts/native.mjs && concurrently \"node scripts/cli.mjs\" \"node scripts/release-notes.mjs\"",
+    "generate-markdown": "node scripts/native.mjs && node scripts/cli.mjs && node scripts/release-notes.mjs",
```
ShaneK, line 17: Nit: now that `generate-markdown` runs sequentially with `&&`, nothing uses `concurrently` anymore. Could drop it from `devDependencies` (line 52). Up to you!

src/components/page/reference/ReleaseNotes/index.tsx
```diff
@@ -26,7 +26,7 @@ https://docs.github.com/en/enterprise-cloud@latest/authentication/authenticating
     return [
       <p>
         Unable to load Releases. Please see all releases{' '}
-        <a href="https://github.com/ionic-team/ionic/releases" target="_blank">
+        <a href="https://github.com/ionic-team/ionic-framework/releases" target="_blank">
```
ShaneK, line 25: Nit, just above this at the `console.warn`: it still tells devs to create a classic PAT and authorize it for SSO, but this PR standardized on a fine-grained PAT in `CONTRIBUTING.md` and `release-notes.mjs` now points there. Might be worth updating the warning to match so the guidance is consistent. No worries if you'd rather leave it.

package.json
ShaneK: Side-note: I hate that GitHub doesn't let you suggest/comment on lines outside of the diff of the PR. Super annoying

ShaneK (approved):
Awesome work!

package.json
brandyscarney: https://github.com/ionic-team/ionic-docs/pull/4598/changes/1db377ff765520f8bd586965b2e7b2b699e5b98b

src/components/page/reference/ReleaseNotes/index.tsx
brandyscarney: I actually didn't even realize the warn was there lol. I updated it to point to the guide: https://github.com/ionic-team/ionic-docs/pull/4598/changes/d6644cb5a4bb61231e71ca27ab644a51c0323486

Also fixed some React errors while I was in there: https://github.com/ionic-team/ionic-docs/pull/4598/commits/95c67b8da51378d396f5bb676e5050024ddc0a03

PR #4606, opened 2026-07-28
fix(stackblitz): pin vite to v7 to avoid rolldown wasm failure in web container

ShaneK:
Renovate bumped the StackBlitz playground scaffolds to `vite@8`, which uses Rolldown as its bundler. Rolldown's native binding fails to load in StackBlitz's WebContainer (`Cannot find native binding`), so every non-Angular playground broke on start. Tracked upstream in [stackblitz/webcontainer-core#2141](https://github.com/stackblitz/webcontainer-core/issues/2141)

This pins `vite` back to `^7` (Rollup-based, no Rolldown) in the react, vue, and html scaffolds for v7 and v8, and regenerates their lockfiles to drop the `rolldown` and `@rolldown/binding-*` entries. Angular already pins `vite@7.3.5` via the CLI, so it was unaffected and is left alone. Also caps `vite` at `^7` in the three vite Renovate rules so it won't re-bump to 8 and reintroduce the failure until we can be sure it's fixed.

Easy preview link: https://ionic-docs-git-fix-vite-ionic1.vercel.app/docs/api/tab-bar

Also, I created FW-7632 to undo this later.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -25,7 +25,6 @@
     {
       "matchPackageNames": ["vite", "vite-plugin-static-copy"],
       "groupName": "vite-html",
-      "allowedVersions": "^7.0.0",
       "matchFileNames": [
         "static/code/stackblitz/**/html/package.json"
       ]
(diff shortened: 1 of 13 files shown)


thetaPC (approved):
LGTM, thank you!

PR #4609, opened 2026-07-29
docs(angular): remove standalone true from v8 and v9 playgrounds

brandyscarney:
I noticed some playgrounds still include `standalone: true` which is no longer needed as it is the default.


thetaPC (approved):
LGTM

PR #4611, opened 2026-07-31
docs(testing): add componentOnReady guidance and a Vue testing page

### LINKED ISSUE #3808: content: Ionic Angular Testing (Standalone etc)
### URL

https://ionicframework.com/docs/angular/testing

### Issue Description

We are in dire need of some documentation on how to test ionic in angular.

The current documentation explains the theory behind testing but does not go into ionic specifics. I'd argue that the theory behind what is a unit test has been covered elsewhere in more detail and so it's not important for the ionic docs to go over that.

What we could use is knowing how to correctly bootstrap ionic so that components are loaded correctly and can be tested. I know theres some nuances due to ionics web component architecture. Some of it is explained on stencil docs and elsewhere.

While I've been able to figure out some things through looking the ionic source code and also a bit of blind luck there are some things I haven't been able to determine yet. For example when testing a standalone project it's not clear how to bootstrap ionic in testbed or whether its required at all? Using `provideIonicAngular()` doesn't seem to render any components but importing IonicModule does (but only if calling forRoot).

```
TestBed.configureTestingModule({
      providers: [
        importProvidersFrom(IonicModule.forRoot()), <<< - works but is this right?
        provideIonicAngular({}), // does not render but is more alike to real world
      ],
    }).compileComponents()
```

But there are issues with using `IonicModule.forRoot` as described here https://github.com/ionic-team/ionic-framework/issues/23173#issuecomment-1854778055 hence why standalone is a good idea but these components don't seem to render?

Any advice is helpful on this to point us in the right direction 🙏

Zac-Smucker-Bryan:
Resolves #3808

## Description

- Adds guidance about import `componentOnReady` for standalone projects from `@ionic/core` as a helper to improve testing when working with Ionic standalone, for Angular, React, and Vue
- Adds a testing page for Vue (mostly a starter template with the standalone guidance above)
- Adds specific testing example in the case of Angular w/the import implemented and then the helper called in the test.

## Change Type
- [ ] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [x] Documentation
- [ ] Other (CI, chores, etc.)

## Rationale / Problems Fixed

Standalone users can't always consistently test what they need without better guidance.

## Notes / Comments

There is a companion issue (https://github.com/ionic-team/ionic-framework/issues/31312) and PR (https://github.com/ionic-team/ionic-framework/pull/31313) in ionic-framework that might need to be coordinated for language tweaks.

### DIFF
--- a/docs/angular/testing.md
+++ b/docs/angular/testing.md
@@ -109,6 +109,34 @@ describe('TabsPage', () => {
 
 When doing component class testing, the component object is accessed using the component object defined via `component = fixture.componentInstance;`. This is an instance of the component class. When doing DOM testing, the `fixture.nativeElement` property is used. This is the actual `HTMLElement` for the component, which allows the test to use standard HTML API methods such as `HTMLElement.querySelector` in order to examine the DOM.
 
+### Waiting for Components
+
+When testing Ionic components, use the `componentOnReady` helper exported from `@ionic/core` rather than calling `el.componentOnReady()` directly. The `el.componentOnReady()` method only exists on lazy-loaded elements and calling it directly throws an error on custom-element builds, which is what standalone projects use. The helper handles both. It awaits the element's own `componentOnReady()` promise when that exists. Otherwise it waits one animation frame, giving the component's inner contents a chance to render. Wait for the callback before asserting against the rendered DOM or running accessibility tests.
+
+```tsx
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { componentOnReady } from '@ionic/core';
+import { HomePage } from './home.page';
+
+describe('HomePage', () => {
+  let fixture: ComponentFixture<HomePage>;
+
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      imports: [HomePage],
+    }).compileComponents();
+    fixture = TestBed.createComponent(HomePage);
+    fixture.detectChanges();
+  });
+
+  it('renders the submit button', async () => {
+    const button = fixture.nativeElement.querySelector('ion-button');
+    await new Promise<void>((resolve) => componentOnReady(button, () => resolve()));
+    expect(button.textContent).toContain('Submit');
+  });
+});
+```
+
 ## Services
 
 Services often fall into one of two broad categories: utility services that perform calculations and other operations, and data services that perform primarily HTTP operations and data manipulation.
@@ -177,11 +205,6 @@ describe('PayrolService', () => {
   });
 });
 ```
-
-### Waiting for Components
-
-When testing Ionic components, use the `componentOnReady` helper exported from `@ionic/core` rather than calling `el.componentOnReady()` directly. That method only exists on lazy-loaded elements. Calling it directly throws on custom-element builds, which is what standalone projects use. The helper handles both. It awaits the element's own `componentOnReady()` promise when that exists. Otherwise it waits one animation frame, giving the component's inner contents a chance to render. Wait for the callback before asserting against the rendered DOM or running accessibility tests.
-
 
 #### Testing HTTP Data Services
 
--- a/docs/angular/testing.md
+++ b/docs/angular/testing.md
@@ -178,6 +178,47 @@ describe('PayrolService', () => {
 });
 ```
 
+### Import `componentOnReady` for standalone projects
+
+When testing Ionic components, use the exported `componentOnReady` helper from `@ionic/core` instead of calling `el.componentOnReady()` directly. The helper works with both lazy-loaded and custom-element builds, making it more likely the component has finished rendering before making assertions against its rendered DOM or running accessibility tests. See an example here:
+
+```tsx
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { axe, toHaveNoViolations } from 'jasmine-axe';
+import { componentOnReady } from '@ionic/core';
+
+import { Component } from './component';
+
+describe('Component', () => {
+  let component: Component;
+  let fixture: ComponentFixture<Component>;
+
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      imports: [Component],
+    }).compileComponents();
+
+    fixture = TestBed.createComponent(Component);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should pass accessibility test', async () => {
+    const el = fixture.nativeElement.querySelector('ion-button');
+
+    await new Promise<void>((resolve) => {
+      componentOnReady(el, () => resolve());
+    });
+
+    jasmine.addMatchers(toHaveNoViolations);
+
+    const a11y = await axe(fixture.nativeElement);
+
+    expect(a11y).toHaveNoViolations();
+  });
+});
+```
+
 #### Testing HTTP Data Services
 
 Most services that perform HTTP operations will use Angular's HttpClient service in order to perform those operations. For such tests, it is suggested to use Angular's `HttpClientTestingModule`. For detailed documentation of this module, please see Angular's <a href="https://angular.io/guide/http#testing-http-requests" target="_blank">Angular's Testing HTTP requests</a> guide.
--- a/docs/angular/testing.md
+++ b/docs/angular/testing.md
@@ -178,46 +178,10 @@ describe('PayrolService', () => {
 });
 ```
 
-### Import `componentOnReady` for standalone projects
+### Waiting for Components
 
-When testing Ionic components, use the exported `componentOnReady` helper from `@ionic/core` instead of calling `el.componentOnReady()` directly. The helper works with both lazy-loaded and custom-element builds, making it more likely the component has finished rendering before making assertions against its rendered DOM or running accessibility tests. See an example here:
+When testing Ionic components, use the `componentOnReady` helper exported from `@ionic/core` rather than calling `el.componentOnReady()` directly. That method only exists on lazy-loaded elements. Calling it directly throws on custom-element builds, which is what standalone projects use. The helper handles both. It awaits the element's own `componentOnReady()` promise when that exists. Otherwise it waits one animation frame, giving the component's inner contents a chance to render. Wait for the callback before asserting against the rendered DOM or running accessibility tests.
 
-```tsx
-import { ComponentFixture, TestBed } from '@angular/core/testing';
-import { axe, toHaveNoViolations } from 'jasmine-axe';
-import { componentOnReady } from '@ionic/core';
-
-import { Component } from './component';
-
-describe('Component', () => {
-  let component: Component;
-  let fixture: ComponentFixture<Component>;
-
-  beforeEach(async () => {
-    await TestBed.configureTestingModule({
-      imports: [Component],
-    }).compileComponents();
-
-    fixture = TestBed.createComponent(Component);
-    component = fixture.componentInstance;
-    fixture.detectChanges();
-  });
-
-  it('should pass accessibility test', async () => {
-    const el = fixture.nativeElement.querySelector('ion-button');
-
-    await new Promise<void>((resolve) => {
-      componentOnReady(el, () => resolve());
-    });
-
-    jasmine.addMatchers(toHaveNoViolations);
-
-    const a11y = await axe(fixture.nativeElement);
-
-    expect(a11y).toHaveNoViolations();
-  });
-});
-```
 
 #### Testing HTTP Data Services
 
--- a/docs/react/testing/unit-testing/best-practices.md
+++ b/docs/react/testing/unit-testing/best-practices.md
@@ -68,7 +68,7 @@ test('renders the submit button', async () => {
     </IonApp>
   );
 
-  const button = container.querySelector('ion-button');
+  const button = container.querySelector('ion-button')!;
 
   await new Promise<void>((resolve) => componentOnReady(button, () => resolve()));
 
--- a/docs/react/testing/unit-testing/best-practices.md
+++ b/docs/react/testing/unit-testing/best-practices.md
@@ -50,6 +50,24 @@ test('example', async () => {
 
 For more information on `user-event`, see the [user-event documentation](https://testing-library.com/docs/user-event/intro/).
 
-## Import `componentOnReady` for standalone projects
+## Waiting for Components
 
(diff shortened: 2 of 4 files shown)


docs/angular/testing.md
```diff
@@ -178,6 +178,47 @@ describe('PayrolService', () => {
 });
 ```
 
+### Import `componentOnReady` for standalone projects
```
thetaPC: I recommend shortening the title to match the title standards.


```suggestion
### Waiting for Components
```

thetaPC: Move this to the end of `### Pages and Components`, under `## Unit Testing`, because that's where component DOM testing lives and this follows on from that section's closing paragraph about `fixture.nativeElement.querySelector`. Keep it at h3 so it still shows up in the page's Contents sidebar.

docs/angular/testing.md
```diff
@@ -178,6 +178,47 @@ describe('PayrolService', () => {
...
+
+When testing Ionic components, use the exported `componentOnReady` helper from `@ionic/core` instead of calling `el.componentOnReady()` directly. The helper works with both lazy-loaded and custom-element builds, making it more likely the component has finished rendering before making assertions against its rendered DOM or running accessibility tests. See an example here:
```
thetaPC: "More likely" hides a real difference: the lazy-loaded path awaits an actual readiness promise while the custom-element path is just one animation frame, so let's state both paths and the reason the direct call fails.

```suggestion
When testing Ionic components, use the `componentOnReady` helper exported from `@ionic/core` rather than calling `el.componentOnReady()` directly. That method only exists on lazy-loaded elements. Calling it directly throws on custom-element builds, which is what standalone projects use. The helper handles both. It awaits the element's own `componentOnReady()` promise when that exists. Otherwise it waits one animation frame, giving the component's inner contents a chance to render. Wait for the callback before asserting against the rendered DOM or running accessibility tests.
```

docs/angular/testing.md
```diff
@@ -178,6 +178,47 @@ describe('PayrolService', () => {
...
+    });
+
+    jasmine.addMatchers(toHaveNoViolations);
+
+    const a11y = await axe(fixture.nativeElement);
+
+    expect(a11y).toHaveNoViolations();
+  });
+});
+```
```
thetaPC: Trimmed to just the readiness wait so the three lines that matter are visible; jasmine-axe and the unused component variable are a separate concern, and HomePage reads better than Component next to Angular's own decorator.

```suggestion
```tsx
import { ComponentFixture, TestBed } from '@angular/core/testing';
import { componentOnReady } from '@ionic/core';

import { HomePage } from './home.page';

describe('HomePage', () => {
  let fixture: ComponentFixture<HomePage>;

  beforeEach(async () => {
    await TestBed.configureTestingModule({
      imports: [HomePage],
    }).compileComponents();

    fixture = TestBed.createComponent(HomePage);
    fixture.detectChanges();
  });

  it('renders the submit button', async () => {
    const button = fixture.nativeElement.querySelector('ion-button');

    await new Promise<void>((resolve) => componentOnReady(button, () => resolve()));

    expect(button.textContent).toContain('Submit');
  });
});
```
```

File: docs/react/testing/unit-testing/best-practices.md [hunk 4]
```diff
@@ -49,3 +49,7 @@ test('example', async () => {
 ```
 
 For more information on `user-event`, see the [user-event documentation](https://testing-library.com/docs/user-event/intro/).
+
+## Import `componentOnReady` for standalone projects
```
thetaPC:
```suggestion
## Waiting for Components
```

docs/react/testing/unit-testing/best-practices.md
```diff
@@ -49,3 +49,7 @@ test('example', async () => {
...
+
+When testing Ionic components, use the exported `componentOnReady` helper from `@ionic/core` instead of calling `el.componentOnReady()` directly. The helper works with both lazy-loaded and custom-element builds, making it more likely the component has finished rendering before making assertions against its rendered DOM or running accessibility tests.
```
thetaPC: `@ionic/react` only ships the custom elements build, so `el.componentOnReady()` is never present and the standalone/both-builds framing does not apply here; this states the one behavior that's true for every React project.

```suggestion
When you need to wait for an Ionic component to render before asserting against its DOM, use the `componentOnReady` helper exported from `@ionic/core`. Do not call `el.componentOnReady()` directly. `@ionic/react` uses Stencil's custom elements build, where that method does not exist on the element. The helper waits one animation frame instead, giving the component's inner contents a chance to render.
```

docs/react/testing/unit-testing/best-practices.md
thetaPC: We should also include an example like Angular. The example should match the one that I provided for Angular.

docs/vue/testing.md
thetaPC, line 1: Add 'vue/testing' to the Vue category in sidebars.js, after 'vue/platform' to match where Angular puts it. Without it the page builds but is unreachable, and nothing else in the docs links to it.

docs/vue/testing.md
```diff
@@ -0,0 +1,25 @@
...
+</head>
+
+# Testing Ionic Vue
+
+This document provides an overview of how to test an application built with `@ionic/vue`. It covers the basics of testing with Vue, as well as the specific tools and libraries developers can use to test their applications.
+
+## Introduction
+
+Testing is an important part of the development process, and it helps to ensure that an application is working as intended.
+
```
thetaPC: Remove this section as we don't have it in the other pages. We should be consistent.
```suggestion
```

docs/vue/testing.md
```diff
@@ -0,0 +1,25 @@
...
+<head>
+  <title>Vue Unit and End-to-End Testing for Ionic App Components</title>
+  <meta
+    name="description"
+    content="Vue apps created using Ionic are automatically set up for unit and end-to-end testing. Read to learn more about testing tools for Ionic components."
+  />
+</head>
+
+# Testing Ionic Vue
```
thetaPC: We should be consistent with the other pages and not use the frontmatter title.


```suggestion
```

docs/vue/testing.md
```diff
@@ -0,0 +1,25 @@
...
+
+This document provides an overview of how to test an application built with `@ionic/vue`. It covers the basics of testing with Vue, as well as the specific tools and libraries developers can use to test their applications.
```
thetaPC: The old second sentence promised coverage of the specific tools and libraries but the page never named any, so this names the three the Vue starter ships instead.

```suggestion
This document provides an overview of how to test an application built with `@ionic/vue`. Applications generated with the Ionic CLI are set up for unit testing with [Vitest](https://vitest.dev) and [Vue Test Utils](https://test-utils.vuejs.org), and for end-to-end testing with [Cypress](https://www.cypress.io).
```

docs/vue/testing.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Introduction
+
+Testing is an important part of the development process, and it helps to ensure that an application is working as intended.
+
+## Best Practices
+
+### Import `componentOnReady` for standalone projects
```
thetaPC:
```suggestion
### Waiting for Components
```

docs/vue/testing.md
```diff
@@ -0,0 +1,25 @@
...
+
+# Testing Ionic Vue
+
+This document provides an overview of how to test an application built with `@ionic/vue`. It covers the basics of testing with Vue, as well as the specific tools and libraries developers can use to test their applications.
+
+## Introduction
+
+Testing is an important part of the development process, and it helps to ensure that an application is working as intended.
+
+## Best Practices
```
thetaPC: Let's be consistent with the other frameworks:


```suggestion
## Unit Testing
```

docs/vue/testing.md
```diff
@@ -0,0 +1,25 @@
...
+
+### Import `componentOnReady` for standalone projects
+
+When testing Ionic components, use the exported `componentOnReady` helper from `@ionic/core` instead of calling `el.componentOnReady()` directly. The helper works with both lazy-loaded and custom-element builds, making it more likely the component has finished rendering before making assertions against its rendered DOM or running accessibility tests.
```
thetaPC: `@ionic/vue` only ships the custom elements build, so the both-builds hedge doesn't apply here; this states the one behavior Vue apps actually get.

```suggestion
When you need to wait for an Ionic component to render before asserting against its DOM, use the `componentOnReady` helper exported from `@ionic/core`. Do not call `el.componentOnReady()` directly. `@ionic/vue` uses Stencil's custom elements build, where that method does not exist on the element. The helper waits one animation frame instead, giving the component's inner contents a chance to render.
```

thetaPC (changes requested):

docs/react/testing/unit-testing/best-practices.md
Zac-Smucker-Bryan: OK, I tested and added one. Let me know if it needs to be changed.

Zac-Smucker-Bryan:
Thanks for the feedback @thetaPC. I think I addressed all your comments. Let me know what else might need to be changed.

docs/react/testing/unit-testing/best-practices.md
```diff
@@ -49,3 +49,25 @@ test('example', async () => {
...
+test('renders the submit button', async () => {
+  const { container } = render(<App />);
+
+  const button = container.querySelector('ion-button');
+
+  await new Promise<void>((resolve) => componentOnReady(button!, () => resolve()));
+
+  expect(button?.textContent).toContain('Submit');
+});
+```
```
thetaPC, line 77: Wrapped in IonApp to match the first best practice on this page, moved the non-null assertion to the query so the last line reads cleanly, and dropped the vitest import since the other snippets here stay runner agnostic.

```suggestion
```tsx
import { IonApp } from '@ionic/react';
import { render } from '@testing-library/react';
import { componentOnReady } from '@ionic/core';

import Example from './Example';

test('renders the submit button', async () => {
  const { container } = render(
    <IonApp>
      <Example />
    </IonApp>
  );

  const button = container.querySelector('ion-button')!;

  await new Promise<void>((resolve) => componentOnReady(button, () => resolve()));

  expect(button.textContent).toContain('Submit');
});
```
```

File: docs/vue/testing.md [hunk 13]
```diff
@@ -0,0 +1,19 @@
...
+  />
+</head>
+
+This document provides an overview of how to test an application built with `@ionic/vue`. Applications generated with the Ionic CLI are set up for unit testing with [Vitest](https://vitest.dev) and [Vue Test Utils](https://test-utils.vuejs.org), and for end-to-end testing with [Cypress](https://www.cypress.io).
+
+## Unit Testing
+
+### Waiting for Components
+
+When you need to wait for an Ionic component to render before asserting against its DOM, use the `componentOnReady` helper exported from `@ionic/core`. Do not call `el.componentOnReady()` directly. `@ionic/vue` uses Stencil's custom elements build, where that method does not exist on the element. The helper waits one animation frame instead, giving the component's inner contents a chance to render.
```
thetaPC, line 19: Forgot to mention it in the last review. This should also have an example like the other two frameworks. Consistency is very important for our docs.

thetaPC (changes requested):

docs/vue/testing.md
Zac-Smucker-Bryan: Added

Zac-Smucker-Bryan:
PR revised based on feedback

Zac-Smucker-Bryan:
PR revised based on feedback

docs/react/testing/unit-testing/best-practices.md
```diff
@@ -49,3 +49,29 @@ test('example', async () => {
...
+import Example from './Example';
+
+test('renders the submit button', async () => {
+  const { container } = render(
+    <IonApp>
+      <Example />
+    </IonApp>
+  );
+
+  const button = container.querySelector('ion-button');
```
thetaPC: `querySelector` returns `Element | null`, so `button.textContent` on the last line fails under `strictNullChecks`.

```suggestion
  const button = container.querySelector('ion-button')!;
```

docs/vue/testing.md
```diff
@@ -0,0 +1,46 @@
...
+test('renders the submit button', async () => {
+  const wrapper = mount(TestComponent);
+
+  const button = wrapper.element.querySelector('ion-button')!;
+
+  await new Promise<void>((resolve) => componentOnReady(button, () => resolve()));
+
+  expect(button.textContent).toContain('Submit');
+});
+```
```
thetaPC: Dropped the string-template wrapper since it needs Vue's runtime compiler, which isn't in the default Vite build, and switched to `find().element` which is idiomatic VTU and typed non-null.

```suggestion
```ts
import { mount } from '@vue/test-utils';
import { componentOnReady } from '@ionic/core';

import Example from './Example.vue';

test('renders the submit button', async () => {
  const wrapper = mount(Example);

  const button = wrapper.find('ion-button').element;

  await new Promise<void>((resolve) => componentOnReady(button, () => resolve()));

  expect(button.textContent).toContain('Submit');
});
```
```

thetaPC (changes requested):
Almost there!

docs/vue/testing.md
thetaPC: @Zac-Smucker-Bryan why was the [test](https://github.com/ionic-team/ionic-docs/pull/4611/changes#diff-fbcb242e61459f9f9a3c29c253d30057b5653d8ffa49dbf2ef17b8b772ba4ed7) removed?

Zac-Smucker-Bryan: @thetaPC When I have been accepting changes from you on github for the tests specifically (including the other test revisions), they are always getting deleted and I add them back (thus why there are so many commits) - it looks like maybe the suggested change didn't have the closing ```? Would github not like that?

Anyway, it was added back in the most recent commit.

thetaPC (approved):
LGTM, I've also updated the PR title to match

docs/vue/testing.md
thetaPC: Good catch, that's exactly what happened! The suggestion had a code fence inside it and GitHub needs the suggestion fence to be longer than any fence it contains, so it terminated early and applied an empty replacement. My fault, not yours.

PR #4616, opened 2026-08-04
chore(react): upgrade to v19

thetaPC:
Issue URL: internal

The site runs React 18.2 and Docusaurus 3.7.0. Several things block or complicate a React 19 upgrade:

- `npm ci` reports an unmet React peer dependency. `@docusaurus/plugin-debug@3.7.0` pins `react-json-view-lite: ^1.2.0`, and no 1.x release of that package supports React 19.
- `@prismicio/react@2` declares `peerDependencies: { react: "^18" }`, which makes `npm install` fail with ERESOLVE on React 19.
- Playground tooltips use `@tippyjs/react`, unmaintained since April 2022. It reads `children.ref`, which React 19 removed, logging "Accessing element.ref was removed in React 19" on pages with playgrounds.
- The lockfile resolves `react-helmet-async` to the upstream package even though `@docusaurus/types` declares the `npm:@slorber/react-helmet-async` alias. Only that fork supports React 19, so this has been installing the wrong package since the 3.7 upgrade.

## What is the new behavior?

React 19.2.8 and Docusaurus 3.10.2, with `npm ci` running clean and zero React peer violations across the tree.

**React 19**
- `react`, `react-dom` to 19.2.8 and `@types/react` to 19
- `@prismicio/react` to 3 and `@prismicio/client` to 7, which renames `getEndpoint` to `getRepositoryEndpoint`
- Replaced `@tippyjs/react` with `@floating-ui/react` (maintained by Tippy's own author) behind a shared `Tooltip` component
- Corrected the stale `react-helmet-async` lockfile entry to the aliased fork
- Type updates for React 19: the removed global `JSX` namespace (`JSX.Element` to `ReactNode`, custom elements declared via module augmentation) and `ReactElement` props defaulting to `unknown`

**Docusaurus 3.10.2**
- All six root `@docusaurus/*` packages bumped together so the family stays on one version. This brings `react-json-view-lite` to 2.5.0 and clears the last peer warning.
- Ported three swizzled components to their 3.10.2 sources: `Layout` (removed `useKeyboardNavigation`, added `ThemeClassNames.layout.main.container`), `EditMetaRow` (upstream `noPrint` so the edit link is hidden when printing), and `LocaleDropdownNavbarItem` (upstream `useLocaleDropdownUtils` refactor)
- Migrated the deprecated `onBrokenMarkdownLinks` to `markdown.hooks.onBrokenMarkdownLinks`
- Adapted `_search.scss` to DocSearch v4, which applies `all: unset` to the search button and therefore discards the ambient `box-sizing`, plus restyles the border, background variable, and shortcut keys
- Gave `PlaygroundTabs` its own `TabItem`, since `@theme/TabItem` now requires Docusaurus's tabs context and threw "useTabsContext() must be used within a Tabs component" on component pages

- [ ] Yes
- [x] No

## Other information

[Preview](https://ionic-docs-git-fw-6462-v3-ionic1.vercel.app/docs)


src/styles/components/_search.scss
```diff
@@ -5,6 +5,7 @@
 
   --docsearch-key-gradient: var(--token-primary-bg-c);
   --docsearch-key-shadow: none;
+  --docsearch-key-background: var(--token-primary-bg-c);
```
ShaneK:
```suggestion
  --docsearch-key-background: var(--token-primary-bg-c);
  --docsearch-search-button-background: var(--token-primary-bg-c);
```

Kind of a nit, but I think the button background got missed here. DocSearch v4 rewrote `.DocSearch-Button` to `all: unset` plus `background-color: var(--docsearch-search-button-background)`, so it doesn't read `--docsearch-searchbox-background` anymore.

I checked the preview against production in dark mode. Production renders the button at `rgb(3, 6, 11)` on a `rgb(3, 6, 11)` navbar so it blends in, and the preview renders it at `rgb(21, 23, 42)` on that same `rgb(3, 6, 11)` navbar, so you get a lighter box sitting on a darker bar. Light mode hides the problem because v4 happens to default that variable to `#fff`, which is what `--token-primary-bg-c` already is.

The suggestion also drops `--docsearch-key-gradient` and `--docsearch-key-shadow` since neither one exists in v4 anymore, and `--docsearch-searchbox-height` a few lines up is dead for the same reason. Worth keeping `--docsearch-searchbox-background` though, that one still drives the modal's form.

CleanShot 2026-08-05 at 10 21 02@2x

src/theme/Layout/index.tsx
```diff
@@ -44,7 +41,12 @@ export default function Layout(props: Props): ReactNode {
 
       <div
         id={SkipToContentFallbackId}
-        className={clsx(ThemeClassNames.wrapper.main, styles.mainWrapper, wrapperClassName)}
+        className={clsx(
```
ShaneK, line 44: The README in this folder says swizzled components should go in the prettier ignore so they stay diffable against upstream, but only `DocItem` and `DocRoot` actually made it in there. So this file gets reformatted to repo style, which I think is most of the reason these three had to be hand-transcribed from upstream instead of applied as a patch. Telling that `DocItem/Layout` is the one that kept upstream formatting and also the one that needed no work this time.

Might be worth adding the rest of the swizzle dirs to `.prettierignore` so the next Docusaurus bump is closer to a mechanical diff. Follow-up though, definitely not this PR.

ShaneK (approved):
Looks good to me, I just had some suggestions

src/styles/components/_search.scss
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4616/changes/8cc77237752adf582f87b2cbf2495e0b5f38b696

src/theme/Layout/index.tsx
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4617

PR #4617, opened 2026-08-05
chore(prettier): ignore ejected swizzles, not wrapped ones

thetaPC:
Issue URL: <ticket>

`src/theme/README.md` says swizzled components should be added to the prettier ignore so they stay diffable against `@docusaurus/theme-classic`, but only `DocItem` and `DocRoot` were listed.

## What is the new behavior?

- `.prettierignore` lists the seven ejected components. Wrapped components (using `@theme-original`) are our own code, so they keep repo formatting: `DocSidebar`, `TOC`, `MDXComponents/index.tsx`, and `DocRoot/Layout/Main`.
- The seven ejected files are reformatted to upstream's prettier settings.

- [ ] Yes
- [x] No

## Other information

N/A

### DIFF
--- a/src/theme/README.md
+++ b/src/theme/README.md
@@ -3,4 +3,4 @@
 This folder is used to override the base docusaurus theme. It houses [swizzled components](https://docusaurus.io/docs/swizzling). Components should NOT be swizzled unless absolutely necessary to allow for changes in future versions. If it is possible to shallow swizzle a component using the `@theme-original` alias, then that should be heavily considered. Swizzled components should be added to the prettier ignore and all code updates should be marked with comments to allow more seamless version updating. The styles file for components that have been unsafely swizzle should absolutely not be edited. All styling should be done from the [component partials](/src/styles/components).
 
 - Original theme: [`@docusaurus/theme-classic`](https://docusaurus.io/docs/api/themes/@docusaurus/theme-classic)
-- [Original theme source](https://github.com/facebook/docusaurus/tree/26ae4164d6f90c231c6687363a3907b5f9f172b8/packages/docusaurus-theme-classic/src/theme)
+- [Original theme source](https://github.com/facebook/docusaurus/tree/main/packages/docusaurus-theme-classic/src/theme)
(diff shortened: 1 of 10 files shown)


src/theme/README.md
```diff
@@ -3,4 +3,4 @@
 This folder is used to override the base docusaurus theme. It houses [swizzled components](https://docusaurus.io/docs/swizzling). Components should NOT be swizzled unless absolutely necessary to allow for changes in future versions. If it is possible to shallow swizzle a component using the `@theme-original` alias, then that should be heavily considered. Swizzled components should be added to the prettier ignore and all code updates should be marked with comments to allow more seamless version updating. The styles file for components that have been unsafely swizzle should absolutely not be edited. All styling should be done from the [component partials](/src/styles/components).
```
ShaneK: I think this line needs to move with the PR. Right now it says all swizzled components go in the prettier ignore, but the new rule is only the ejected ones, with wrapped ones deliberately left on repo formatting. The comment block you added in `.prettierignore` points here by name, so it's sending people to the one place that still describes the old behavior.

Worth getting right because the per-file list is manual now. Nothing in CI covers `src/theme`, so if someone ejects a component and forgets the entry, lint just reformats it and they commit the diff without noticing.

Something like: wrapped components import from `@theme-original` and are our own code, so they follow the repo's Prettier config. Ejected components are full copies of upstream, so each one gets added to `.prettierignore` by path to keep upstream's formatting and stay diffable against `@docusaurus/theme-classic`.

src/theme/README.md
```diff
@@ -3,4 +3,4 @@
...
 
 - Original theme: [`@docusaurus/theme-classic`](https://docusaurus.io/docs/api/themes/@docusaurus/theme-classic)
-- [Original theme source](https://github.com/facebook/docusaurus/tree/26ae4164d6f90c231c6687363a3907b5f9f172b8/packages/docusaurus-theme-classic/src/theme)
+- [Original theme source](https://github.com/facebook/docusaurus/tree/main/packages/docusaurus-theme-classic/src/theme)
```
ShaneK:
```suggestion
- [Original theme source](https://github.com/facebook/docusaurus/tree/v3.10.2/packages/docusaurus-theme-classic/src/theme)
```

Would a release tag work better here than `main`? The lockfile has `@docusaurus/core` at 3.10.2, so that tag matches what we're actually diffing against, and it gives a natural thing to bump when we upgrade. The `main` link will drift away from whatever version we're on.

Not a big deal either way, the old pin was from 2023 and never got bumped so it was already stale, and the per-file `@link` headers all point at `main` anyway. Up to you!

ShaneK (approved):
LGTM! Just some nits

src/theme/README.md
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4617/changes/310653008c383be6fbbdcdb1d2898b46c4a3956b

src/theme/README.md
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4617/changes/ec4ad77bd1981718d3423761789d1f663caf0900

PR #4619, opened 2026-08-06
chore(docusaurus): enable faster

thetaPC:
Issue URL: internal

The docs build with the default Docusaurus toolchain: Webpack for bundling, Babel for transpiling, Terser for JS minification, and html-minifier-terser for HTML. A full English build takes about 91s wall clock and 166s of CPU.

Docusaurus Faster was experimental until 3.10, which marks it stable and renames the config flag. It becomes the default in v4 and is already used for newly initialized v3 sites, so staying on the old toolchain leaves us behind on both build times and v4 readiness.

## What is the new behavior?

Docusaurus Faster is enabled, swapping in Rspack, SWC, and Lightning CSS.

- Adds `@docusaurus/faster` and sets `future.faster: true`
- Enables `future.v4.removeLegacyPostBuildHeadAttribute`, which the `ssgWorkerThreads` part of `faster` requires. Nothing in this repo reads the legacy `head` argument in `postBuild`, so this is inert for us
- Deletes `babel.config.js`, which SWC replaces and which Docusaurus warns about while it is still present

Build time drops from 91s to 21s wall clock, and CPU from 166s to 51s.

SWC's HTML minifier is spec-strict where html-minifier-terser was permissive, so enabling Faster surfaced 16 pages of pre-existing invalid markup. Those are fixed here rather than suppressed, taking the build from 16 SSG warnings to 0:

| Cause | Pages |
| --- | --- |
| A URL used as its own link text, which Markdown autolinked into a nested anchor | The 6 Swiper migration pages, current and v7 |
| An email address used as its own link text, same mechanism | Code of conduct, current and v7 |
| A multi-line raw `<p>`, which MDX gave a second paragraph inside | Components and glossary (current and v7), JavaScript overview |
| The above, plus an anchor starting the paragraph | CSS Shadow Parts, current and v7 |
| A code fence that closed early, rendering markup whose inner `<div>` landed inside a `<p>` | v7 Dynamic Font Scaling |

Link text is now descriptive rather than a raw URL, which also makes these links usable with a screen reader, and the surrounding prose uses "refer to" in place of "see".

**`.prettierignore` gained an entry for the versioned glossaries, and it is not a suppression.** The v7 glossary was prettier-clean before this PR. Each definition is a single line of prose inside a JSX `<section>`, and prettier's mdx parser reflows those children so that link text lands on its own line. MDX then wraps that text in a paragraph, producing `<a><p>Android SDK</p></a>`, which is the exact invalid markup this PR removes. Running `npm run lint` without the entry would undo the fix, so the ignore is what makes it durable. The current-version glossary has been ignored for the same reason for some time; the rationale is now written above both entries in the file. For the same underlying reason, verify markup fixes in these files with `npx prettier --check <file>` rather than relying on a build alone.

- [ ] Yes
- [x] No

## Other information

**`npm run clear` is less forgiving now.** `future.faster` includes `rspackPersistentCache`, and deleting `node_modules/.cache` while a dev server is running makes Rspack panic and abort the process instead of falling back to a cold rebuild. Restarting `npm start` recovers fully. This only happens if the cache is destroyed underneath a live compile, so it is a heads-up rather than a blocker.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -8,6 +8,7 @@ docs/native
 versioned_docs/version-v*/native
 docs/cli/commands
 docs/reference/glossary.md
+versioned_docs/version-v*/reference/glossary.md
 
 static/code/stackblitz
 
--- a/.prettierignore
+++ b/.prettierignore
@@ -7,6 +7,10 @@ docs/api
 docs/native
 versioned_docs/version-v*/native
 docs/cli/commands
+# Each definition in these files is one line of prose inside a JSX <section>.
+# Prettier's mdx parser reflows those children and moves link text onto its own
+# line, which MDX then wraps in a paragraph, rendering invalid HTML such as
+# <a><p>Android SDK</p></a>. Formatting these files reintroduces that markup.
 docs/reference/glossary.md
 versioned_docs/version-v*/reference/glossary.md
 
--- a/docs/angular/slides.md
+++ b/docs/angular/slides.md
@@ -340,7 +340,7 @@ Effects such as Cube or Fade can be used in Swiper Element with no additional im
 ```
 
 :::note
-For more information on effects in Swiper, please refer to the <a href="https://swiperjs.com/swiper-api#fade-effect" target="_blank" rel="noopener noreferrer">Swiper API effects documentation</a>.
+For more information on effects in Swiper, please refer to the <a href="https://swiperjs.com/swiper-api#fade-effect" target="_blank" rel="noopener noreferrer">Swiper API fade effect documentation</a>.
 :::
 
 ## Wrap Up
--- a/docs/angular/slides.md
+++ b/docs/angular/slides.md
@@ -71,7 +71,7 @@ From there, we just have to replace `ion-slides` elements with `swiper-container
 
 By default, make sure you import the `register` function from `swiper/element/bundle`. This uses the bundled version of Swiper, which automatically includes all modules and stylesheets needed to run Swiper's various features.
 
-If you would like to use the Core version instead, which does not include additional modules automatically, see <a href="https://swiperjs.com/element#core-version-and-modules" target="_blank" rel="noopener noreferrer">https://swiperjs.com/element#core-version-and-modules</a>. The rest of this migration guide will assume you are using the bundled version.
+If you would like to use the Core version instead, which does not include additional modules automatically, refer to <a href="https://swiperjs.com/element#core-version-and-modules" target="_blank" rel="noopener noreferrer">Swiper's core version and modules documentation</a>. The rest of this migration guide will assume you are using the bundled version.
 
 ## Swiping with Style
 
@@ -93,7 +93,7 @@ If you were using the CSS custom properties found on `ion-slides`, below is a li
 | `--scroll-bar-background`          | `--swiper-scrollbar-bg-color`               |
 | `--scroll-bar-background-active`   | `--swiper-scrollbar-drag-bg-color`          |
 
-For additional custom CSS, because Swiper Element uses Shadow DOM encapsulation, styles will need to be injected into the Shadow DOM scope. See <a href="https://swiperjs.com/element#injecting-styles" target="_blank" rel="noopener noreferrer">https://swiperjs.com/element#injecting-styles</a> for instructions.
+For additional custom CSS, because Swiper Element uses Shadow DOM encapsulation, styles will need to be injected into the Shadow DOM scope. Refer to <a href="https://swiperjs.com/element#injecting-styles" target="_blank" rel="noopener noreferrer">Swiper's guide on injecting styles</a> for instructions.
 
 ### Additional `ion-slides` Styles
 
@@ -227,7 +227,7 @@ Below is a full list of property changes when going from `ion-slides` to Swiper 
 | pager   | Use the `pagination` property instead.                                                                                                  |
 
 :::note
-All properties available in Swiper Element can be found at <a href="https://swiperjs.com/swiper-api#parameters" target="_blank" rel="noopener noreferrer">https://swiperjs.com/swiper-api#parameters</a>.
+All properties available in Swiper Element can be found in the <a href="https://swiperjs.com/swiper-api#parameters" target="_blank" rel="noopener noreferrer">Swiper API parameters documentation</a>.
 :::
 
 ## Events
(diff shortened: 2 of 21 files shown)


versioned_docs/version-v7/reference/glossary.md
```diff
@@ -98,17 +60,7 @@ title: Glossary
...
-    interacting with a program. The common command-line app for a Mac user is the Terminal app, and Windows users often
-    use Command Prompt. The Ionic community often uses this term to refer to{' '}
-    <a href="https://ionicframework.com/docs/cli">Ionic's CLI</a>. Ionic's CLI can be used for a number of things, such
-    as creating production builds of an app, running the development server, and accessing{' '}
-    <a href="https://ionic.io/appflow" target="_blank">
-      Ionic commercial services
-    </a>
-    .
-  </p>
+  A CLI, or <strong>C</strong>ommand-<strong>L</strong>ine <strong>I</strong>nterface, is a text-based interface for interacting with a program. The common command-line app for a Mac user is the Terminal app, and Windows users often use Command Prompt. The Ionic community often uses this term to refer to <a href="https://ionicframework.com/docs/cli">Ionic's CLI</a>. Ionic's CLI can be used for a number of things, such as creating production builds of an app, running the development server, and accessing <a href="https://ionic.io/appflow" target="_blank">Ionic commercial services</a> .
```
ShaneK: The structural fix on this page is a solid improvement, those definitions used to render with the leading link on its own line. One thing left over from it though: collapsing the `<p>` bodies onto one line left a space between `</a>` and the punctuation, so this one renders as "Ionic commercial services ." Seven spots in this file have it, searchable as `</a> .` and `</a> ,`. The current-version glossary got the same treatment and has none of them, so it's just this copy.

versioned_docs/version-v7/layout/dynamic-font-scaling.md
```diff
@@ -102,6 +101,7 @@ However, the `em` unit has a compounding effect which can cause issues. In the f
     <div style={{ fontSize: '2em' }}>Child element with 80px</div>
   </div>
 </div>
+```
```
ShaneK, line 104: Is losing the demo here intended? Moving the fence down swallows the second block, which used to render as a working em compounding demo at 20/40/80px. Now it shows up as JSX source inside an `html` fence, so the highlighting is wrong and it isn't copy-pasteable. The current-version page never had this demo, but that could just as easily mean adding it there instead of dropping it here. The demo's own markup was a bit off, which is probably what drew you to it, so fixing that might be the better option. Up to you!

docs/angular/slides.md
```diff
@@ -340,7 +340,7 @@ Effects such as Cube or Fade can be used in Swiper Element with no additional im
 ```
 
 :::note
-For more information on effects in Swiper, please see <a href="https://swiperjs.com/swiper-api#fade-effect" target="_blank" rel="noopener noreferrer">https://swiperjs.com/swiper-api#fade-effect</a>.
+For more information on effects in Swiper, please refer to the <a href="https://swiperjs.com/swiper-api#fade-effect" target="_blank" rel="noopener noreferrer">Swiper API effects documentation</a>.
```
ShaneK:
```suggestion
For more information on effects in Swiper, please refer to the <a href="https://swiperjs.com/swiper-api#fade-effect" target="_blank" rel="noopener noreferrer">Swiper API fade effect documentation</a>.
```

The anchor lands on the Fade Effect section specifically, so the text promises a bit more than the link gives you. The old raw URL at least exposed the fragment. Same on the v7 copy. Nit, and it's the only one of the rewritten links where the text and the destination don't match, the rest are all accurate.

.prettierignore
```diff
@@ -8,6 +8,7 @@ docs/native
 versioned_docs/version-v*/native
 docs/cli/commands
 docs/reference/glossary.md
+versioned_docs/version-v*/reference/glossary.md
```
ShaneK, line 15: This entry is doing more work than it looks. The v7 glossary was prettier-clean before this PR and the edit makes it dirty, so CI needs the ignore. Prettier wants to move the link text onto its own lines, which is the exact shape that produced the nested paragraphs you're removing, so the entry is the right call. Worth a line in the description though if that's possible, since as written it reads like a suppression.

docs/contributing/coc.md
```diff
@@ -8,6 +8,6 @@ We promise to extend courtesy and respect to everyone involved in this project r
 
 If any member of the community violates this code of conduct, the maintainers of the Ionic project may take action, removing issues, comments, and PRs or blocking accounts as deemed appropriate.
 
-If you are subject to or witness unacceptable behavior, or have any other concerns, please email us at <a href="mailto:devrel@ionic.io">devrel@ionic.io</a>.
+If you are subject to or witness unacceptable behavior, or have any other concerns, please <a href="mailto:devrel@ionic.io">email us</a>.
```
ShaneK:
```suggestion
If you are subject to or witness unacceptable behavior, or have any other concerns, please <a href="mailto:devrel@ionic.io">email us at devrel@ionic.io</a>.
```

Was this one intentional? An email address as the text of a `mailto:` link is already self-describing, and there's no invalid markup here for the minifier to catch, so I think this is the one link on the page that didn't need touching. The address is gone from the page now too, so anyone without a mail client configured gets a link that does nothing and no text to copy, which I'd rather avoid on the reporting path specifically.

The line just below still reads "Please click here to review to Ionic's full code of conduct", which has the non-descriptive link text plus a `review to` typo. Might be worth grabbing while you're in here.

ShaneK (changes requested):
Awesome work! I found some things I'd like to point out for your consideration before we merge this, but overall great work 🎉

versioned_docs/version-v7/reference/glossary.md
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4619/changes/b24d1eac7d96c6797635862549d8cebe0f9ff511

versioned_docs/version-v7/layout/dynamic-font-scaling.md
thetaPC: Yes, it's actually an issue that was fixed for [v8](https://github.com/ionic-team/ionic-docs/pull/4428) but v7 was forgotten.

docs/angular/slides.md
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4619/changes/504f53df30d406bef259c9c5e5be8006075c493e

.prettierignore
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4619/changes/074d705bee0d50e52c35d57fea467107bb39ec37

docs/contributing/coc.md
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4619/changes/dfa6d9cc359b43312ccf8a9184f1283560929bc7

Good call, keeping the address visible. Markdown autolinks bare email addresses into `mailto:` links automatically, so dropping the hand-written anchor keeps the [working link](https://ionic-docs-git-fw-7656-ionic1.vercel.app/docs/contributing/coc) and the copyable text, and removes the nesting that came from the address being both the href and the link text.

ShaneK (approved):
LGTM! Great work 🎉

PR #4622, opened 2026-08-06
chore(ci): check translation keys on pull requests

thetaPC:
Issue URL: N/A

No PR check builds a non-default locale. `build:preview` is `docusaurus build --locale en` and only `build:production` builds all locales, so anything that fails specifically in `ja` gets through review and breaks the main deploy instead.

That just [happened](https://github.com/ionic-team/ionic-docs/pull/4621): the Docusaurus 3.10 upgrade added a validation that rejects duplicate sidebar translation keys, our three "Build Your First App" categories collide on one key, and every check on that PR was green because none of them touched `ja`.

## What is the new behavior?

A `Check Translation Keys` step runs on pull requests, before `Check Diff`:

```yaml
- name: Check Translation Keys
  if: strategy.job-index == 0
  shell: bash
  run: |
    trap 'rm -rf i18n' EXIT
    npm run generate-markdown
    npx docusaurus write-translations --locale ja
```

`write-translations` extracts translatable strings through the same code path the localized build uses, so it hits the same validation, but it never compiles anything. It caught the real failure in 2 seconds, versus minutes for `docusaurus build --locale ja`.

The workflow is also renamed from `Install Dependencies, Lint` to `Validate Docs`. The old name was already stale, since it didn't mention Spell Check, and naming it after the steps is what let it drift.

- [ ] Yes
- [x] No

## Other information

How to test:

The last two commits on this branch differ only by a deliberately duplicated sidebar category, so comparing their runs shows both sides of the check:

| commit | change | result |
| --- | --- | --- |
| `6f8c4db8a` | adds a second category labelled "Techniques" | [fails](https://github.com/ionic-team/ionic-docs/actions/runs/31219316074/job/93000077180) |
| `ee6e0446d` | reverts it | [passes](https://github.com/ionic-team/ionic-docs/actions/runs/31219758924) |

The failing run reports the problem directly:

```
[ERROR] Error: Multiple docs sidebar items produce the same translation key.
- `sidebar.docs.category.Techniques`: 2 duplicates found:
  - Techniques (The label for category 'Techniques' in sidebar 'docs')
  - Techniques (The label for category 'Techniques' in sidebar 'docs')
```


.github/workflows/actions/check-translations/action.yml
thetaPC, line 1: Made a new file in hopes that one day we can add more checks.

.github/workflows/CI.yml
```diff
@@ -13,22 +13,28 @@ jobs:
...
+        # Build processes work differently across operating systems, so the 
+        # checks run on each one to make sure anyone can contribute to the
+        # docs.
         os: [windows-latest, macOS-latest]
 
     steps:
     - uses: actions/checkout@v4
-    - name: Use Node.js ${{ matrix.node_version }}
+    - name: ⚙️ Use Node.js ${{ matrix.node_version }}
       uses: actions/setup-node@v4
```
gnbm, line 24: @thetaPC these actions should be pinned with the commit hash

gnbm (dismissed):

.github/workflows/CI.yml
thetaPC: I'll be making those in a separate PR.

ShaneK (approved):
LGTM!

PR #4633, opened 2026-08-12
chore(docs): remove obsolete TODO and placeholder comments

thetaPC:
Issue URL: N/A

Five commented-out notes are scattered through the docs, left as reminders to come back and write something later. Two are section stubs in `datetime-button.md`, one is a block in `item.md` describing a CSS variable, one asks whether the CLI help section needs a screenshot, and two are placeholders where a table of contents used to be generated inline.

All five have been overtaken by changes elsewhere, so none of them describe work that still needs doing.

## What is the new behavior?

Removes the five notes from the current docs and their v7 copies. These are MDX comments, so nothing changes in the rendered pages.

Each was checked against its current replacement before removal:

- `item.md` described enabling the detail arrow on `md` items with `--item-detail-push-show`. That variable no longer exists in the framework. The `detail` property does this now, and the prose directly above the comment already documents it.
- `datetime-button.md` had empty `Customization`, `Buttons`, and `Theming` stubs. The component exposes one CSS shadow part and no custom properties, both already listed by the auto-generated tables on the page.
- `cli.md` asked whether to add an image of the `--help` output. That output is generated as real text under `docs/cli/commands/`, which stays current as the CLI changes. A static image would go stale, would not be translatable or searchable, and would contradict the note above it explaining that help output differs per project type.
- `versioning.md` and `what-are-progressive-web-apps.md` had `TOC goes here` placeholders from the pre-Docusaurus generator. Docusaurus renders the table of contents in the sidebar, and neither page sets `hide_table_of_contents`.

- [ ] Yes
- [x] No

## Other information

N/A


docs/cli.md
```diff
@@ -37,8 +37,6 @@ Be sure to run `ionic <command> --help` in your project directory.
 For some commands, such as `ionic serve`, the help documentation is contextual to the type of your project, e.g. React vs Angular.
 :::
 
-{/* TODO: image? */}
-
 ## Architecture
```
ShaneK, line 40: While you're in here, a few playground index files still open with a bare `# TODO FW-4608` heading instead of a comment, which is probably why the sweep missed them. [The v8 range basic playground](https://github.com/ionic-team/ionic-docs/blob/9b019d337d69deea745c8ab1f73db579ff8f9f2b/static/usage/v8/range/basic/index.md#L1) is one, and there are two more under the v7 usage folder.

They're [kept for the JP site](https://github.com/ionic-team/ionic-docs/pull/3004) rather than the English docs, so I'd leave the files alone. FW-4608 is ROU-7122 now and it's Discarded, so the marker could just say why they're there instead of pointing at a dead ticket.

Also, [the gestures page](https://github.com/ionic-team/ionic-docs/blob/9b019d337d69deea745c8ab1f73db579ff8f9f2b/docs/utilities/gestures.md#L122) still says "Full React wrappers are coming soon!" directly above a sample that already imports `createGesture` from `@ionic/react`. I couldn't find a ticket for those wrappers, so I don't know if that's still planned or just got forgotten about. Up to you whether you want it in scope.

ShaneK (approved):
LGTM! A minor suggestion, but a nit at best

docs/cli.md
thetaPC: I'm going to leave the FW-4608 as is as a reminder to come back and investigate those pages. Apparently JP is no longer referencing them so we should be able to remove them without issues. I'll add it as a todo for the MDX migration. I did remove the coming soon text. https://github.com/ionic-team/ionic-docs/pull/4633/changes/e9d8b8c098b9ebc8ff37274e74f5648f12f7d41c

PR #4634, opened 2026-08-13
chore(deps): upgrade dependencies and drop unused ones

thetaPC:
Issue URL: N/A

Sixteen dependencies are behind, several by multiple majors. Three are declared but unused: `html-loader` has no references anywhere, `@docusaurus/plugin-client-redirects` is never registered in the plugins array (redirects are handled in `vercel.json`), and `@docusaurus/mdx-loader` is not imported and already comes in through `@docusaurus/core` and `@docusaurus/preset-classic`.

`node-fetch` is imported by `docusaurus.config.js`, the component API plugin, and two build scripts, but it is not declared in `package.json`. It resolves only because `cspell@6` happens to pull it in transitively.

CI installs with `--legacy-peer-deps`, added in 2023 when the workflow was first created. The matrix declares a `node_version` axis with a single value.

## What is the new behavior?

Removes the three unused dependencies and upgrades the rest. Minor and patch bumps for `@mdx-js/react`, `@prismicio/client`, `@stackblitz/sdk`, `docusaurus-plugin-copy-page-button`, `prism-react-renderer`, `sass`, `semver`, and `typescript`. Majors for `clsx` 2, `fs-extra` 11, `modern-normalize` 3, `@ionic-internal/ionic-ds` 8, and `cspell` 10, each in its own commit.

Replaces `node-fetch` with the native `fetch` available since Node 18. This had to land before the cspell bump, because `cspell@6` was the only thing providing the package and upgrading it removes `node-fetch` from the tree entirely.

Requires Node 24. `cspell@10` needs at least 22.18, and this repo is self contained so there are no consumers to consider.

Adds `--gitignore` to the spellcheck script so it stops checking build output. Locally this drops the file count from 15471 to 7942, all of the difference being generated files under `build/`. CI never built, so it was already checking the smaller set; this makes local runs match CI.

Drops `--legacy-peer-deps` from CI and removes the single value `node_version` matrix axis.

- [ ] Yes
- [x] No

## Other information

Two upgrades change rendering rather than just versions:

`modern-normalize` v3 drops `hr { height: 0 }`. Because `theme.customCss` is injected after Infima, that rule was overriding `--ifm-hr-height` and flattening every horizontal rule. Rules now render at their intended 1px. This affects the 48 Capacitor plugin pages, which are the only pages in the site containing an `hr`. Compare the [Action Sheet page in production](https://ionicframework.com/docs/native/action-sheet), where the separators between API entries are missing, against the [same page in this branch's preview](https://ionic-docs-git-packages-ionic1.vercel.app/docs/native/action-sheet), where they are visible.

Anything whose blast radius reaches beyond dependency resolution was deliberately left out, so that each one gets a reviewable diff of its own rather than being buried in this one.

`prettier` 2 to 3 and `@ionic/prettier-config` 3 to 4 move together, since the config package exists to pair with prettier 3. That upgrade reformats every file in the repo, so the diff would be thousands of lines of whitespace with the real changes hidden inside it. It needs a PR where the reformat is the only thing under review.

`typescript` 5 to 7 is a compiler major. Nothing typechecks today (`tsc` is in no npm script and not in CI, and the repo has 42 pre-existing errors on both 5.3 and 5.9), so upgrading it means first deciding whether typechecking should be enforced at all. That is a separate conversation from keeping dependencies current.

`crowdin` is deprecated on npm, but the situation is more involved than a version swap. The package ships no binary and is never imported, while `crowdin.yml` is configured for the official CLI (`@crowdin/cli`), which is not declared anywhere. The npm deprecation notice points at `@crowdin/crowdin-api-client`, which is a library with no CLI and would not resolve the `crowdin upload` command either. Before changing it, someone needs to confirm how that command currently resolves on Vercel, since it is the path that keeps the Japanese docs in sync.


.github/workflows/CI.yml
```diff
@@ -12,20 +12,19 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        node_version: [20]
-        # Build processes work differently across operating systems, so the 
+        # Build processes work differently across operating systems, so the
         # checks run on each one to make sure anyone can contribute to the
         # docs.
         os: [windows-latest, macOS-latest]
```
brandyscarney, line 18: The node version is useful for testing that the repo works in multiple node versions, are we sure we do not want to keep that?

For example: https://github.com/ionic-team/ionic-cli/blob/develop/.github/workflows/ci.yml#L18-L20

package.json
brandyscarney, line 1: There are a lot of vulnerabilities and a deprecation, can we do anything about these?

Image

brandyscarney (changes requested):
Looks good, just curious if we can fix some of the install vulnerabilities.

.github/workflows/CI.yml
thetaPC: We never tested it out so I didn't think we would mind removing it. Plus, we would line up with Vercel's recommendation to be on 24. But if you feel like it can be very useful then I could add it back in and test out the supported node versions that [Vercel supports](https://vercel.com/docs/functions/runtimes/node-js/node-js-versions). Let me know.

package.json
thetaPC: Managed to get it lower: https://github.com/ionic-team/ionic-docs/pull/4634/changes/c630f25d08fc3d115c1ba5af67fce2e93f1c2478

The rest:
- The remaining 26 sit inside `@docusaurus/core`'s own dependency tree (`image-size`, `serialize-javascript`, `uuid`), and Docusaurus is already at its latest release (3.10.2), so there is nothing on our side to bump.
- The last one (`got`) comes from `crowdin`, which is deprecated. We're planning to move to the maintained `@crowdin/cli` in a separate PR, which clears that advisory and the deprecation warning together.

.github/workflows/CI.yml
brandyscarney: It's fine I'm not sure we need to test this repo with multiple node versions. We can always add it back in if we need to.

package.json
brandyscarney: Thank you!

brandyscarney (approved):
LGTM, thank you!

PR #4638, opened 2026-08-13
chore(prettierignore): add versioned api

thetaPC:
Issue URL: N/A

Prettier ignore has the API page added but not the versioned API page.

## What is the new behavior?

- Added the versioned API page to prettier ignore

- [ ] Yes
- [x] No

## Other information

N/A


ShaneK (approved):
LGTM!

PR #4628, opened 2026-08-10
chore(docs): migrate admonition titles and heading IDs to MDX syntax

thetaPC:
Issue URL: internal

## What is the current behavior?

We use two proprietary Docusaurus syntaxes that are being phased out:

- **Admonition titles** use the space form, `:::warning My Title`. 43 occurrences across 38 files.
- **Heading IDs** use the classic form, `{#my-id}`. 2 occurrences.

Neither is standard. Docusaurus calls both "proprietary Docusaurus syntax, leading to ecosystem incompatibilities" and recommends migrating:

- https://docusaurus.io/blog/releases/3.10#strict-admonitions
- https://docusaurus.io/blog/releases/3.10#strict-heading-ids

Both only work today because of the MDX v1 compatibility shims in `mdx-loader`. On every build, `preprocessContent()` rewrites our files in memory before MDX ever parses them: `admonitionTitleToDirectiveLabel()` converts `:::warning My Title` to `:::warning[My Title]`, and `escapeMarkdownHeadingIds()` escapes `{#my-id}` to `\{#my-id}` so MDX does not try to evaluate it as an expression.

Those shims are the `markdown.mdx1Compat` options, which Docusaurus v4 disables by default via `future.v4.mdx1CompatDisabledByDefault`.

Separately, `troubleshooting/runtime.md` has a malformed admonition where the closing `:::` sits mid-sentence:

```md
:::note
This flag is automatically included when creating an Ionic app via
:::
the Ionic CLI.
```

That renders as a note ending in "via", followed by an orphaned "the Ionic CLI." paragraph outside it. It dates back to at least v5.

## What is the new behavior?

Both syntaxes are migrated to the standard forms, so the files no longer depend on the compat shims.

- `:::type My Title` becomes `:::type[My Title]`, the Markdown Directive syntax
- `{#my-id}` becomes `{/* #my-id */}`, a native MDX comment

The malformed note is also repaired so the sentence stays intact inside the admonition.

40 files changed, +45/-45. Scope is limited to the live docs trees, `docs/` and `versioned_docs/version-v7`. The archived v5 and v6 trees are deliberately untouched, since they are absent from `versions.json` and are served from standalone deployments rather than built from this repo.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

This is part of a series preparing the docs for the `.md` to `.mdx` migration.

### How to test

Review the modified pages below and confirm each admonition still renders with its title and correct styling.

Two specific things to check on the runtime page:

[...]

PR #4639, opened 2026-08-13
chore(api): generate heading IDs with MDX comment syntax

thetaPC:
Issue URL: internal

The component API plugin generates heading IDs using the classic `{#my-id}` syntax: `spinner {#prop-spinner}`

 Docusaurus calls this "proprietary Docusaurus syntax, leading to ecosystem incompatibilities" and [recommends](https://docusaurus.io/blog/releases/3.10#strict-heading-ids) the native MDX comment form instead.. We migrated our hand written docs off this syntax in #4628, but the plugin kept emitting it, so it comes back on every build. On `main` that is **1435 headings across 910 generated partials** (1263 property headings, 172 method headings).

## What is the new behavior?

The plugin emits the MDX comment form: `spinner {/* #prop-spinner */}`

Two lines change, one for properties and one for methods, which covers all 1435 headings. No other partial type is affected: events, parts, slots and custom-props contain tables rather than headings.

- [ ] Yes
- [x] No

## Other information

Part of the mdx migration.

**No anchors change.** Both syntaxes produce the same id, so `#prop-spinner`, `#method-dismiss` and every other API anchor continue to resolve. 

### How to test

Check that each heading renders its name only, with no visible `{/* ... */}` text, and that both the anchor link and the table of contents entry jump to it.

- **Properties:** https://ionic-docs-git-fw-6456-pt4-ionic1.vercel.app/docs/api/loading
  - The `spinner` heading should link to `#prop-spinner`
- **Methods:** https://ionic-docs-git-fw-6456-pt4-ionic1.vercel.app/docs/api/alert
  - The `dismiss` heading should link to `#method-dismiss`
- **Deprecated properties:** https://ionic-docs-git-fw-6456-pt4-ionic1.vercel.app/docs/v7/api/input
  - `accept (deprecated)` should link to `#prop-accept`
  - This is the case where the comment sits next to the `(deprecated)` interpolation on the same line
- **Links into generated anchors:** https://ionic-docs-git-fw-6456-pt4-ionic1.vercel.app/docs/api/datetime#showing-confirmation-buttons
  - In the **Showing Confirmation Buttons** section, the sentence "The default Done and Cancel buttons are already preconfigured to call the `confirm` and `cancel` methods" has two links
  - Both should jump down to those method headings under **Methods**
  - This is the only page in the repo with hand written links into plugin generated anchors

### DIFF
--- a/plugins/docusaurus-plugin-ionic-component-api/index.js
+++ b/plugins/docusaurus-plugin-ionic-component-api/index.js
@@ -213,7 +213,7 @@ ${properties
     }
 
     return `
-### ${prop.name} ${isDeprecated ? '(deprecated)' : ''} {#${propertyHeadingId(prop.name)}}
+### ${prop.name} ${isDeprecated ? '(deprecated)' : ''} {/* #${propertyHeadingId(prop.name)} */}
 
 | | |
 | --- | --- |
@@ -275,7 +275,7 @@ function renderMethods({ methods }) {
 ${methods
   .map(
     (method) => `
-### ${method.name} {#${methodHeadingId(method.name)}}
+### ${method.name} {/* #${methodHeadingId(method.name)} */}
 
 | | |
 | --- | --- |


ShaneK (approved):
LGTM!

PR #4640, opened 2026-08-13
chore(api): generate component API partials as .mdx

thetaPC:
Issue URL: internal

The component API plugin writes its generated partials with a `.md` extension: props, events, methods, parts and slots. Those files are MDX, not CommonMark, so the extension is inaccurate. The sixth partial, custom-props, is already `.mdx`, which suggests this was hit once before and fixed for a single file.

Each API page then imports them by explicit filename, so the extension is repeated across every page, for example importing Props from `@ionic-internal/component-api/v8/button/props.md`.

## What is the new behavior?

The plugin emits `.mdx` for the five remaining partials, and the import specifiers are updated to match.

- 5 lines in `plugins/docusaurus-plugin-ionic-component-api/index.js`
- 910 import specifiers across 181 pages in `docs/` (94) and `versioned_docs/version-v7` (87)

Every doc file change is a specifier swap; there are no other edits.

- [ ] Yes
- [x] No

## Other information

Part of the mdx migration.

### How to test

Check a page where all six partial types have content:

- **Current version:** https://ionic-docs-git-fw-6456-pt4b-ionic1.vercel.app/docs/api/modal
  - Scroll to the bottom and confirm **Properties**, **Events**, **Methods**, **CSS Shadow Parts**, **CSS Custom Properties** and **Slots** all render with real content
- **Versioned tree:** https://ionic-docs-git-fw-6456-pt4b-ionic1.vercel.app/docs/v7/api/select
  - Same six sections; this confirms the version-v7 specifiers were rewritten too

The content, datetime, popover and select pages also populate all six if you want another sample.

### DIFF
--- a/plugins/docusaurus-plugin-ionic-component-api/index.js
+++ b/plugins/docusaurus-plugin-ionic-component-api/index.js
@@ -81,11 +81,37 @@ module.exports = function (context, options) {
          * directory within the plugin directory.
          */
         promises.push(
+          createData(`${basePath}/props.mdx`, data.props),
+          createData(`${basePath}/events.mdx`, data.events),
+          createData(`${basePath}/methods.mdx`, data.methods),
+          createData(`${basePath}/parts.mdx`, data.parts),
+          createData(`${basePath}/custom-props.mdx`, data.customProps),
+          createData(`${basePath}/slots.mdx`, data.slots)
+        );
+
+        /**
+         * TODO(FW-6456): Remove once all branches import the `.mdx` names instead of `.md`.
+         * Transitional: the `.md` names are still imported by pages this repo does not
+         * control on every branch at once.
+         *
+         * - `scripts/i18n.sh` copies the `translation/jp` branch's `docs/` tree into
+         *   `i18n/ja` on every build, and those copies still import `.md`.
+         * - Major branches keep their own api pages. Until each is rewritten, a sync that
+         *   brings this plugin over would otherwise leave them importing names it no longer
+         *   emits.
+         *
+         * Emitted unconditionally on purpose. Guarding on locale only holds on a branch whose
+         * pages have already been rewritten, so it reintroduces exactly the coupling this
+         * avoids. These land in `.docusaurus`, not a docs content root, so they are never
+         * routed as pages and the duplicate basenames cannot collide.
+         *
+         * Remove once `translation/jp` and every live branch import the `.mdx` names.
+         */
+        promises.push(
           createData(`${basePath}/props.md`, data.props),
           createData(`${basePath}/events.md`, data.events),
           createData(`${basePath}/methods.md`, data.methods),
           createData(`${basePath}/parts.md`, data.parts),
-          createData(`${basePath}/custom-props.mdx`, data.customProps),
           createData(`${basePath}/slots.md`, data.slots)
         );
       }
--- a/docs/api/accordion-group.md
+++ b/docs/api/accordion-group.md
@@ -1,12 +1,12 @@
 ---
 title: "ion-accordion-group"
 ---
-import Props from '@ionic-internal/component-api/v9/accordion-group/props.md';
-import Events from '@ionic-internal/component-api/v9/accordion-group/events.md';
-import Methods from '@ionic-internal/component-api/v9/accordion-group/methods.md';
-import Parts from '@ionic-internal/component-api/v9/accordion-group/parts.md';
+import Props from '@ionic-internal/component-api/v9/accordion-group/props.mdx';
+import Events from '@ionic-internal/component-api/v9/accordion-group/events.mdx';
+import Methods from '@ionic-internal/component-api/v9/accordion-group/methods.mdx';
+import Parts from '@ionic-internal/component-api/v9/accordion-group/parts.mdx';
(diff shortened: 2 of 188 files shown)


plugins/docusaurus-plugin-ionic-component-api/index.js
```diff
@@ -81,12 +81,12 @@ module.exports = function (context, options) {
          * directory within the plugin directory.
          */
         promises.push(
-          createData(`${basePath}/props.md`, data.props),
-          createData(`${basePath}/events.md`, data.events),
-          createData(`${basePath}/methods.md`, data.methods),
-          createData(`${basePath}/parts.md`, data.parts),
+          createData(`${basePath}/props.mdx`, data.props),
```
ShaneK, line 84: This one worries me a bit, it breaks the ja build.

Our `i18n.sh` step pulls the `translation/jp` branch's `docs/` tree into `i18n/ja` on every build, and those are whole-file copies with their own import block, still pointing at `props.md` and the rest. Building `--locale ja` on this commit fails with 470 `Cannot find module` errors against the old partial names. English is fine, and previews only build `--locale en`, so nothing here catches it. Production builds both.

The `custom-props` partial needed the same treatment when it flipped in [60443e2](https://github.com/ionic-team/ionic-docs/commit/60443e2d0d32325839f35f8df6926ae550638ffa). There's a jp sync open right now in #4641 and it still has the old imports on all 94 pages, so that one won't cover it either.

Could you keep writing the five `.md` partials alongside the new `.mdx` ones, and drop them once `translation/jp` has caught up? They go to `.docusaurus` rather than a docs content root, so the duplicate basenames won't collide.

ShaneK (changes requested):
I'm worried about the impact of these changes on i18n

plugins/docusaurus-plugin-ionic-component-api/index.js
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4640/changes/98bfab70d14b3404f5917e5cff7a20425aae95cb and https://github.com/ionic-team/ionic-docs/pull/4640/changes/df557ba4688db809f4aa49e369f3e88b410399bd

plugins/docusaurus-plugin-ionic-component-api/index.js
```diff
@@ -81,13 +81,39 @@ module.exports = function (context, options) {
...
```
ShaneK, line 84: The plugin change is global but the page rewrites only live on main, so this leaves the major branches inconsistent. Their api pages still import the `.md` partials, both in `docs/api` and in the versioned copies, and `major-10.0` only gets this through `major-9.0` since a direct sync from main won't run.

On this line the base has `v8` with `.md`, main has `v8` with `.mdx`, and `major-9.0` has `v9` with `.md`, so resolving it the way we normally do, where the major branch wins, quietly drops the migration and leaves those pages on `.md` while the plugin only emits `.mdx` for en. For this reason, I believe this merge will require extra care and is kind of dangerous to do at this time. Could you try to do companion commits on the major branches alongside this one? Emitting the `.md` names unconditionally would work too and it's closer to what I asked for originally, since it's safe whatever order things merge in.

To see what I'm concerned about, it may be worth creating a worktree off of major-9.0 and try merging this branch into it and see what gets left out.

plugins/docusaurus-plugin-ionic-component-api/index.js
```diff
@@ -81,13 +81,39 @@ module.exports = function (context, options) {
...
+         * build renders translated current-version pages (importing `.md`) alongside
+         * untranslated versioned pages that fall back to English (importing `.mdx`). Both
+         * names must exist within that pass, which is why the extension itself cannot be
+         * switched per locale, only supplemented.
+         *
+         * Remove once `translation/jp` imports the `.mdx` partials. These are written to
+         * `.docusaurus` rather than a docs content root, so they are never routed as pages
+         * and the duplicate basenames cannot collide.
+         */
+        if (context.i18n.currentLocale === 'ja') {
```
ShaneK, line 108: Did you get a chance to run a ja build against this?

Could you also file something to track dropping these? There's no issue linked from the comment and `translation/jp` moves slowly enough that I think this outlives the PR by a good while.

ShaneK (approved):
Wearily approved with comments. Still not sure right before v9 releases is the best time to merge this one, in case of issues caused by it.

plugins/docusaurus-plugin-ionic-component-api/index.js
thetaPC: I ended up adding both mdx and md until we can fully migrate. This would make sure that JP and any branch that isn't main will not break. I added a not to remove the duplicate files once we wrap up work.

https://github.com/ionic-team/ionic-docs/pull/4640/changes/497e9b805bf1042131c041ad27cc7eee3ba50d9b#diff-f3a8d253eef1b60367839f3396ef3826a35d4e5b1b10c4df17ea9a6b0ac37bd9

plugins/docusaurus-plugin-ionic-component-api/index.js
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4640#discussion_r3845829877

ShaneK (approved):
LGTM!

PR #4642, opened 2026-08-14
docs(v9): update demos and support info to use v9 version

brandyscarney:
- Updates all demos to point to v9
- Updates StackBlitz versions to v9
- Changes the npm tag to use `latest` for the docusaurus build
- Updates support table with v9 release date & v8 maintenance/support end dates
- Updates playground template to include v9
- Updates old references to v8

### DIFF
--- a/static/code/stackblitz/v9/html/package-lock.json
+++ b/static/code/stackblitz/v9/html/package-lock.json
@@ -6,7 +6,7 @@
     "": {
       "name": "html-starter",
       "dependencies": {
-        "@ionic/core": "8.8.16-dev.11785274070.1b3387c9",
+        "@ionic/core": "9.0.0",
         "ionicons": "8.1.0"
       },
       "devDependencies": {
@@ -16,7 +16,7 @@
       }
     },
     "node_modules/@ionic/core": {
-      "version": "8.8.16-dev.11785274070.1b3387c9",
+      "version": "9.0.0",
       "license": "MIT",
       "dependencies": {
         "@stencil/core": "^4.43.5",
--- a/versioned_docs/version-v8/reference/support.md
+++ b/versioned_docs/version-v8/reference/support.md
@@ -22,7 +22,7 @@ The current status of each Ionic Framework version is:
 
 | Version |     Status      |   Released   | Maintenance Ends | Ext. Support Ends |
 | :-----: | :-------------: | :----------: | :--------------: | :---------------: |
-|   V8    | **Maintenance** | Apr 17, 2024 |   Feb 17, 2026   |   Aug 17, 2027    |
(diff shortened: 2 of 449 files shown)


versioned_docs/version-v8/reference/support.md
```diff
@@ -22,7 +22,7 @@ The current status of each Ionic Framework version is:
 
 | Version |     Status      |   Released   | Maintenance Ends | Ext. Support Ends |
 | :-----: | :-------------: | :----------: | :--------------: | :---------------: |
-|   V8    | **Maintenance** | Apr 17, 2024 |       TBD        |        TBD        |
+|   V8    | **Maintenance** | Apr 17, 2024 |   Feb 17, 2026   |   Aug 17, 2027    |
```
ShaneK:
```suggestion
|   V8    | **Maintenance** | Apr 17, 2024 |   Feb 17, 2027   |   Aug 17, 2027    |
```

The v9 copy of this table says `Feb 17, 2027` for the same row. I think that's the right one, since maintenance ends 6 months after the next major everywhere else in the table and both files already agree on `Aug 17, 2027` for the ext support date.

static/code/stackblitz/v9/html/package-lock.json
```diff
@@ -16,7 +16,7 @@
       }
     },
     "node_modules/@ionic/core": {
-      "version": "8.8.16-dev.11785274070.1b3387c9",
+      "version": "9.0.0",
```
ShaneK, line 19: Nothing to do here, just noting it for later. The lockfiles only moved the `@ionic/*` version strings, so `@ionic/core` still lists the same `@stencil/core` dep it had under the old dev pin. Probably fine since that pin was already a `major-9.0` build, but might be worth a real `npm install` in the four template dirs at some point, since these get copied straight into the generated StackBlitz project.

ShaneK (changes requested):
This looks great, the sweep itself is complete as far as I can tell, but need a couple of changes.

One thing that isn't from this PR but will cause us grief once v9 is live: `vercel.json` still redirects `/docs/v8/:match*` over to `/docs/:match*`, plus the `ja` equivalent. On this PR's preview there are 326 `/docs/v8/` URLs in the sitemap and every one I tried 308s to the v9 page instead. I think we want to drop both lines and add `/docs/v9/:match*` versions so links against the new current version still land somewhere.

versioned_docs/version-v8/reference/support.md
brandyscarney: Oh yeah I noticed the year was wrong after I copied it apparently. 😂

static/code/stackblitz/v9/html/package-lock.json
brandyscarney: Yeah we will want to test out the StackBlitz before we merge `major-9.0`.

ShaneK (approved):

PR #4643, opened 2026-08-17
docs(react): document typed overlay hook props and typescript 5.4 requirement

ShaneK:
Currently, nothing documents that `useIonModal` and `useIonPopover` type `componentProps` against the component in v9, and the React support table still says TypeScript 3.7+.

With this change, the v9 migration guide covers both, `docs/react/overlays.md` explains where prop types are inferred from, and the support table moves to 5.4+ with a footnote. The two React playgrounds that passed props the new typing rejects are fixed.

Note: this depends on [the framework change](https://github.com/ionic-team/ionic-framework/pull/31362) landing first, so the guide describes type errors that don't happen yet.

### DIFF
--- a/docs/reference/support.md
+++ b/docs/reference/support.md
@@ -56,7 +56,7 @@ The Ionic team has compiled a set of recommendations for using the Ionic Framewo
 [^1]: Angular 14.x supported starting in Ionic v6.1.9. Angular 15.x supported starting in Ionic v6.3.6.
 [^2]: Angular 17.x supported starting in Ionic v7.5.4.
 [^3]: Angular 18.x supported starting in Ionic v8.2.0.
-[^4]: Ionic v9 supports TypeScript 5.4+ for compatibility with Angular 18. Using Angular 21 requires TypeScript 5.9 or later, and Angular 22 requires TypeScript 6.0 or later, per Angular's own requirements.
+[^4]: `@ionic/angular` and `@ionic/react` require TypeScript 5.4+. Using Angular 21 requires TypeScript 5.9 or later, and Angular 22 requires TypeScript 6.0 or later, per Angular's own requirements.
 
 **Angular 13+ Support On Older Versions of iOS**
 
@@ -68,14 +68,12 @@ Note that later versions of Ionic do not support iOS 13; refer to the [mobile su
 
 | Framework | Required React Version | TypeScript |
 | :-------: | :--------------------: | :--------: |
-|    v9     |          v18+          |  5.4+[^5]  |
+|    v9     |          v18+          |  5.4+[^4]  |
 |    v8     |          v17+          |    3.7+    |
 |    v7     |          v17+          |    3.7+    |
 |    v6     |          v17+          |    3.7+    |
 |    v5     |         v16.8+         |    3.7+    |
 |    v4     |         v16.8+         |    3.7+    |
-
-[^5]: Ionic React v9 requires TypeScript 5.4+. `useIonModal` and `useIonPopover` type `componentProps` against the component they are given, and their published type definitions use `NoInfer`, which TypeScript added in 5.4. This matches the minimum Ionic Angular v9 requires.
 
 #### Ionic Vue
 
--- a/docs/reference/support.md
+++ b/docs/reference/support.md
@@ -68,12 +68,14 @@ Note that later versions of Ionic do not support iOS 13; refer to the [mobile su
 
 | Framework | Required React Version | TypeScript |
 | :-------: | :--------------------: | :--------: |
-|    v9     |          v18+          |    3.7+    |
+|    v9     |          v18+          |  5.4+[^5]  |
 |    v8     |          v17+          |    3.7+    |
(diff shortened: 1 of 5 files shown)


docs/updating/9-0.md
```diff
@@ -180,6 +181,42 @@ npm install react@latest react-dom@latest
 npm install @ionic/react@latest @ionic/react-router@latest
 ```
 
+#### TypeScript {#react-typescript}
```
thetaPC: We should be using the new heading IDs [syntax](https://docusaurus.io/blog/releases/3.10#strict-heading-ids):


```suggestion
#### TypeScript {/* react-typescript */}
```

docs/updating/9-0.md
```diff
@@ -180,6 +181,42 @@ npm install react@latest react-dom@latest
...
+This surfaces existing mistakes at build time rather than at runtime, so an app that was passing incorrect props will see new type errors. Fix the call sites to match the component. Components typed as `FC<any>` stay permissive, so an app whose overlay components are untyped needs no changes. To pin the props type instead of relying on inference, pass it as a type argument: `useIonModal<Props>(Component, props)`.
+
+The prop types are inferred from the component, not from `componentProps`. A component defined inline needs its props annotated:
+
+```diff
+- const [present, dismiss] = useIonModal(({ name }) => <div>Hello {name}.</div>, { name: 'Dave' });
++ const [present, dismiss] = useIonModal(({ name }: { name: string }) => <div>Hello {name}.</div>, { name: 'Dave' });
+```
+
+Running `npx @ionic/migrate --experimental` can annotate the inline case for you. The remaining call sites are reported rather than rewritten, since the correct fix depends on what the component is meant to accept.
```
thetaPC: I wonder if we should have a link pointing to the "Automated Migration". Some devs might have forgot about the migration tool or completely skipped that section. It could be nice to have the link here so they can quickly read about it.

thetaPC (approved):
LGTM, non-blocker comments

docs/updating/9-0.md
ShaneK: Done! [e5c7856](https://github.com/ionic-team/ionic-docs/pull/4643/commits/e5c7856a6e06173f5b4026b37285c7f2d6678af1)

docs/updating/9-0.md
ShaneK: Good call, done! [e5c7856](https://github.com/ionic-team/ionic-docs/pull/4643/commits/e5c7856a6e06173f5b4026b37285c7f2d6678af1)

docs/reference/support.md
```diff
@@ -68,13 +68,15 @@ Note that later versions of Ionic do not support iOS 13; refer to the [mobile su
...
 | :-------: | :--------------------: | :--------: |
-|    v9     |          v18+          |    3.7+    |
+|    v9     |          v18+          |  5.4+[^5]  |
 |    v8     |          v17+          |    3.7+    |
 |    v7     |          v17+          |    3.7+    |
 |    v6     |          v17+          |    3.7+    |
 |    v5     |         v16.8+         |    3.7+    |
 |    v4     |         v16.8+         |    3.7+    |
 
+[^5]: Ionic React v9 requires TypeScript 5.4+. `useIonModal` and `useIonPopover` type `componentProps` against the component they are given, and their published type definitions use `NoInfer`, which TypeScript added in 5.4. This matches the minimum Ionic Angular v9 requires.
```
brandyscarney: Optional: Couldn't we combine this with the 1st footnote?

Image

Or can we simplify it and write it the same way?

```suggestion
[^5]: Ionic v9 requires TypeScript 5.4+, as its published type definitions use `NoInfer`, which was introduced in TypeScript 5.4.
```

docs/updating/9-0.md
```diff
@@ -180,6 +181,42 @@ npm install react@latest react-dom@latest
 npm install @ionic/react@latest @ionic/react-router@latest
 ```
 
+#### TypeScript {/* #react-typescript */}
+
+The `@ionic/react` package requires TypeScript 5.4 or later. Its type definitions use `NoInfer`, which TypeScript added in 5.4. This matches the minimum `@ionic/angular` v9 requires.
```
brandyscarney:
```suggestion
The `@ionic/react` package requires TypeScript 5.4 or later. Its type definitions use `NoInfer`, which TypeScript added in 5.4.
```

I don't think we need to mention Angular

brandyscarney (approved):
Looks good, just some optional changes!

docs/updating/9-0.md
ShaneK: Done: [aac0f51](https://github.com/ionic-team/ionic-docs/pull/4643/commits/aac0f518b8aba8633437f9cc0899757b502c5978)

docs/reference/support.md
ShaneK: Done: [aac0f51](https://github.com/ionic-team/ionic-docs/pull/4643/commits/aac0f518b8aba8633437f9cc0899757b502c5978)

PR #4657, opened 2026-08-20
chore(ci): check build memory against vercel machine size

thetaPC:
Issue URL: N/A

No workflow builds the site. CI.yml lints, runs tests, spellchecks, and checks translation keys, but never runs `npm run build`.

Vercel preview deployments do build, but `build:preview` is `docusaurus build --locale en`, so the Japanese locale is only ever built by production. That means nothing before a merge exercises the full build, and nothing measures how much memory it needs. There is no signal ahead of a release about whether the build still fits the Vercel build machine.

## What is the new behavior?

Adds a manually triggered workflow that measures the full production build and compares it against the Vercel build machine the project is pinned to.

It builds all locales on Ubuntu, records peak memory with `/usr/bin/time -v`, and emits exactly one annotation:

| Outcome | Annotation |
|---|---|
| Fits | notice, "Success! No action needed" |
| Too close to the ceiling | error, "Move to {tier} before releasing" |
| Elastic selected | warning, "Check the machine Elastic will assign" |
| Build did not finish | warning, with the peak it reached before stopping |

The machine tier is a dropdown input rather than a constant, and the capacities live in a single `TIERS` string that both the recommendation and the capacity lookup read, so correcting or adding a tier is one edit. Elastic is reported rather than checked, because it sizes each build from recent successful builds and has no fixed capacity to compare against.

The measurement log is uploaded as an artifact and the report step runs even when the build fails. GNU time writes its report before the command's exit status is known, so a build killed for running out of memory still leaves its peak in the log, which is the case where the number matters most.

Node is pinned to 24 to match production. `engines.node` is `>=20.0.0`, which overrides the Project Settings version and resolves to the latest 24.x per [Vercel's mapping table](https://vercel.com/docs/functions/runtimes/node-js/node-js-versions).

Run it before a release. Adding a documentation version is what moves the number, so that is when the answer matters.

- [ ] Yes
- [x] No

## Other information

**Why this exists.** The v9 release pushed the production build past the 8 GB ceiling of Vercel's Standard machine, and three consecutive production deploys were killed partway through the Japanese locale. Preview builds stayed green throughout, because they only build English, which is the gap this workflow closes.

### DIFF
--- a/.github/workflows/build-memory.yml
+++ b/.github/workflows/build-memory.yml
@@ -37,15 +37,23 @@ jobs:
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
-      - name: ⚙️ Use Node.js 20
+      # Must match the major Vercel builds on, or the measurement is taken on a
+      # different engine than production. `engines.node` in package.json is
+      # ">=20.0.0", which overrides the Project Settings version and resolves to
+      # the latest 24.x per Vercel's mapping table:
+      # https://vercel.com/docs/functions/runtimes/node-js/node-js-versions
+      - name: ⚙️ Use Node.js 24
         uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
-          node-version: 20
+          node-version: 24
 
       - name: 🕸️ Install Dependencies
         run: npm ci --legacy-peer-deps
 
       - name: 🏗️ Build All Locales
+        # Named so the report step can read whether the build actually
+        # finished, since it runs even when this one fails.
+        id: build
         env:
           # `npm run build` resolves to build:${VERCEL_ENV:-preview}, and
           # build:preview is English only. Without this the job would measure
@@ -61,9 +69,23 @@ jobs:
         run: /usr/bin/time -v -o build-memory.log npm run build
         shell: bash
 
+      # Kept even when the build fails. GNU time writes its report before the
+      # command's exit status is known, so a build killed for running out of
+      # memory still leaves its peak in the log, which is the case where the
+      # number matters most.
+      - name: 📤 Upload Measurement Log
+        if: always() && steps.build.outcome != 'skipped'
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: build-memory-log
+          path: build-memory.log
+          if-no-files-found: warn
+
       - name: 📊 Report Against Machine Capacity
+        if: always() && steps.build.outcome != 'skipped'
         env:
           MACHINE: ${{ inputs.machine }}
+          BUILD_OUTCOME: ${{ steps.build.outcome }}
         run: |
           set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build-memory.yml
```diff
@@ -0,0 +1,129 @@
...
+          #
+          # This margin is a judgment call, not a number published by Vercel.
+          # It only becomes load bearing when the build comes within a couple
+          # of GB of a tier boundary.
+          THRESHOLD_PCT=75
+
+          # Smallest fixed tier this build fits on under the threshold. Falls
+          # back to Turbo, the largest, when nothing fits.
+          RECOMMENDED=$(awk -v peak="$PEAK_GB" -v t="$THRESHOLD_PCT" 'BEGIN {
+            split("Standard:8 Enhanced:16 Turbo:60", tiers, " ")
```
ShaneK, line 92: The PR description says there's no hardcoded capacity to keep up to date, but the dropdown only carries the tier name. The capacities are literals here and again in the `case` below, and the tier count is hardcoded a third time as the `i <= 3` bound.

The values are all correct against Vercel's current specs, so nothing's wrong today. It's the next change I'd worry about: patching the `case` but not this string gives you a correct pass or fail with the wrong recommendation, and adding a fifth option to the dropdown would fall through to `Turbo` without anyone noticing. Could both halves read one shared tier table instead? Worth fixing the claim in the PR body too, since that's what someone reads first when they come to update it.

.github/workflows/build-memory.yml
```diff
@@ -0,0 +1,129 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # prebuild runs `crowdin upload` when this is non-empty, which pushes
+          # source strings to Crowdin. This job only measures memory and must
+          # never publish, so it is pinned empty rather than relying on the
+          # secret not being exposed to this workflow.
+          CROWDIN_PERSONAL_TOKEN: ''
+        run: /usr/bin/time -v -o build-memory.log npm run build
+        shell: bash
+
+      - name: 📊 Report Against Machine Capacity
```
ShaneK, line 64: This step has no `if:`, so it inherits `success()` and gets skipped whenever the build fails, and `build-memory.log` isn't uploaded anywhere so it goes away with the runner.

That's worth keeping, because the log still has the number in it. GNU time writes its whole report before working out its own exit status, so an OOM-killed build leaves the peak RSS in the log with a `Command terminated by signal 9` line in front of it, and exits 137. The `awk` match already tolerates that preamble. So `if: always()` plus an `upload-artifact` step would get you the figure in the case where it's least ambiguous.

.github/workflows/build-memory.yml
```diff
@@ -0,0 +1,129 @@
...
+    # Linux, so the numbers are comparable.
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: ⚙️ Use Node.js 20
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: 20
```
ShaneK, line 43: I think Vercel is building this on Node 24, not 20. Their docs say `engines.node` overrides whatever's picked in Project Settings, and their mapping table puts `>=20.0.0` in the 24.x row, resolving to the latest 24.x. This repo's `package.json` declares exactly that.

If that's right, the measurement is on a different major than production, and GC behavior moves between majors enough to undercut the comparison. The quick fix is `node-version: 24`. A committed `.nvmrc` that both this workflow and Vercel read would be the durable one, since there's no `.nvmrc` or Volta pin right now.

.github/workflows/build-memory.yml
```diff
@@ -0,0 +1,129 @@
...
+
+          # Smallest fixed tier this build fits on under the threshold. Falls
+          # back to Turbo, the largest, when nothing fits.
+          RECOMMENDED=$(awk -v peak="$PEAK_GB" -v t="$THRESHOLD_PCT" 'BEGIN {
+            split("Standard:8 Enhanced:16 Turbo:60", tiers, " ")
+            for (i = 1; i <= 3; i++) {
+              split(tiers[i], tier, ":")
+              if (peak / tier[2] * 100 <= t) { print tier[1]; exit }
+            }
+            print "Turbo"
```
ShaneK, line 97: Nit: when nothing fits under the threshold this prints `Turbo`, which is also the largest tier, so on Turbo you'd get "Move to Turbo before releasing", and the Elastic warning would say the build fits on Turbo when it doesn't. It needs a peak over 45 GB to get here, so it's out of reach on a 16 GB runner today. A distinct "nothing fits" value would stop the message claiming a fit that isn't there, but up to you, this is cosmetic while the runner is the ceiling.

ShaneK (commented):
Nice piece of work, and the reasoning in the comments is great

.github/workflows/build-memory.yml
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4657/changes/c4857cc8e910fb8d13b0b881e2c5be76c28a7bcb

.github/workflows/build-memory.yml
thetaPC: https://github.com/ionic-team/ionic-docs/commit/c4857cc8e910fb8d13b0b881e2c5be76c28a7bcb

.github/workflows/build-memory.yml
thetaPC: https://github.com/ionic-team/ionic-docs/commit/c4857cc8e910fb8d13b0b881e2c5be76c28a7bcb

`.nvmrc` wouldn't help though, Vercel's build only reads Project Settings and `engines.node`, so `node-version-file: package.json` is the durable version if we want it.

Aligning the repo's Node version with Vercel's is on my todo. It's waiting on the dependency updates already in flight, since a lot of packages here are out of date.

.github/workflows/build-memory.yml
thetaPC: https://github.com/ionic-team/ionic-docs/commit/c4857cc8e910fb8d13b0b881e2c5be76c28a7bcb

PR #4660, opened 2026-08-24
chore(archive): add missing lint ignores and document archiving

thetaPC:
Issue URL: N/A

v7 was archived in #4647, but it was never added to `.prettierignore` or `cspell.json`, so Prettier and cspell still walk archived v7 content. The `static/usage/v6` and `v7` playground folders were never ignored either.

Separately, the framework landing redirects in `vercel.json` cover v6 and v7 but not v8. No version ships an index page for those routes, so `/docs/v8/angular`, `/docs/v8/react`, and `/docs/v8/vue` currently fail to redirect to their overview pages, and `/docs/v8/javascript` has no versioned redirect at all.

The archiving instructions in `CONTRIBUTING.md` only covered `versions.json` and `versionsArchived.json`, which is how the v7 lint ignores were missed in the first place.

## What is the new behavior?

Config catch up:

- `.prettierignore`: adds `versioned_docs/version-v7`, `static/usage/v6`, and `static/usage/v7`
- `cspell.json`: adds `versioned_docs/version-v7`
- `vercel.json`: adds v8 to the `angular`, `react`, and `vue` landing redirects, and adds a v8 scoped `javascript` redirect

The archiving steps in `CONTRIBUTING.md` now also cover `vercel.json`, `.prettierignore`, `cspell.json`, and `renovate.json`, so the next archive has a complete checklist.

- [ ] Yes
- [x] No

## Other information

**Why the `vercel.json` step is ordered first.** An archived version is served from a frozen Vercel deployment that ships its own copy of `vercel.json`. A version scoped redirect therefore has to exist *before* the archive build, or it never reaches the archive and those links 404 there permanently. This is not hypothetical: v6 and v7 both work today only because the redirect pattern predates their most recent promoted builds.

**The v8 redirects fix a live issue.** They are not only preparation for archiving v8. Those routes are broken on `main` right now, and archiving would have frozen the breakage in place.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -221,10 +221,25 @@ Archived versions are served from a frozen Vercel deployment instead of being re
 
 The archived URL has to point at a build that _included_ the version, so you build it first, then move it to `versionsArchived.json`:
 
-1. **Build the version.** Make sure it is in `versions.json`. If you are refreshing an already-archived version, move it out of `versionsArchived.json` and back into `versions.json`. Commit, push and let Vercel deploy.
-2. **Promote the deployment.** In the Vercel dashboard, open that deployment and **Promote to Production** so it does not get cleaned up. Wait for the build to finish before pushing again, or it may get canceled.
-3. **Copy its URL.** Use the deployment's unique `ionic-docs-<hash>-ionic1.vercel.app` URL, not the branch or production alias.
-4. **Archive it.** Remove the version from `versions.json`, then add it to `versionsArchived.json` with `/docs/<version>` appended and no trailing slash (a trailing slash causes a brief 404 flash):
+1. **Add the version to the `vercel.json` redirects.** The frozen deployment serves its own copy of [`vercel.json`](./vercel.json), so a version scoped redirect has to be in place _before_ the build in the next step, or it will be missing from the archive for good. Add the version to each `:version(...)` group that applies to it. The `angular`, `react`, `vue` and `javascript` landing pages all need this, since no version ships an index page for them:
+
+   ```json
+   {
+     "source": "/docs/:version(v6|v7|v8|v<version>)/angular",
+     "destination": "/docs/:version/angular/overview"
+   },
+   {
+     "source": "/docs/:version(v8|v<version>)/javascript",
+     "destination": "/docs/:version/javascript/overview"
+   }
+   ```
+
+   The `javascript` group only goes back to v8, since v6 and v7 have no `javascript/` section.
+
+2. **Build the version.** Make sure it is in `versions.json`. If you are refreshing an already-archived version, move it out of `versionsArchived.json` and back into `versions.json`. Commit, push and let Vercel deploy.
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
brandyscarney: Can we add a note that this only includes older versions not the current version? Since v9 is the current version it should be excluded here.

CONTRIBUTING.md
```diff
@@ -221,18 +221,54 @@ Archived versions are served from a frozen Vercel deployment instead of being re
...
-2. **Promote the deployment.** In the Vercel dashboard, open that deployment and **Promote to Production** so it does not get cleaned up. Wait for the build to finish before pushing again, or it may get canceled.
-3. **Copy its URL.** Use the deployment's unique `ionic-docs-<hash>-ionic1.vercel.app` URL, not the branch or production alias.
-4. **Archive it.** Remove the version from `versions.json`, then add it to `versionsArchived.json` with `/docs/<version>` appended and no trailing slash (a trailing slash causes a brief 404 flash):
+1. **Add the version to the `vercel.json` redirects.** The frozen deployment serves its own copy of [`vercel.json`](./vercel.json), so a version scoped redirect has to be in place _before_ the build in the next step, or it will be missing from the archive for good. Add the version to each `:version(...)` group that applies to it. The `angular`, `react`, `vue` and `javascript` landing pages all need this, since no version ships an index page for them:
```
brandyscarney: These steps are a bit confused. The steps before this change are for archiving a version (`v7`) but in this example you are wanting to add `v8` to the list of redirects.

It might be helpful to add example versions to these steps so all the versions are added in the right places. For example, if we are adding a new current version (`v9`) we would need to add the previous current version (`v8`) to the redirects below, but the steps for building are related to the newly archived version (`v7`) that points to a Vercel link at the end of these steps.

brandyscarney (changes requested):
Requested some clarification related to what versions are being added where!

CONTRIBUTING.md
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4660/changes/1c37b51d7c0897d827d26495d767c3d4b1c6877a

CONTRIBUTING.md
thetaPC: I reword it to be neutral, hoping that helps. Let me know! https://github.com/ionic-team/ionic-docs/pull/4660/changes/9c134376c423702c7ba2a12b1640e75ec2fcfb04

brandyscarney: Here is what I was thinking: https://gist.github.com/brandyscarney/e7172fa6a18d21c2249ac4210f9b1a32

thetaPC: https://github.com/ionic-team/ionic-docs/pull/4660/changes/7e34b4a482b8a45c923177a4b6c13b4fefdf7ba8

brandyscarney (approved):
Looks good, thank you for doing this!

PR #4670, opened 2026-08-27
chore(docs): correct ignore lists and drop archived playground versions

thetaPC:
Issue URL: N/A

`.prettierignore` covers `docs/cli/commands` but not the versioned copies, so `npm run lint` reformats 46 generated files locally after any build.

`docs/api` is in `.prettierignore`, but those 94 files are tracked and hand-maintained, not generated. They've drifted from `versioned_docs/version-v8/api`, which is already prettier-formatted.

The playground generator still offers v6 and v7, which are archived and no longer built. The archiving checklist in CONTRIBUTING doesn't mention updating it.

## What is the new behavior?

- Added `versioned_docs/version-v*/cli/commands` to `.prettierignore`
- Removed `docs/api` and `versioned_docs/version-v*/api` from `.prettierignore`, formatting 93 files
- Playground generator now offers only `8` and `9`
- Added a step to the CONTRIBUTING archiving checklist for the generator

- [ ] Yes
- [x] No

## Other information

**The api reformat closes a gap rather than opening one.** `versioned_docs/version-v8/api` was already formatted this way. After this, the two trees differ only by `v8` → `v9` import paths (87 files) and 7 genuine v9 content changes.

**The cli/commands gap never affected CI.** Those files are git-ignored and CI never runs `generate-markdown`, so it only showed up locally.


.prettierignore
```diff
@@ -12,11 +12,11 @@ src/theme/prism-include-languages.ts
 legacy-stencil-components
 scripts/bak
 
-docs/api
-versioned_docs/version-v*/api
+# Auto-generated files
```
brandyscarney, line 15: Shouldn't this also be updating the v8 api docs?

docs/api/input-otp.md
```diff
@@ -99,16 +106,18 @@ import States from '@site/static/usage/v9/input-otp/states/index.mdx';
...
 
 The component will prevent users from entering any characters that don't match the specified pattern. Developers can override these defaults by providing their own pattern string to match specific input requirements.
 
 :::tip
 When using a custom `pattern`, remember that the `type` property controls which keyboard appears on mobile devices:
+
 - Use `type="number"` for numeric-only patterns to show the numeric keyboard
 - Use `type="text"` for patterns that include letters to show the alphanumeric keyboard
-:::
+  :::
```
brandyscarney, line 120: Why is this indented?

.prettierignore
thetaPC: Because v8's api docs were only ignored very recently. `versioned_docs/version-v*/api` was added to `.prettierignore` on August 21, 2026, and before that they were formatted on every `npm run lint`. So they still have the Prettier format so you won't be seeing any diffs for those files.

docs/api/input-otp.md
thetaPC: Prettier did that. You can see that it was also applied in the v8 page.

brandyscarney: Isn't this wrong though? The beginning of it is not indented.

thetaPC: I think it might be how the admonition is being set up. I can look into it on another PR especially if it leads to a lot of file changes. If you prefer it in this PR, let me know.

brandyscarney (approved):

PR #4671, opened 2026-08-27
chore(scripts): stop generating content for archived versions

thetaPC:
Issue URL: N/A

`scripts/native.mjs` and `scripts/cli.mjs` write generated content to v6 and v7 on every build. Both are archived and served from frozen Vercel deployments, so that's 140 files per build that nothing reads.

The archiving checklist in CONTRIBUTING doesn't mention the scripts, so the lists go stale on each release.

## What is the new behavior?

- Both scripts now write only to `docs/` (current) and `versioned_docs/version-v8` (the one older version still in `versions.json`)
- Added step 9 to the CONTRIBUTING archiving checklist covering both scripts

- [ ] Yes
- [x] No

## Other information

The generated files are git-ignored, so nothing was committed and nothing is removed from the repo. The frozen deployments already contain their own baked copies.

Verified by backdating the v6 and v7 output, then re-running both scripts: those directories stayed untouched while `docs/` and `version-v8` were rewritten.


CONTRIBUTING.md
brandyscarney: Should we also mention to add the archiving version back to scripts if it was removed? These steps kind of cover both archiving an existing version and an already removed version.

brandyscarney (approved):
Looks good, just noted one place we could add a note.

CONTRIBUTING.md
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4671/changes/ad6d7a8bda8cdd4eddd4bb0edd8b0cfed7685a99

PR #4672, opened 2026-08-27
docs: note a community live-update option on distribute guides

TorichanCapgo:
## Why this matters for Ionic customers

Readers finishing the Photo Gallery tutorial land on the Distribute guides, which correctly walk through Appflow end-to-end. Some teams then need a community live-update plugin (for example if they are already on a community stack, or planning what comes next for OTA). Today those guides do not point anywhere for that path.

This PR is meant to leave a clear, minimal breadcrumb for **your** readers, without changing how Appflow is presented.

## What stays the same

- The full Appflow walkthrough (repo connect, SDK install, live update, package, automations, environments, native configs)
- All Appflow CTAs, pricing plan mentions, and Quickstart links
- Appflow as Ionic’s official live-update / Mobile DevOps recommendation

## What changes

One sentence after **Deploy a Live Update** on Angular / React / Vue (current + v8), naming Capgo (`@capgo/capacitor-updater`) as a community option.

## Ask

Happy to trim further (for example drop the Capgo name and only say “community live-update plugins exist”) if that is easier for maintainers to ship.

### DIFF
--- a/docs/angular/your-first-app/8-distribute.md
+++ b/docs/angular/your-first-app/8-distribute.md
@@ -66,6 +66,8 @@ Assuming the app is configured correctly to listen to the channel you deployed t
 
 To dive into more details on the steps to deploy a live update, as well as additional information such as disabling deploy for development, check out the [Deploy a Live Update](https://ionic.io/docs/appflow/quickstart/deploy) section inside the Appflow docs.
 
+Appflow is Ionic's official live-update path for this tutorial. Teams that need a community live-update plugin can also use [Capgo](https://github.com/Cap-go/capacitor-updater) (`@capgo/capacitor-updater`).
+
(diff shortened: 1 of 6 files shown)


Benziza:
I think y should add capgo to cspell-wordlist.txt, because the spell check currently fails in all six modified files. i think y should clarify that new Appflow sales have been discontinued and that existing customers only retain access through December 31 2027

ShaneK:
Thank you for the pull request, but I'm going to close this because we don't want the docs to become a place to advertise community services. Capgo is not the only Appflow competitor that has risen up in the past few years and we don't want to spur up PR wars trying to get them all included in the docs.

Going forward, we'll likely even be removing the mentions of Appflow and live updates, given the discontinuation of the service.

PR #4673, opened 2026-08-27
chore(renovate): update syntax and ignore specific versions

brandyscarney:
This reconfigures `renovate.json` to update the following:
- Use the latest cron syntax for scheduling
- Use the latest syntax for matching package names
- Restrict the version of Angular to less than 21 for v8 and less than 23 for v9
- Restrict the version of React Router to v5 for Ionic 8 and v6 for Ionic 9

See the comment by renovate for the expected results of the reconfigure.


thetaPC (approved):
LGTM

PR #4688, opened 2026-09-01
chore(typescript): scope typechecking to src and add it to CI

thetaPC:
Issue URL: internal

`tsconfig.json` is three lines:

```json
{
  "extends": "@docusaurus/tsconfig",
  "exclude": ["static/code/stackblitz/"]
}
```

Specifying `exclude` replaces TypeScript's defaults, and `node_modules` is one of them. So TypeScript walks `node_modules` and `build/`, pulling in **995 project files** and 2,039 in total. Nothing runs `tsc`, so none of the **126 errors** it reports have ever surfaced.

**97 of those 126 are inside `build/`**, which means the error count depends on whether you happen to have run a build. That scan is also load-bearing by accident: the real `@theme/*` types arrive only because `node_modules` is being walked, not because anything asks for them.

## What is the new behavior?

`npm run typecheck` runs `tsc --noEmit` over **61 files at zero errors**, and CI runs it.

Config:

- `include` scopes to `src` plus `index.d.ts`, the only declaration for `*.module.scss`. This also restores the default excludes, which is the actual fix for the above.
- `types` names `@docusaurus/module-type-aliases` and `@docusaurus/theme-classic` deliberately. Without it the catch-all `declare module '@theme/*'` in `index.d.ts` wins and every swizzled component becomes `any`.
- `baseUrl` is redeclared. Docusaurus sets it, but `extends` resolves it relative to the declaring file, so `@site/*` pointed inside `node_modules/@docusaurus/tsconfig`.
- `allowArbitraryExtensions` lets `release-notes.d.json.ts` type the generated `release-notes.json`, so typechecking needs neither a build nor a GitHub token.

Code:

- Declared `ion-icon`, `docs-card` and `docs-cards` in `src/declarations.d.ts`
- Added `DocsFrontMatter` for our `demoUrl` and `demoSourceUrl` front matter, intersected with Docusaurus's `DocFrontMatter`
- Removed a dead `@stencil/core` import left over from a Stencil to React port
- Declared `@docusaurus/types`, which `@docusaurus/module-type-aliases` imports from in 12 places and which resolved only because npm hoists it out of `@docusaurus/core`

- [ ] Yes
- [x] No

## Other information

**`strict` is not enabled.** It surfaces 144 errors concentrated in four files, with `Playground/index.tsx` alone at 60. That belongs in its own PR, and it is worth doing before Docusaurus v4, which requires TypeScript 6, where `strict` defaults to on.

**TypeScript stays at 5.9.3**, the newest 5.x. 6 and 7 are [blocked](https://github.com/facebook/docusaurus/issues/11893) by `@docusaurus/tsconfig` setting `baseUrl`, which 6 deprecates and 7 removes outright. Docusaurus drops it in v4 (facebook/docusaurus#11915). The `ignoreDeprecations: "6.0"` escape hatch works, but it silences the whole deprecation class permanently, so it is not worth taking.

**Ejected theme components are checked rather than skipped.** The three errors in `DocItem/Layout` came from our own front matter fields, so the type lives beside it in `frontMatter.interface.ts` instead of being invented inline. The change to the upstream copy is 3 lines, all inside existing `// CUSTOM CODE` markers. The other 8 ejected files needed nothing once `types` was set.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -29,6 +29,8 @@ jobs:
       run: npm run lint
     - name: Test
       run: npm run test
+    - name: 🧩 Typecheck
+      run: npm run typecheck
     - name: 🔤 Spell Check
       run: npm run spellcheck
     - uses: ./.github/workflows/actions/check-translations
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,6 +35,7 @@
       "devDependencies": {
         "@docusaurus/module-type-aliases": "^3.10.2",
         "@docusaurus/tsconfig": "^3.10.2",
+        "@docusaurus/types": "^3.10.2",
         "@ionic/prettier-config": "^4.0.0",
         "@types/react": "^19.2.18",
         "cspell": "^10.0.1",
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "prettier": "prettier \"./**/*.{html,ts,tsx,js,jsx,md,mdx}\" --cache",
     "start": "docusaurus start",
     "test": "vitest run",
+    "typecheck": "tsc --noEmit",
     "swizzle": "docusaurus swizzle",
     "spellcheck": "cspell --no-progress --gitignore \"**/*.{md,mdx}\""
   },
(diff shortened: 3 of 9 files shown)


brandyscarney (approved):
LGTM

PR #4690, opened 2026-09-01
chore(typescript): enable strict mode

thetaPC:
Issue URL: internal

`strict` is off. Turning it on reports **144 errors** across 21 files, so much of `src/` is implicitly `any` and unchecked.

Docusaurus v4 requires TypeScript 6, where `strict` defaults to on, so this work is on the critical path for that upgrade rather than optional.

## What is the new behavior?

`"strict": true`, at zero errors.

The config change is one line. The other 25 files are the 144 errors it surfaced, and they collapse into a few causes:

- **Untyped props.** Around 20 components took `props` with no annotation. Most became `ComponentProps<'div'>` or similar.
- **Types that were wrong at the source**, where one fix cleared a cluster. `generateColor` declared it returned `ColorVariable`, whose fields are all optional, while it populates every one. Three iframe helpers took `HTMLIFrameElement` despite each already having an `if (frame)` guard inside. Two `useState([])` calls inferred `never[]`. `MdxContent` was typed `() => {}`, which is not renderable.
- **Nullable refs and DOM lookups**, mostly in the playground, handled by hoisting the value and guarding once rather than chaining at each use.

- [ ] Yes
- [x] No

## Other information

**Real bugs surfaced**, all previously invisible because the surrounding code was `any`:

- `aria-hidden={... : null}` in four places. React omits an attribute for `undefined`; `null` is not an accepted value.
- A ref callback returning a value: `ref={(tabControl) => tabRefs.push(tabControl)}`. React 19 reads a returned value as a cleanup function.

### DIFF
--- a/src/components/global/BestPracticeFigure/index.tsx
+++ b/src/components/global/BestPracticeFigure/index.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { type ReactNode } from 'react';
 
 import './best-practice-figure.css';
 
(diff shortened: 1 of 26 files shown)


brandyscarney (approved):
Looks good!

PR #4694, opened 2026-09-02
docs(angular): correct OnPush guidance in lifecycle note

ShaneK:
## What's currently wrong

The lifecycle page has said for years that a component using `ion-nav` or `ion-router-outlet` shouldn't use `OnPush`, and that setting it will stop `ngOnInit` from firing. That's wrong on a couple of counts. The `ngOnInit` part isn't true in either change detection mode, and `ion-nav` was never affected either, because its pages get attached as root views instead of as children of the component hosting the nav. The page also never explained the why, which is what #1758 was asking for back in 2020.

## What's changing and why

When Angular runs change detection it walks down from the root of your app, and a clean `OnPush` component stops the walk there and skips everything underneath it. So if any component above your `ion-router-outlet` is `OnPush`, the routed page inside it never gets checked, and a plain field you set after an `await` won't render. On Angular 18 through 21 that only happens if you set `OnPush` yourself. Angular 22 makes `OnPush` the default for any component that doesn't declare a strategy, so it can now happen without you touching anything.

We fixed our side of it in ionic-team/ionic-framework#31414, where every Ionic component declares a strategy explicitly so `ion-router-outlet` and `ion-tabs` stay eager. This PR helps users understand why if they happen to be using zone.js, Angular 22, and plain fields. The lifecycle note now says which components have to stay eager and what your options are if you'd rather not do that. The zoneless guide gets a "Change detection on Angular 22" section with an `app-root` example.


docs/angular/lifecycle.mdx
```diff
@@ -28,7 +28,9 @@ For more info on the Angular Component Life Cycle events, visit their [component
 
 :::note
 
-Components that use `ion-nav` or `ion-router-outlet` should not use the `OnPush` change detection strategy. Doing so will prevent lifecycle hooks such as `ngOnInit` from firing. Additionally, asynchronous state changes may not render properly.
+If your pages keep state in plain fields rather than signals, the component hosting `ion-router-outlet` or `ion-tabs` needs eager change detection, as does every component between it and your application root. A change detection pass starts at the application root and skips a clean `OnPush` view along with everything below it, so an `OnPush` component above the outlet stops updates from reaching the routed pages under it. The pages themselves can use `OnPush`, as long as their state is a signal or they call `markForCheck()`.
+
+On Angular 18 through 21 this only affects you if you set `OnPush` on those components yourself, because a component that does not declare a strategy is eager. Angular 22 makes `OnPush` the default for components that do not declare one, so refer to [Change detection on Angular 22](/docs/angular/zoneless.mdx#change-detection-on-angular-22) for what your app shell has to declare.
```
thetaPC:
```suggestion
On **Angular 18 through 21** this only affects you if you set `OnPush` on those components yourself, because a component that does not declare a strategy is eager.

**Angular 22** makes `OnPush` the default for components that do not declare one, so refer to [Change detection on Angular 22](/docs/angular/zoneless.mdx#change-detection-on-angular-22) for what your app shell has to declare.
```

thetaPC (approved):
LGTM, minor suggestion to make it easier for devs to quickly read the info they need based on their version.

PR #4696, opened 2026-09-02
chore(deps): replace the deprecated crowdin package with @crowdin/cli

thetaPC:
Issue URL: internal

`crowdin@3.5.0` is deprecated, and its dependency tree is the source of the repo's only two outstanding advisories, `form-data` (critical) and `got` (moderate).

## What is the new behavior?

`@crowdin/cli@5.0.1`, which is maintained and is the package [Docusaurus's i18n guide](https://docusaurus.io/docs/i18n/crowdin) prescribes. Both advisories go with the old tree, and the lockfile drops **333 lines**: 50 transitive packages out, 8 in.

It's also the package our setup was always written for. `crowdin:sync` calls `crowdin upload && crowdin download`, but the old package is an API client with no `bin`, so those were commands that didn't exist.

- [ ] Yes
- [x] No

`prebuild` only calls `crowdin:sync` when `CROWDIN_PERSONAL_TOKEN` is set, and it isn't set in Vercel. The sync didn't run before and doesn't now.

## Other information

**Whether Crowdin is still in use is unresolved.** `crowdin.yml` hasn't been touched since April 2023, while Japanese prose arrives through `scripts/i18n.sh` from the actively maintained `translation/jp` branch. Until we know which it is, being on a maintained package beats sitting on a deprecated one with a critical advisory. Removing it outright is the other option, but that decision needs the answer first.

Verified without credentials: the CLI parses `crowdin.yml` and resolves `api_token_env`; with a placeholder token it reaches Crowdin and stops at authorization, uploading nothing. Plain `npm ci` works and the binary survives it.

One thing to note for whoever picks up the ownership question. Two things used to prevent the sync running, no binary and no token. Now only the token does, and if one is added, `crowdin download` runs after `i18n.sh` and writes into the same path the `translation/jp` tarball unpacks into.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,7 @@
       "name": "ionic-docs",
       "version": "0.0.0",
       "dependencies": {
+        "@crowdin/cli": "^5.0.1",
         "@docusaurus/core": "^3.10.2",
         "@docusaurus/faster": "^3.10.2",
         "@docusaurus/preset-classic": "^3.10.2",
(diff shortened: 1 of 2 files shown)


brandyscarney (approved):
LGTM!

PR #4697, opened 2026-09-02
chore(renovate): manage the root package.json and scope every rule

thetaPC:
Issue URL: internal

Renovate's `includePaths` covers only `static/code/stackblitz/**/*/package.json` and `.github/workflows/**`, so the root `package.json` gets no dependency PRs at all. Every upgrade there is manual, which is how `typescript` came to read `^5.9.3` while 5.3.3 was actually installed.

## What is the new behavior?

The root `package.json` is in scope.

Three rules were added for it, all at the end of `packageRules`:

- **`docusaurus`** groups the six `@docusaurus/*` packages, which sit at `^3.10.2` and must move together
- **`react-root`** caps react and react-dom below 20. `@docusaurus/core` declares `"react": "^18.0.0 || ^19.0.0"`
- **`typescript`** stays below 6, since `@docusaurus/tsconfig` sets `baseUrl`, which 6 deprecates and 7 removes

Each names the Docusaurus constraint behind it, so they can be revisited together at v4.

**Every rule now carries `matchFileNames`.** Four didn't, and they were all written when stackblitz was the only thing managed. Left alone, `@ionic/**` and `react` would have quietly started applying to the root: the first carries `"minimumReleaseAge": "0 days"` so Ionic releases reach the examples immediately, which would have let `@ionic/prettier-config` skip the repo's 3-day soak and land grouped with example bumps.

That's the reason for the blanket rule rather than only fixing the two that bite. An unscoped rule widens on its own whenever a path is added to `includePaths`.

- [ ] Yes
- [x] No

Config only. Validated with `renovate-config-validator`.

## Other information

The `docusaurus` group is strictly redundant: `config:recommended` extends `group:monorepos`, and Docusaurus is in Renovate's monorepo data, so those packages would group anyway. Kept deliberately, because the rule documents the constraint rather than leaving it implicit in data we don't control.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -11,6 +11,7 @@
   "semanticCommits": "enabled",
   "rebaseWhen": "never",
   "includePaths": [
+    "package.json",
     "static/code/stackblitz/**/*/package.json", 
     ".github/workflows/**"
   ],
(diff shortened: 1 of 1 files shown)


brandyscarney (approved):
LGTM!

PR #4698, opened 2026-09-03
chore(v7): rename archived v7 pages and usage partials to .mdx

thetaPC:
Issue URL: internal

Every live doc page has moved to `.mdx` as part of the Strict MDX migration, but the archived versions were deliberately left alone, since they are never built.

That holds only while they stay archived. Most of these pages are MDX in everything but name: 185 of v7's 240 pages contain imports and 136 contain JSX. Once Strict MDX is enabled, reviving this version would parse those as plain CommonMark, turning every import into a line of literal text on the page. It would not fail the build, so the first sign of trouble would be a reader seeing an import statement in the middle of a paragraph.

Archived versions are revivable by design: the archiving guide already documents moving one back into `versions.json`. So leaving them behind converts a documented, supported operation into one that silently produces broken pages.

There is a second, more immediate problem. The CLI and Capacitor plugin pages are now generated as `.mdx`, but v7's pages still link to them as `.md`. Those links are already dead in a build where v7 is present.

## What is the new behavior?

v7 is migrated to match the live versions: 239 pages and 401 playground partials renamed, along with the references between them, 522 component API imports, 405 playground imports, and 281 internal links.

`versioned_docs/version-v7/README.md` keeps its extension. It is a contributor note excluded from the site by the `exclude` setting, which names the file by extension, so renaming it would publish it as a page.

Nothing about the archived state changes. v7 stays in `versionsArchived.json`, stays out of `versions.json`, and is still not built. This only makes it safe to revive later.

- [ ] Yes
- [x] No

## Other information

This lands before the transitional compatibility code is removed, and that order matters. Those archived pages currently import 435 component API partials by their `.md` names, which only resolve because the plugin still writes both. Removing that first would leave the archived versions broken and then fix them afterwards. Doing the versions first means nothing anywhere imports the old names by the time it comes out.

Verification was done by temporarily reviving the version, since archived pages are otherwise never built and a mistake here would have no signal at all.

### How to test

Archived versions are not built, so nothing in a normal preview exercises these pages. To check them, revive the version locally:

1. Move `v7` out of `versionsArchived.json` and back into `versions.json`.
2. Add `versioned_docs/version-v7` back to the write targets in `scripts/native.mjs` and `scripts/cli.mjs`, so the generated CLI and plugin pages exist for it.
3. Run `npm run generate-markdown`, then `npx docusaurus build` with no `--locale` flag.
4. Revert all of the above afterwards. None of it belongs in this PR.

The build should complete with no errors, and `/docs/v7/` should render. Worth spot checking an API page for its generated tables and playgrounds, and following a link from a v7 page into the CLI or plugin docs, since those are the links this PR repairs.


ShaneK (approved):
LGTM! 🚀

PR #4704, opened 2026-09-03
chore(cleanup): remove commented-out code and fix what it pointed at

thetaPC:
Issue URL: internal

Commented-out code scattered across `src/` and `scripts/`, none of it reachable.

The largest is **111 lines** in `scripts/cli.mjs`: six commented-out `render*` functions. They render Stencil component tables (properties with `attr` and `default`, events, slots), but that script generates CLI command pages, whose data has `inputs` and `options` instead. Working versions of all six already live in `plugins/docusaurus-plugin-ionic-component-api`. The script was started from that generator as a template and the inapplicable parts were commented rather than deleted.

## What is the new behavior?

**141 lines removed**, and one comment turned into the fix it was pointing at.

Removed:

- `scripts/cli.mjs` — the six render functions and a debug `console.log`
- `_code.scss` — a 15-line block labelled *"alternative solution ... with it's own drawbacks"*, and a leftover `padding`
- `demo.css` — a disabled `@media (max-width: 1160px)` rule
- `Codepen` — a debug `console.log`. Its `status` variable existed only to feed it, so the `useScript` call is now assignment-free.
- `_markdown.scss` — a leftover `letter-spacing`

`DocsButton` had `// const isInternal = /^\/docs/.test(href);`, an unfinished attempt at what `DocsCard` already does. Rather than delete the idea, the component now uses Docusaurus's `Link`, which routes internal hrefs client-side and adds `target="_blank"` plus `rel="noopener noreferrer"` to external ones. That's less code than the check would have been. It also fixes the href: `href="theming"` was resolving against the browser URL, and now resolves to `/docs/developer-resources/guides/first-app-v4/theming`.

Also fixes the developer-resources redirect, which was scoped to v5 and current only. It now covers every version, which is what "developer resources live at ionic.io now" implies.

- [ ] Yes
- [x] No

## Other information

**The `DocsButton` change can only be verified locally**, since `vercel.json` doesn't apply to `npm start`:

`/docs/developer-resources/guides/first-app-v4/creating-photo-gallery-device-storage`

The Continue button at the bottom should navigate without a full page reload. Every `href` usage of `DocsButton` sits on a developer-resources page, and the redirect fix in this PR sends all of those off-site, so there's no preview URL that exercises it. Its only other usages are ~180 archived v5 native pages that pass no `href` and take the `<button>` branch.

That's worth knowing beyond this PR: if developer-resources is gone for good, `DocsButton`'s linking path is effectively dead and the component deserves the same audit treatment as `ion-icon`.


ShaneK (approved):
LGTM 🚀

PR #4708, opened 2026-09-05
chore(i18n, v5): rename archived v5 pages to .mdx

thetaPC:
Issue URL: https://ionic-cloud.atlassian.net/browse/FW-6456

The archived v5 docs on this branch are still `.md`. Docusaurus's Strict MDX guidance asks for `.mdx`, and v5 is the last tree that has to move before the transitional `.md` shims in the component API plugin can be removed.

## What is the new behavior?

The jp companion to #4700. Same tree, same result.

- All **460** pages in `versioned_docs/version-v5` renamed. v5 has no `README.md` and no `static/usage/v5`, so the whole tree moves.
- **849** imports and links updated to match

Unlike v6 and v7, this one fixes nothing. v5's `native/` and `cli/commands` pages are committed rather than generated, so the tree was already internally consistent and none of its links pointed at anything that had moved. The build is warning for warning identical to the base branch.

`versions.json` is not in this PR, so v5 stays archived and nothing about the deployed site changes. Refer to the testing steps below to build it locally.

- [ ] Yes
- [x] No

## Other information

The v5 link and import graph matches #4700 entry for entry across all 460 files.

The generation scripts do not write to v5 and this PR does not change that. Adding v5 to them writes 24 `.mdx` over the 191 committed `native/` pages, which produces duplicate routes and silently drops the build from 460 pages to 286.

### How to test

v5 is archived, so a normal build skips it. To build it:

1. Add `"v5"` to `versions.json`: `["v8", "v5"]`
2. `npm run build`
3. Confirm **460** v5 pages build with no `couldn't be resolved` warnings. 460 is also the number of pages tracked in the tree, which is the check that the 191 `native/` and 53 `cli/commands` pages came through.
4. Revert `versions.json`

Spot check `/docs/v5/api/accordion` and `/docs/v5/native/camera`, one hand-written page and one of the committed native pages.


brandyscarney (approved):
LGTM

PR #4716, opened 2026-09-08
chore(mdx): adopt strict MDX syntax and remove the transitional shims

thetaPC:
Issue URL: internal

Two transitional shims in `plugins/docusaurus-plugin-ionic-component-api` are still carrying the `.md` to `.mdx` rename: the component API plugin emits every partial twice, and an `extensionAlias` lets any `.md` import resolve to a `.mdx` file.

Three of the four items in [Docusaurus 3.10's Strict MDX guidance](https://docusaurus.io/blog/releases/3.10#strict-mdx) are also outstanding, so `future.v4.mdx1CompatDisabledByDefault` cannot be turned on. Turning it on today fails the build outright:

```bash
Error: MDX compilation failed for file "docs/updating/9-0.mdx"
Cause: Could not parse expression with acorn
```

## What is the new behavior?

Both shims are gone. Nothing on any live branch still imports the `.md` names, so the dual write was already dead code, and the alias was load-bearing for exactly six references, now pointed at the files that actually exist.

The remaining Strict MDX items are done:

- **8** heading IDs in `docs/updating/9-0.mdx`, `{#id}` to `{/* #id */}`
- **100** HTML comments to `{/* */}`, in archived v5 and v6
- **4** admonition titles to `:::tip[Title]`, in archived v6

That clears the last blocker on the v4 flag. The flag itself is deliberately not in this PR, because `translation/jp` and `major-10.0` still need this same change first.

A further **9** commented-out blocks were deleted rather than converted. Each is a placeholder that v7 onward already dropped, and nothing else in the repo references them: an `--item-detail-push-show` TODO, four `TOC goes here`, two `TODO: image?`, and a stubbed `## Customization` section.

- [ ] Yes
- [x] No

### How to test

Archived versions are the risk here, since 45 of the 48 files are in v5 and v6, so verify them with the flag on:

2. Add `mdx1CompatDisabledByDefault: true` under `future.v4` in `docusaurus.config.js`
3. `npx docusaurus clear` (clearing only `.docusaurus` leaves the rspack cache, which replays the previous compile and reports a false green)
4. `npm run build`
5. Revert both files

All five versions build: **390** current, **326** v8, **310** v7, **306** v6, **200** warnings, the same set as before this change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,17 +99,17 @@ npm run spellcheck
 - A **directory** or anything matching a **regular expression**, update the following file: `cspell.json`
   - For example, we don't want to flag anything inside of code ticks (<code>`</code>) or code blocks (<code>```</code>), so there are regular expressions added to ignore anything inside of these.
 - An **entire line**, add the following comment above it:
-  ```markdown
-  <!-- cspell:disable-next-line -->
+  ```mdx
+  {/* cspell:disable-next-line */}
   ```
 - **Multiple lines**, add comments above and below the lines to be ignored:
 
-  ```markdown
-  <!-- cspell:disable -->
+  ```mdx
+  {/* cspell:disable */}
 
   <p>Everything inside of these comments will be ignored by the spell checkr. Proofread your own words carefully.</p>
 
-  <!-- cspell:enable -->
+  {/* cspell:enable */}
   ```
 
 > [!IMPORTANT]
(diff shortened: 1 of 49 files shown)


brandyscarney (changes requested):
Can we update the Contributing guide to reflect the proper comment syntax: [CONTRIBUTING.md](https://github.com/ionic-team/ionic-docs/blob/072d9a536a3e6c78ba4b2963e006a800da3c3998/CONTRIBUTING.md?plain=1#L93-L117)

brandyscarney (approved):
LGTM!

PR #4717, opened 2026-09-08
chore(i18n, mdx): adopt strict MDX syntax and remove the transitional shims

thetaPC:
Issue URL: internal

Two transitional shims in `plugins/docusaurus-plugin-ionic-component-api` are still carrying the `.md` to `.mdx` rename: the component API plugin emits every partial twice, and an `extensionAlias` lets any `.md` import resolve to a `.mdx` file.

The items in [Docusaurus 3.10's Strict MDX guidance](https://docusaurus.io/blog/releases/3.10#strict-mdx) are also outstanding, so `future.v4.mdx1CompatDisabledByDefault` cannot be turned on. Turning it on today fails the build:

```bash
Error: MDX compilation failed for file ".docusaurus/.../v5/action-sheet/methods.mdx"
Cause: Could not parse expression with acorn
```

## What is the new behavior?

The jp companion to #4716. Both shims are gone, and the Strict MDX items are done:

- **8** heading IDs in `docs/updating/9-0.mdx`, `{#id}` to `{/* #id */}`
- **111** HTML comments to `{/* */}`, in archived v5 and v6
- **4** admonition titles to `:::tip[Title]`, in archived v6

A further **9** commented-out blocks were deleted rather than converted, matching #4716. Each is a placeholder that v7 onward already dropped and nothing else references: an `--item-detail-push-show` TODO, four `TOC goes here`, two `TODO: image?`, and a stubbed `## Customization`.

That clears the last blocker on the v4 flag, which is deliberately not in this PR.

- [ ] Yes
- [x] No

## Other information

This branch needed one fix that main did not. The plugin was still emitting classic heading IDs into every generated partial, `### dismiss {#method-dismiss}`, where main had already moved to `{/* #method-dismiss */}`. That is a Strict MDX violation in generated output rather than in any source file, so no audit of the repo would surface it. It broke **544** generated partials under the flag until the two template lines were updated. Anchors are unchanged: all **359** api pages across v5 to v8 produce identical anchor lists before and after.

### How to test

Archived versions are the risk here, since 44 of the 48 files are in v5 and v6, so verify them with the flag on:

2. Add `mdx1CompatDisabledByDefault: true` under `future.v4` in `docusaurus.config.js`
3. `npx docusaurus clear` (clearing only `.docusaurus` leaves the rspack cache, which replays the previous compile and reports a false green)
4. `npm run build`
5. Revert both files

All four versions build: **460** v5, **307** v6, **310** v7, **326** v8, with **852** warnings, the same set as before this change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,17 +99,17 @@ npm run spellcheck
 - A **directory** or anything matching a **regular expression**, update the following file: `cspell.json`
   - For example, we don't want to flag anything inside of code ticks (<code>`</code>) or code blocks (<code>```</code>), so there are regular expressions added to ignore anything inside of these.
 - An **entire line**, add the following comment above it:
-  ```markdown
-  <!-- cspell:disable-next-line -->
+  ```mdx
+  {/* cspell:disable-next-line */}
   ```
 - **Multiple lines**, add comments above and below the lines to be ignored:
 
-  ```markdown
-  <!-- cspell:disable -->
+  ```mdx
+  {/* cspell:disable */}
 
   <p>Everything inside of these comments will be ignored by the spell checkr. Proofread your own words carefully.</p>
 
-  <!-- cspell:enable -->
+  {/* cspell:enable */}
   ```
 
 > [!IMPORTANT]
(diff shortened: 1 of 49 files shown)


brandyscarney (approved):
Looks good as long as this gets the contributing guide update too.

brandyscarney (approved):

PR #4724, opened 2026-09-10
test: prove the admonition check fires

thetaPC:
Throwaway. Confirms Check Admonitions reports the legacy title and ignores the bracketed, untitled, fenced and inline forms. Close without merging.


thetaPC:
Only needed for CI testing

PR #4725, opened 2026-09-10
chore(i18n, mdx): enable the v4 strict MDX flag and guard admonition syntax

thetaPC:
Issue URL: internal

Docusaurus still applies the MDX v1 compatibility shims: `mdx1Compat.comments`, `.admonitions` and `.headingIds` all default to `true`. Nothing on this branch relies on them any more, so they are no-ops that hide regressions. A page reintroducing `<!-- comment -->`, `:::note Title` or `{#heading-id}` builds fine today and breaks on the v4 upgrade instead.

## What is the new behavior?

The jp companion to #4723. `future.v4.mdx1CompatDisabledByDefault: true`, which flips all three to `false`. One functional line; the rest of that diff moves the existing `future` comment so each key is documented above itself.

Two of the three now fail the build outright. The third does not: `:::note My Title` stops being a directive and renders as literal `:::note` text on the page, with no warning and a successful build. The `check-admonitions` action covers that one case, on the files each PR changes.

- [ ] Yes
- [x] No

## Other information

The action is byte-identical to the one on `main`, so it is not re-tested here. Its behavior was proven on #4724, a throwaway PR that plants one legacy title alongside four forms that must not be flagged: the [failing check run](https://github.com/ionic-team/ionic-docs/actions/runs/34516694957/job/103003754714?pr=4724) reports only the legacy one, and the annotation lands [on the offending line](https://github.com/ionic-team/ionic-docs/pull/4724/changes#diff-db7c35e431e7659fc0d4a3dd9c24ff796f2574a405792760c49de43d47cd27f5R28). This branch has **0** legacy admonition titles, so the check lands green, and on this PR it reports `No MDX files changed`.

Docusaurus's own `unusedDirectives` warning is already enabled and catches `:::typo`, but it cannot catch this: it visits directive nodes, and `:::note My Title` parses to a plain paragraph. There is no node to visit, which is why the check has to work on the raw text before parsing.

The build produces **744** pages and **407** warnings, identical to the same tree with the flag off: nothing introduced, nothing resolved.

### How to test

1. `npm run build` and confirm it completes
2. Add `<!-- test -->` to any page under `docs/` and rebuild. It should now fail with `Could not parse expression with acorn` or `Unexpected character !`, where before it built silently. Revert.

Archived v5 to v7 are not built by default. They were verified with the flag on in #4717 against content identical to this branch, building **460** v5, **307** v6, **310** v7 and **326** v8 pages. To repeat it:

2. Run `npx docusaurus clear` (clearing only `.docusaurus` leaves the rspack cache)
3. Run `npm run build`
4. Revert

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -32,6 +32,7 @@ jobs:
       run: npm run test
     - name: 🔤 Spell Check
       run: npm run spellcheck
+    - uses: ./.github/workflows/actions/check-admonitions
     - uses: ./.github/workflows/actions/check-translations
     # Lint and spell check changes should be pushed
     # to the branch before the branch is merge eligible.
--- a/.github/workflows/actions/check-admonitions/action.yml
+++ b/.github/workflows/actions/check-admonitions/action.yml
@@ -1,0 +1,92 @@
+name: 'Check Admonitions'
+description: 'Fails on admonition titles using the MDX v1 syntax'
+runs:
+  using: 'composite'
+  steps:
+    # `:::note My Title` was MDX v1 syntax that Docusaurus rewrote for us.
+    # With `future.v4.mdx1CompatDisabledByDefault` that shim is off, and
+    # the block `:::note My Title` stops being a directive: it renders as
+    # literal `:::note` text on the page, with no warning and a successful
+    # build.
+    #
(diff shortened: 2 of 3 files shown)


ShaneK (approved):
Looks good to me 🚀

PR #4667, opened 2026-08-26
chore(docs): use markdown syntax for external links

thetaPC:
Issue URL: N/A

## What is the current behavior?

External links are written two ways: mostly markdown, but 338 as raw `<a target="_blank">`. 198 of those are missing `rel="noopener noreferrer"`.

The glossary is built from 41 `<section>` blocks with raw anchors, which prettier mangles into invalid HTML. So both glossaries sit in `.prettierignore`, and `index.md` needs a `prettier-ignore`.

Those anchors are HTML `id` attributes, not heading IDs, so Docusaurus can't validate them. 195 broken anchor references went unchecked, including a `[Babel](#babel)` link with no Babel entry.

## What is the new behavior?

- 338 raw anchors converted to markdown links
- Both glossaries rewritten as markdown headings (`### Term {/* #id */}`), all 41 IDs preserved
- Babel entry added, fixing the dead anchor
- Both prettier exclusions removed

Left alone: 3 anchors that wrap markup, and 7 inside code fences.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

**Readers see no difference.** Docusaurus already adds `target="_blank" rel="noopener noreferrer"` to external URLs, so markdown links render the same as the raw ones did.

Wins:

- **198 links gain `rel="noopener noreferrer"`**, closing a `window.opener` gap
- **Broken glossary anchors: 195 → 0**, now that Docusaurus can validate them
- **Glossary gets a table of contents** (42 terms), since raw `<h3>` is invisible to it

PR #4729, opened 2026-09-11
chore(i18n): sync main into translation/jp

thetaPC:
## Why

The last preserved merge of `main` into `translation/jp` was `e7bdaed5d`, **2022-07-28**. Every sync since has been squashed, so git never recorded them. That is why the MDX migration needed **9 hand-authored PRs** on this branch mirroring work already done on `main`, and why rename detection keeps failing across a four-year diff.

After this lands, the base advances to today and future syncs reconcile weeks instead of years.

## Resolution

**1,353** conflicts, **1,299** mechanical:

- `static/usage/` and `versioned_docs/` took `main`, neither is translated
- Translated `docs/` kept Japanese
- `docs/` under 20% translated took `main`
- 26 paths deleted on both sides, both renamed `.md` to `.mdx` independently
- `sidebars.js` kept all 52 Japanese labels, took main's new Upgrade Guides category

Three that needed judgement:

- **`docusaurus.config.js`** auto-merged into a duplicate `VERSIONS_JSON` block and would not parse. Fixed.
- **The component API plugin** still required `node-fetch`, which `main` has dropped. Kept jp's local-file read, removed the require.
- **`docs/angular/lifecycle.mdx`** had Japanese text saying components "must not use `OnPush`", which `main` has replaced with the opposite guidance. Swapped that one note to English pending translation, since untranslated beats wrong.

## Follow-up

27 files keep jp's older `<a href>` link markup. #4667 needs reapplying to the Japanese tree separately, kept out to avoid mixing a cosmetic sweep into this.


docs/angular/lifecycle.mdx
```diff
@@ -27,7 +27,13 @@ Ionic は Angular が提供するライフサイクルイベントを取り入
 Angular のコンポーネントライフサイクルイベントの詳細については、[コンポーネントライフサイクルのドキュメント](https://angular.jp/guide/lifecycle-hooks)をご覧ください。
 
 :::note
-`ion-nav` または `ion-router-outlet` を使用するコンポーネントは、 `OnPush` 変更検出方式を使用しないでください。そうすることで、 `ngOnInit` などのライフサイクル・フックが起動するのを防ぐことができます。また、非同期状態の変更は正しくレンダリングされない場合があります。
+
+If your pages keep state in plain fields rather than signals, the component hosting `ion-router-outlet` or `ion-tabs` needs eager change detection, as does every component between it and your application root. A change detection pass starts at the application root and skips a clean `OnPush` view along with everything below it, so an `OnPush` component above the outlet stops updates from reaching the routed pages under it. The pages themselves can use `OnPush`, as long as their state is a signal or they call `markForCheck()`.
```
thetaPC, line 31: JP said components using `ion-nav` or `ion-router-outlet` must not use `OnPush`, and that doing so prevents `ngOnInit` from firing. Main replaced that in #4694 with the opposite: you can use `OnPush` on pages, provided state is a signal or you call `markForCheck()`, plus the Angular 22 default change.

docs/reference/glossary.mdx
```diff
@@ -12,318 +12,203 @@ title: 用語集
...
+
+### Android Studio {/* #android-studio */}
+
+[Android Studio](https://developer.android.com/studio/)は、ネイティブAndroidアプリ開発の公式統合開発環境（IDE）です。
+
+### Autoprefixer {/* #autoprefixer */}
+
+[Autoprefixer](https://github.com/postcss/autoprefixer)は、手書きのSass/CSSコードにベンダー固有のプレフィックスを追加するツールです。これにより、あなたが書いた標準化されたCSSルールがすべての対応ブラウザで適用されることが保証されます。たとえば、さまざまなブラウザで使用されるすべてのフレックスボックスの構文を覚える代わりに、Autoprefixerを使えば単に<code>display: flex;</code>と書くだけで、自動的に正しいCSSが挿入されます。
+
+### Babel {/* #babel */}
```
thetaPC, line 31: New section, not translated

docs/angular/lifecycle.mdx
thetaPC: @rdlabo tagging you in case you want to add it to your translate queue.

docs/reference/glossary.mdx
thetaPC: @rdlabo tagging you in case you want to add it to your translate queue.

PR #4734, opened 2026-09-14
chore(ci): split OS-independent checks into their own jobs

thetaPC:
Issue URL: N/A

One matrix job runs every check on Windows and macOS. Several of those read files and report a verdict that cannot vary by operating system, so they run twice for the same answer.

From a recent run, Windows / macOS: Spell Check **27s / 13s**, Typecheck **4s / 2s**, Test **2s / 1s**. macOS bills at **10x** Linux, Windows at **2x**.

## What is the new behavior?

Four jobs:

- **Check Admonitions** (Linux) — the composite action only. It needs no dependencies, so it no longer waits on `npm ci`, lint, tests and spellcheck before reporting, and it stops annotating every finding twice.
- **Checks** (Linux) — Test, Typecheck, Spell Check
- **Cross-platform on {os}** (Windows, macOS) — Lint, check-translations, Check Diff
- **Verify** — aggregates the three

Also: `cache: npm` on both installing jobs, `NODE_VERSION` declared once at the top, and the matrix job renamed from `test`, which no longer ran any.

Lint stays in the matrix deliberately. Prettier rewrites line endings and Check Diff catches it, so it is the one check that genuinely differs by OS. It is also the most expensive step at **59s / 32s**.

Now that the admonition check has its own job, `fetch-depth: 0` on its checkout replaces the targeted `git fetch` it used to do. That workaround existed to avoid pulling history for a job full of unrelated steps, which no longer applies.

- [ ] Yes
- [x] No

## Other information

**This needs a branch protection change the moment it merges, or merges break.**

`main` and `major-10.0` currently require `Test on macOS-latest` and `Test on windows-latest`. Those job names no longer exist, so both branches will be waiting on checks that never report.

Switch both to requiring `Verify` alone. It fails if any of the three fail, and it keeps working if the matrix changes later, which is why it exists. `translation/jp` has no required checks, so nothing to do there.

One ordering note: GitHub's UI only offers names it has seen, so `Verify` becomes selectable after this runs once. Worth having someone with admin ready before merging.

The `Verify` pattern is borrowed from ionic-framework's `verify-*` jobs, which exist for the same reason.

The two `Test on ...` checks will never report: this PR renames those jobs, and branch protection still requires the old names. Updating it to require `Verify` instead clears them.


.github/workflows/CI.yml
```diff
@@ -6,9 +6,45 @@ name: Validate Docs
...
+
+    steps:
+    - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      with:
+        # The check diffs against the pull request base, so it needs history.
+        fetch-depth: 0
+    - uses: ./.github/workflows/actions/check-admonitions
+
+  # These produce the same verdict on any operating system, so they run once.
+  checks:
```
ShaneK, line 32: I'm less sure about the case-sensitivity reasoning. I dropped `import DocsCard from '@components/global/docsCard'` into `docs/index.mdx` and typecheck, spellcheck, test and lint all pass quite happily. Build that same tree on a case-sensitive filesystem and it dies with `Cannot find module '@components/global/docsCard'`. So the bug is real, but nothing on this job goes looking for it, and there's no `docusaurus build` step anywhere in the workflow on any OS.

Typecheck can't cover it either. Neither our `tsconfig.json` nor `@docusaurus/tsconfig` sets `forceConsistentCasingInFileNames`, so it's been on by default the whole time and Windows and macOS were already catching cased imports under `src`. The 279 pages importing from `@site/static/usage/` are the exposed ones, and `include` is `["src", "index.d.ts"]`, so tsc never sees any of them.

Adding `npm run build:preview` here is what would catch it. Happy for that to be a follow-up, but I think the description should drop that reasoning if it's not happening in this PR.

.github/workflows/CI.yml
```diff
@@ -22,23 +58,33 @@ jobs:
...
     - name: 🖌️ Lint
       run: npm run lint
-    - name: Test
-      run: npm run test
-    - name: 🧩 Typecheck
-      run: npm run typecheck
-    - name: 🔤 Spell Check
-      run: npm run spellcheck
-    - uses: ./.github/workflows/actions/check-admonitions
     - uses: ./.github/workflows/actions/check-translations
```
ShaneK: Lint's reason for staying in the matrix is clear, but this one doesn't get one and it's the other expensive step. Was that deliberate?

I did check whether the shared Check Diff was what pinned it here, and it isn't. Everything the translations check writes is either gitignored or untracked, and `git diff --exit-code` doesn't look at untracked files. So it could move over to the Linux job without dragging a guard along.

.github/workflows/CI.yml
```diff
@@ -22,23 +58,33 @@ jobs:
...
+    # eligible.
     #
     # The translation keys check should not produce any changes.
     - name: 🔍 Check Diff
       run: git diff --exit-code
       shell: bash
+
+  # One stable name for branch protection, so the matrix can change freely.
+  verify:
+    # The required check on main and the feature branches. Renaming this
```
ShaneK: Nit: there are two comments making the same point on this job, one above `verify:` and one here, so maybe collapse them into one.

The `admonitions` and `checks` jobs have the same thing going on. The reason you give in the description for splitting `admonitions` out is the better one anyway, that it needs no dependencies so it stops queueing behind `npm ci`, lint and the rest before it can report. Up to you!

.github/workflows/CI.yml
```diff
@@ -6,9 +6,45 @@ name: Validate Docs
 
 on: [pull_request]
```
ShaneK, line 7: Worth adding a `concurrency` group while you're restructuring this? It takes the workflow from 2 concurrent jobs to 5, two of them windows and macOS legs, so a second push to a PR stacks a lot more than it used to. Every multi-job workflow in ionic-framework has one, with the reasoning spelled out in `build.yml` as not consuming more runners than we need to.

```yaml
concurrency:
  group: ${{ github.ref }}
  cancel-in-progress: true
```

.github/workflows/actions/check-admonitions/action.yml
```diff
@@ -14,18 +14,12 @@ runs:
...
-    # rather than through `fetch-depth` on the checkout, which would pull
-    # the full history for every step in the job just to serve this one.
     - name: 🔎 Check Admonitions
       shell: bash
       env:
         BASE_SHA: ${{ github.event.pull_request.base.sha }}
         CHANGED_FILES: ${{ runner.temp }}/changed-files.txt
       run: |
-        git fetch --quiet --no-tags --depth=1 origin "$BASE_SHA"
         git diff --name-only --diff-filter=ACMR "$BASE_SHA" HEAD > "$CHANGED_FILES"
```
ShaneK, line 26: This action now depends on the caller setting `fetch-depth: 0`, which a composite action can't declare for itself. On a depth-1 clone this line gives `fatal: bad object <sha>` and nothing else, so someone adding a second caller or tidying the checkout later has nothing pointing them at the cause.

Could the requirement go in a comment here rather than only at the call site? The perf worry in the comment you removed doesn't really apply now either, the Check Admonitions job ran in 11s with full history.

ShaneK (approved):
Nice, this is a good split. The run dropped from 189s to 158s, and the admonition annotations show up at 11s now instead of waiting on Windows to finish, which is the bit I care about most.

One thing on the description though, the 10x/2x multipliers only kick in on private repos, so we're not actually paying for any of this. Doesn't make the change less worth doing, I'd just pitch it on the wall clock instead.

Left a few questions inline, the main one being whether the Linux job really buys us the case-sensitivity coverage. Happy for you to push back on any of them.

.github/workflows/CI.yml
thetaPC: Dropped that paragraph from the description. Adding npm run build:preview to the Linux job is worth doing, and I'd rather open a follow-up than grow this one.

.github/workflows/CI.yml
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4734/changes/dfa136e92dbfb72adb5ee5c3c5347118172597e3

.github/workflows/CI.yml
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4734/changes/44a11fd47464099e29648504d7548e91c5bc37f7

.github/workflows/CI.yml
thetaPC: Added, keyed on ${{ github.workflow }}-${{ github.ref }} rather than the ref alone, since groups are repo-wide and a nightly ja build is on the backlog here. Happy to match yours exactly if you prefer. https://github.com/ionic-team/ionic-docs/pull/4734/changes/a5df9cd4457e1b79bcf99d28464058137341d368

.github/workflows/actions/check-admonitions/action.yml
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4734/changes/a7cfb0fe80630af35f1883c60735f28a6931ed18

ShaneK (approved):
LGTM!

PR #4689, opened 2026-09-01
docs(modal): reorganize Styling and add Sizing to Content section

brandyscarney:
## Description
- Reorganized the Modal API Styling section to have a top-level Theming heading and subsections
- Added a `Sizing to Content` section detailing how to set `--height` to a content size to have the modal resize to its rendered content & added a playground example
- Sibling PR to https://github.com/ionic-team/ionic-framework/pull/31413

## Other information

Previews:
- [Before](https://ionicframework.com/docs/api/modal#styling)
- [After](https://ionic-docs-git-fw-7731-ionic1.vercel.app/docs/api/modal#theming)

PR #4736, opened 2026-09-15
fix(links): drop the hardcoded baseUrl and repair stale anchors

thetaPC:
Issue URL: internal

The Japanese site renders **724** broken links. English renders 21, plus **50** broken anchors nobody had noticed. The archived v5, v6 and v7 docs carry the same defects and have never been checked, because they are not built.

Nearly every broken link is one defect: links hardcode the `/docs` baseUrl. Docusaurus prepends the baseUrl of the locale being built, and the Japanese one is `/docs/ja/`, so the prefix doubles into `/docs/ja/docs/theming/basics`. English hides it, because a value that already starts with its own baseUrl is passed through untouched.

The anchors are unrelated: fragments pointing at headings renamed years ago. `#style-placement` and `#optimizing-your-build` were v5 sections, dead through four majors.

## What is the new behavior?

English is at **0** broken links and **0** broken anchors. Japanese links drop **724 to 152**, with nothing left that originates in an English source file. Built with all five versions, the archived trees come out clean apart from five links noted below.

- Markdown links use root absolute paths carrying the target's extension, `/api/toolbar.mdx`. The extension matters: the resolver only fires on `.md` and `.mdx`, and it searches the localized content path first, which makes this the one style that is both locale correct and version correct.
- JSX `href` attributes take the same path without an extension, since no resolver runs there and the value goes straight to `Link`.
- One sidebar link is rendered into every docs page and accounted for **261** by itself.
- The API plugin now drops the baseUrl as it renders the fetched JSON. Those links reach us from three separately generated sources, so an upstream PR alone would have missed v8 and the Japanese API data.
- Anchors were resolved against the ids that actually exist in the built HTML, per version. A section removed in one version often still exists in an older one, so only the genuinely stale links changed.
- Links pointing into archived versions now use the frozen archive URLs, since those versions are not built.

- [ ] Yes
- [x] No

## Other information

**60 anchors will still be reported, and all 60 work.** Docusaurus collects anchors only from components that call `collectAnchor`, so a raw `<section id="cli">` written straight into MDX is invisible to it. The v5 and v6 glossaries are built that way. These should not be "fixed".

**5 links stay broken in v5.** They come from `@ionic/docs@5`, which uses a relative link style dropped after v5, and exist only in a build-time partial, so there is no file to edit. v5 is not built.

Japanese broken anchors rise 66 to 148. That is not new breakage: links that used to 404 now land on the correct translated page, whose heading slug is Japanese, so the fragment misses instead.

ja never builds in PR previews, so none of this is visible from the preview link. I verified with a local two-locale build, and again with v5, v6 and v7 temporarily added to `versions.json`.

Everything still broken lives in `translation/jp`. The plan is this PR, then a normal main to jp sync, which carries most of the prose fixes as conflicts to resolve, then a small jp PR for the rest. **103** of the Japanese leftovers are a single missing heading id in the jp `fundamentals.mdx`, on a heading whose slug is Japanese.


versioned_docs/version-v8/api/modal.mdx
```diff
@@ -181,7 +181,7 @@ Modals are presented at the root of your application so they overlay your entire
 
 :::note
 
-If you are building an Ionic Angular app, the styles need to be added to a global stylesheet file. Read [Style Placement](#style-placement) in the Angular section below for more information.
+If you are building an Ionic Angular app, the styles need to be added to a global stylesheet file.
```
thetaPC: The section only existed in v5.

docs/api/modal.mdx
```diff
@@ -181,7 +181,7 @@ Modals are presented at the root of your application so they overlay your entire
 
 :::note
 
-If you are building an Ionic Angular app, the styles need to be added to a global stylesheet file. Read [Style Placement](#style-placement) in the Angular section below for more information.
+If you are building an Ionic Angular app, the styles need to be added to a global stylesheet file.
```
thetaPC: The section only existed in v5.

docs/troubleshooting/native.mdx
```diff
@@ -48,7 +48,7 @@ Running an app on an iOS device requires a provisioning profile. If a provisioni
 Error: Error code 65 for command: xcodebuild with args: -xcconfig,/Users/ionitron/projects/my-project/platforms/ios/cordova/build-debug.xcconfig,-workspace,SC project.xcworkspace,-scheme,SC project,-configuration,Debug,-sdk,iphonesimulator,-destination,platform=iOS Simulator,name=iPhone X,build,CONFIGURATION_BUILD_DIR=/Users/ionitron/projects/my-project/platforms/ios/build/emulator,SHARED_PRECOMPS_DIR=/Users/ionitron/projects/my-project/platforms/ios/build/sharedpch
 ```
 
-This error is an error code from Xcode that can be caused by provisioning issues or outdated cordova dependencies. To fix this error first make sure a provisioning profile has been generated using the above instructions and then try to [run the app from Xcode](../developing/ios.mdx#running-with-xcode).
+This error is an error code from Xcode that can be caused by provisioning issues or outdated cordova dependencies. To fix this error first make sure a provisioning profile has been generated using the above instructions and then try to [run the app from Xcode](/developing/ios.mdx).
```
thetaPC, line 51: That section only existed in v5 and v6.

versioned_docs/version-v8/troubleshooting/native.mdx
```diff
@@ -48,7 +48,7 @@ Running an app on an iOS device requires a provisioning profile. If a provisioni
 Error: Error code 65 for command: xcodebuild with args: -xcconfig,/Users/ionitron/projects/my-project/platforms/ios/cordova/build-debug.xcconfig,-workspace,SC project.xcworkspace,-scheme,SC project,-configuration,Debug,-sdk,iphonesimulator,-destination,platform=iOS Simulator,name=iPhone X,build,CONFIGURATION_BUILD_DIR=/Users/ionitron/projects/my-project/platforms/ios/build/emulator,SHARED_PRECOMPS_DIR=/Users/ionitron/projects/my-project/platforms/ios/build/sharedpch
 ```
 
-This error is an error code from Xcode that can be caused by provisioning issues or outdated cordova dependencies. To fix this error first make sure a provisioning profile has been generated using the above instructions and then try to [run the app from Xcode](../developing/ios.mdx#running-with-xcode).
+This error is an error code from Xcode that can be caused by provisioning issues or outdated cordova dependencies. To fix this error first make sure a provisioning profile has been generated using the above instructions and then try to [run the app from Xcode](/developing/ios.mdx).
```
thetaPC, line 51: That section only existed in v5 and v6.

versioned_docs/version-v7/api/modal.mdx
```diff
@@ -166,7 +166,7 @@ import SheetHandleBehaviorExample from '@site/static/usage/v7/modal/sheet/handle
 Modals are presented at the root of your application so they overlay your entire app. This behavior applies to both inline modals and modals presented from a controller. As a result, custom modal styles can not be scoped to a particular component as they will not apply to the modal. Instead, styles must be applied globally. For most developers, placing the custom styles in `global.css` is sufficient.
 
 :::note
-If you are building an Ionic Angular app, the styles need to be added to a global stylesheet file. Read [Style Placement](#style-placement) in the Angular section below for more information.
+If you are building an Ionic Angular app, the styles need to be added to a global stylesheet file.
```
thetaPC: The section only existed in v5.

versioned_docs/version-v7/troubleshooting/native.mdx
```diff
@@ -48,7 +48,7 @@ Running an app on an iOS device requires a provisioning profile. If a provisioni
 Error: Error code 65 for command: xcodebuild with args: -xcconfig,/Users/ionitron/projects/my-project/platforms/ios/cordova/build-debug.xcconfig,-workspace,SC project.xcworkspace,-scheme,SC project,-configuration,Debug,-sdk,iphonesimulator,-destination,platform=iOS Simulator,name=iPhone X,build,CONFIGURATION_BUILD_DIR=/Users/ionitron/projects/my-project/platforms/ios/build/emulator,SHARED_PRECOMPS_DIR=/Users/ionitron/projects/my-project/platforms/ios/build/sharedpch
 ```
 
-This error is an error code from Xcode that can be caused by provisioning issues or outdated cordova dependencies. To fix this error first make sure a provisioning profile has been generated using the above instructions and then try to [run the app from Xcode](../developing/ios.mdx#running-with-xcode).
+This error is an error code from Xcode that can be caused by provisioning issues or outdated cordova dependencies. To fix this error first make sure a provisioning profile has been generated using the above instructions and then try to [run the app from Xcode](../developing/ios.mdx).
```
thetaPC, line 51: That section only existed in v5 and v6.

versioned_docs/version-v8/api/modal.mdx
brandyscarney: I removed these in my PR also: [link](https://github.com/ionic-team/ionic-docs/pull/4689/changes/BASE..f73878ac893e79840013d3eefd03e596e2682682#diff-263c9867f3f554b638456cb0ae96144ed26ff6d7befc5974787dfeeb281caae9)

versioned_sidebars/version-v8-sidebars.json
```diff
@@ -51,7 +51,7 @@
         {
           "type": "link",
           "label": "Responsive Grid",
-          "href": "/docs/api/grid"
+          "href": "/api/grid"
```
ShaneK:
```suggestion
          "href": "/v8/api/grid"
```

This renders as `/docs/api/grid`, so the v8 sidebar sends people to the v9 docs. Pre-existing rather than something you introduced, but the line's already open here, and the v6/v7 sidebars have the same entry.

plugins/docusaurus-plugin-ionic-component-api/index.js
```diff
@@ -204,7 +219,7 @@ ${properties
 
     let docs = prop.docs;
     if (isVirtual) {
-      docs = `${docs}\n\nThis is a [virtual property](/docs/core-concepts/fundamentals#virtual-properties) that is set once during initialization and will not update if you change its value after the initial render.`;
+      docs = `${docs}\n\nThis is a [virtual property](/core-concepts/fundamentals#virtual-properties) that is set once during initialization and will not update if you change its value after the initial render.`;
```
ShaneK, line 222: Under `ja` this resolves to the right page but the Japanese fundamentals heading has no explicit id, so it comes out as 103 broken anchors (they were hard 404s before this PR, so still a win). Adding `{/* #virtual-properties */}` to that heading in the `translation/jp` copy clears all 103, matching what's already there for `#adaptive-styling`.

versioned_docs/version-v5/native.mdx
```diff
@@ -36,7 +36,7 @@ Build native-powered app experiences with a collection of open source and premiu
     {/* prettier-ignore */}
     <p>A modern, open source native runtime built and maintained by the Ionic team and the Capacitor community. Our recommended native solution.</p>
   </DocsCard>
-  <DocsCard header="Cordova plugins" img="/img/native/cordova@2x.png" href="/native/community">
+  <DocsCard header="Cordova plugins" img="/img/native/cordova@2x.png" href="/v5/native/community">
```
ShaneK, line 39: Any reason this one keeps an in-site path rather than the archive URL from `versionsArchived.json` like the others here? It does work and the file's in an unbuilt tree anyway, so purely consistency - up to you!

ShaneK (approved):
Looks good to me, great work! English comes out at 0 broken links and 0 broken anchors in a local build, down from 7 and 40 on main, and the couple of things I left are nits. 🚀

versioned_sidebars/version-v8-sidebars.json
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4736/changes/d33c6c533b2dc8a05d76d6b853fc8c87b7599118

plugins/docusaurus-plugin-ionic-component-api/index.js
thetaPC: Agreed. The jp heading is `### 仮想プロパティ`, so it needs `{/* #virtual-properties */}` the same way `#adaptive-styling` already does two headings up.

It points at a bigger question though: 1,072 translated ja headings have no pinned id, so their anchors can't match English. I want to look into whether the standard should be that headings in `docs/` carry explicit ids.

versioned_docs/version-v5/native.mdx
thetaPC: The archive URLs elsewhere are for targets that don't exist in any built tree. Here the target is in the same tree, `native-community.mdx` with slug: `/native/community`, so there's a real page to point at.

Kept it internal because the archive URL is never the better of the two. If v5 were ever unarchived it would send people to a frozen snapshot instead of the page that just got built, and it's another hardcoded deploy URL we'd have to remember to update if that archive ever moves, which nothing in the build would catch.

PR #4738, opened 2026-09-16
chore(i18n): sync main into translation/jp

thetaPC:
sync from main to translation/jp


docs/api/modal.mdx
thetaPC, line 1: Sections have been reorganized so their translations are no longer applicable. English was taken for the up to date.

@rdlabo tagging you in case you want to add it to your translate queue.

PR #4739, opened 2026-09-16
fix(links): repair the broken links on the Japanese site

thetaPC:
Issue URL: internal

The Japanese site renders **152** broken links, and none of them are visible from this branch. It builds as a single locale at `/docs/`, so the defect only appears once this content is consumed as the `ja` locale of the main site, where the baseUrl is `/docs/ja/`.

That is the cause: links hardcode the `/docs` baseUrl, so Docusaurus prepends the locale's own and the prefix doubles into `/docs/ja/docs/theming/basics`.

## What is the new behavior?

**0** broken links. **129** link fixes across 41 files.

Rather than re-deriving the rules, I extracted the old to new mapping from main's own diff in #4736, **97** distinct rewrites, and applied it here. Two of them map the same source to different targets depending on context, `/docs/components` becoming `/components` in a JSX `href` but `/components.mdx` in a markdown link, so they are applied context-aware.

That covered 123. The remaining 6 have no counterpart on main:

- `react.mdx` is jp-only, and its "Add Ionic to Existing React App" card pointed at a page renamed upstream long ago. The page here is `react/add-to-existing.mdx`.
- `troubleshooting/build.mdx` pointed at `/docs/faq/runtime#provider-`, a page that does not exist.
- `api/app.mdx` linked `./リップルエフェクト`, which does not exist. The page is `api/ripple-effect.mdx`.
- `api/item.mdx` had `href="useBaseUrl('api/modal')"` quoted where it needed braces.

- [ ] Yes
- [x] No

## Other information

**145 broken anchors are left in place on purpose**, and **103** of them are a single heading: `### 仮想プロパティ` needs `{/* #virtual-properties */}`. Main's English heading carries no pinned id, so this branch mirrors main until we settle whether headings should carry explicit ids. The other 42 are the same class, translated headings whose slugs cannot match the English anchor.


docs/cli.mdx
```diff
@@ -39,14 +39,14 @@ $ ionic <command> <subcommand> --help
...
-Ionic CLI は、[TypeScript](/docs/reference/glossary#typescript)と [Node.js](/docs/reference/glossary#node)で構築されています。Node 10.3+をサポートしていますが、最新の Node LTS が推奨です。開発状況は、オープンソースの [GitHub リポジトリ](https://github.com/ionic-team/ionic-cli)で確認できます。
+Ionic CLI は、[TypeScript](/reference/glossary.mdx#typescript)と [Node.js](/reference/glossary.mdx#node)で構築されています。Node 10.3+をサポートしていますが、最新の Node LTS が推奨です。開発状況は、オープンソースの [GitHub リポジトリ](https://github.com/ionic-team/ionic-cli)で確認できます。
 
 ## トラブルシューティング
 
 Ionic CLI の問題をトラブルシューティングするには、以下の方法が役立ちます:
 
 - 最新の Ionic CLI がインストールされていることを確認してください。インストール済みの Ionic CLI のバージョンは、`ionic --version` を実行して確認できます。Ionic CLI のバージョンは Ionic Framework のバージョンとは関係ありません。
-- 最新の Node LTS がインストールされていることを確認してください。設定手順については [Node & npm](/docs/intro/environment#node-npm) の環境設定を参照してください。
+- 最新の Node LTS がインストールされていることを確認してください。設定手順については [Node & npm](/intro/environment.mdx#node--npm) の環境設定を参照してください。
```
ShaneK, line 49: This fragment doesn't resolve. The heading it points at [pins `node-npm`](https://github.com/ionic-team/ionic-docs/blob/5c34e7bb92b037c41bc6badbbb7cb87c27b3d6f9/docs/intro/environment.mdx#L31) on this branch, while `node--npm` is main's slug for `## Node & npm`, so it looks like it came along with the mapping from #4736.

In a `ja` build it's broken from two pages, this one and [a self-link on the target page](https://github.com/ionic-team/ionic-docs/blob/5c34e7bb92b037c41bc6badbbb7cb87c27b3d6f9/docs/intro/environment.mdx#L13) you haven't touched. Once we pin ids in the English source it'll be `node--npm` on both sides and this line is already right, so it's the jp pin that needs to move. [One other link](https://github.com/ionic-team/ionic-docs/blob/5c34e7bb92b037c41bc6badbbb7cb87c27b3d6f9/docs/studio/setup/installation.mdx#L8) still points at `node-npm`, so that one needs moving too.

docs/core-concepts/fundamentals.mdx
```diff
@@ -77,4 +77,4 @@ Ionic Framework は[CSS](https://developer.mozilla.org/en-US/docs/Web/CSS)を使
 
 `mode`プロパティは、コンポーネントに使用するプラットフォームスタイルを決定する仮想プロパティです。コンポーネントレベルで設定することも、アプリ設定を通じてグローバルに設定することもできます。どちらの場合も、初期化中に一度設定され、コンポーネントのライフサイクル中は変更されません。
```
ShaneK, line 78: The `### 仮想プロパティ` heading just above is the one from the #4736 thread, [where you agreed it needs a pinned id](https://github.com/ionic-team/ionic-docs/pull/4736#discussion_r4029161400). It's still 103 of the 145 broken anchors in a ja build, so it's the biggest single win in here.

There are 11 more this PR retargets with the same problem, where the fragment is main's English slug and the ja heading has no pin. The `#methods` one turns up in four quickstarts, and the rest are spread across the glossary, the 6.0 updating guide and a couple of others.

All of it's covered by the heading-id standard we settled on, so no need to chase them here unless you want the `仮想プロパティ` one while the file's already open.

docs/reference/glossary.mdx
```diff
@@ -95,7 +95,7 @@ ES5はEcmaScriptの第5版を指します。簡単に言えば、ES5は今日の
 ### Genymotion {/* #genymotion */}
 
 Genymotionはサードパーティ製のAndroidエミュレーターです。非常に高速で、Android上でアプリを素早くテストするのに役立ちます。
-Genymotionに関する詳細情報は、[リソースセクション](../developing/tips#using-genymotion-android)をご覧ください。
+Genymotionに関する詳細情報は、[リソースセクション](../developing/tips#using-the-genymotion-android-emulator)をご覧ください。
```
ShaneK:
```suggestion
Genymotionに関する詳細情報は、[リソースセクション](/developing/tips.mdx#using-the-genymotion-android-emulator)をご覧ください。
```

This is the one link the PR edits that didn't pick up the root-absolute form with the extension. Without it the markdown resolver never fires, so this one skips the mechanism the rest of the PR is standardizing on. The anchor half is one of the 11, so it sorts itself out with the heading ids.

ShaneK (changes requested):
Great work on this, ja comes out at 0 broken links in a local build, down from 100. Just the anchor in `cli.mdx` and the glossary suggestion and I think we're there.

docs/cli.mdx
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4739/changes/c59f32b15bce70c93de6e66a5d104a531105a6f2

docs/core-concepts/fundamentals.mdx
thetaPC: Leaving these for the heading-id PR, including `仮想プロパティ`. Pinning one here would just be a partial version of that change. https://github.com/ionic-team/ionic-docs/commit/c59f32b15bce70c93de6e66a5d104a531105a6f2

docs/reference/glossary.mdx
thetaPC: https://github.com/ionic-team/ionic-docs/pull/4739/changes/24312dae54ca977603997c13030246a50ad68a94

ShaneK (approved):
Great work!

PR #4744, opened 2026-09-18
docs(angular): document boolean input changes in the v10 migration guide

thetaPC:
Issue URL: https://github.com/ionic-team/ionic-framework/pull/31442

The v10 migration guide has no Angular section. Boolean inputs on Ionic components are getting an input transform in v10, and nothing documents it.

## What is the new behavior?

Adds **Getting Started > Angular > Boolean Inputs** to the v10 migration guide, covering:

- The new shorthand, `<ion-item button>`, which previously failed to compile
- The break: declaring the transform turns on type checking Angular did not do before, so `[button]="items.length"` now fails with `TS2322` and needs coercing
- `null` and `undefined` pass through instead of coercing to `false`, the opposite of Angular's own `booleanAttribute`, because components like `ion-item` and sheet `ion-modal` treat "not set" as a third state

- [ ] Yes
- [x] No

## Other information

N/A


docs/updating/10-0.mdx
```diff
@@ -13,3 +13,37 @@ For a **complete list of breaking changes** from Ionic 9 to Ionic 10, please ref
...
+```
+
+Coerce the expression to a boolean:
+
+```diff
+- <ion-item [button]="items.length"></ion-item>
++ <ion-item [button]="items.length > 0"></ion-item>
+```
+
+Coercion also moves from Stencil to Angular, which changes the result for numbers. `0` and `NaN` previously became `false` and now become `true`, matching Angular's own `booleanAttribute`. Without `strictTemplates` there is no compile error to catch the binding above, so an empty list now disables the item rather than enabling it. Coercing the expression fixes both the type error and the runtime change.
```
ShaneK, line 42:
```suggestion
Coercion also moves from Stencil to Angular, which changes the result for numbers. `0` and `NaN` previously became `false` and now become `true`, matching Angular's own `booleanAttribute`. Without `strictTemplates` there is no compile error to catch the binding above, so an empty list now renders the item as tappable rather than plain. Coercing the expression fixes both the type error and the runtime change.
```

This is the same correction as [the one on the framework PR](https://github.com/ionic-team/ionic-framework/pull/31442#discussion_r4073332037), since the paragraph came across with it. The example binds `button` rather than `disabled`.

docs/updating/10-0.mdx
```diff
@@ -13,3 +13,37 @@ For a **complete list of breaking changes** from Ionic 9 to Ionic 10, please ref
...
+
+Boolean inputs now declare an input transform, so attribute presence is an explicitly supported way to set them:
+
+```html
+<!-- Both set `button` to `true` -->
+<ion-item button></ion-item>
+<ion-item [button]="true"></ion-item>
+```
+
+Declaring the transform also turns on type checking for these inputs, which Angular did not do before. The generated component wrappers declare no class fields, so Angular had nothing to check a binding against and accepted any value. A binding that passes something outside `boolean | string | null | undefined` now fails to compile. The common case is a truthiness binding on a number:
```
ShaneK, line 29: This holds under `strictTemplates`, though the caveat doesn't turn up until the coercion paragraph further down, where it's attached to a different point. Could it go here instead? Someone without `strictTemplates` reads this, expects a compile error, gets none, and never reaches the part that actually affects them.

ShaneK (approved):
Looks good to me! I checked the behavior claims against the v10 source and both the `ion-item` and `ion-modal` examples hold up. One correction inline on the coercion paragraph, which came across from the framework PR.

